PR #701, opened 2022-10-19
BugFix: Add support for `withControlled` from vee-validate@4.7.0

pleek91:
# Description
Vee validate introduced a method property on `handleSubmit` which we are wrapping in our `useForm` composition. This caused a type error because we didn't implement the same type. 

Added support for `withControlled` which fixes the type error.


zhen0:
Do we still need this one @pleek91?

pleek91:
> Do we still need this one @pleek91? 

We do if we want to update. There's no reason to not merge this. It just never got a review and was forgotten

zhen0:
hmmm that wasn't what I expected to happen

zhen0:
Closing in favor of #1068

PR #884, opened 2022-12-06
Bugfix: Radar app + node refreshes

znicholasbrown:
This PR addresses an issue where Radar wasn't updating appropriately, requiring a refresh to get run results. It also slightly tweaks the method used to compute task run duration. 

Closes: https://github.com/PrefectHQ/prefect/issues/7660

### DIFF
--- a/src/models/TaskRun.ts
+++ b/src/models/TaskRun.ts
@@ -84,6 +84,6 @@ export class TaskRun implements ITaskRun {
   }
 
   public get duration(): number {
-    return this.totalRunTime ?? this.estimatedRunTime ?? 0
+    return (this.totalRunTime && this.totalRunTime > 0 ? this.totalRunTime : this.estimatedRunTime) ?? 0
   }
 }
--- a/src/components/RadarApp.vue
+++ b/src/components/RadarApp.vue
@@ -104,10 +104,6 @@
     return items
   })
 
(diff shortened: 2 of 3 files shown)


src/models/TaskRun.ts
```diff
@@ -84,6 +84,6 @@ export class TaskRun implements ITaskRun {
   }
 
   public get duration(): number {
-    return this.totalRunTime ?? this.estimatedRunTime ?? 0
+    return (this.totalRunTime && this.totalRunTime > 0 ? this.totalRunTime : this.estimatedRunTime) ?? 0
```
znicholasbrown, line 87: Changed this to address the fact that the API returns a `totalRunTime` of 0 until the run has entered a terminal state so this would never fall back to the `estimatedRunTime`

pleek91: Isn't there a duration getter on the task model? There is on a flow run. @zhen0 and I fixed this same bug with flow runs.

znicholasbrown: On the task run model? That's what this, no? Am I missing something else?

pleek91: I'm blind 🤦‍♂️

brandonreid (approved):

PR #895, opened 2022-12-07
Enhancement: Hiding actions for the Flow Runs title

nicolascribbles:
## What changed
- Added an optional boolean prop that allows me to hide the actions when I don't need them

## Why it changed
- [Add checklist to "run a flow to get started" screens #2024](https://github.com/PrefectHQ/nebula-ui/issues/2024)

## Related PRs
- [Added a copy button for the P-Code element #534](https://github.com/PrefectHQ/prefect-design/pull/534)


src/components/PageHeadingFlowRuns.vue
```diff
@@ -1,14 +1,24 @@
...
+  import { defineProps, withDefaults, computed } from 'vue'
   import PageHeading from '@/components/PageHeading.vue'
   import  SavedFilters from '@/components/SavedFilters.vue'
 
+
+  const props = withDefaults(defineProps<{
+    hideActions?: boolean,
+  }>(), {
+    hideActions: false,
+  })
```
stackoverfloweth, line 19: no need to define a default for this, as the default is (and always should be) falsey
```suggestion
  const props = defineProps<{
    hideActions?: boolean,
  }>()
```

src/components/PageHeadingFlowRuns.vue
```diff
@@ -1,14 +1,24 @@
...
+
+  const hideActions = computed(() => props.hideActions)
```
stackoverfloweth, line 21: I'm guessing this line is only here to ensure props.hideActions is reactive? I don't think that's necessary here (@pleek91 can you verify?), but if it is I would suggest vue's `toRef`

```ts
const hideActions = toRef(props, 'hideActions')
```

pleek91: If you're only using it in the template it's not needed.

stackoverfloweth: @nicolascribbles I still don't think you need this line

nicolascribbles: @stackoverfloweth updated!

stackoverfloweth (approved):

pleek91 (approved):

PR #897, opened 2022-12-08
Enhancement: Add a quick run button and custom run buttons to the deployments table

marichka-offen:
This PR extract Quick Run and Custom Run deployment overflow items into separate reusable components. I added both buttons back to the Deployment page header (RunMenu and 3-dot menu for small screens) and to the Deployments table 3-dot menu.

Screenshot 2022-12-08 at 12 23 20 PM
Screenshot 2022-12-08 at 12 23 05 PM
Screenshot 2022-12-08 at 12 22 33 PM

Closes https://github.com/PrefectHQ/orion-design/issues/737


src/components/DeploymentCustomRunLink.vue
```diff
@@ -0,0 +1,19 @@
+<template>
```
znicholasbrown, line 1: Can we rename this component to something like `DeploymentCustomRunOverflowMenuItem`? I think that better captures the view layer responsibility

src/components/DeploymentCustomRunLink.vue
```diff
@@ -0,0 +1,19 @@
...
+  <router-link :to="routes.deploymentFlowRunCreate(deploymentId)">
+    <p-overflow-menu-item class="run-menu__overflow-menu-item">
```
znicholasbrown, line 3:
```suggestion
    <p-overflow-menu-item>
```
This class isn't needed (and was incorrect)

src/components/DeploymentQuickRun.vue
```diff
@@ -0,0 +1,42 @@
+<template>
+  <p-overflow-menu-item class="run-menu__overflow-menu-item" @click="run">
```
znicholasbrown, line 2:
```suggestion
  <p-overflow-menu-item @click="run">
```
This class isn't needed (and was incorrect)

src/components/DeploymentQuickRun.vue
```diff
@@ -0,0 +1,42 @@
+<template>
```
znicholasbrown, line 1: Same comment as above - can we rename this to something like `DeploymentQuickRunOverflowMenuItem`?

znicholasbrown (commented):
Just a few small requests @marichka-offen

src/components/DeploymentQuickRun.vue
pleek91: +1

src/components/DeploymentQuickRunOverflowMenuItem.vue
```diff
@@ -0,0 +1,42 @@
...
+  import { showToast } from '@prefecthq/prefect-design'
+  import { h } from 'vue'
+  import { useRouter } from 'vue-router'
+  import ToastFlowRunCreate from './ToastFlowRunCreate.vue'
+  import { useWorkspaceApi, useWorkspaceRoutes } from '@/compositions'
+  import { localization } from '@/localization'
+  import { Deployment } from '@/models'
+
+  const props = defineProps<{
+    deployment: Deployment,
```
pleek91, line 19: You're only using the id in this component so I would accept just the `deploymentId` similar to the other menu item.

src/components/PageHeadingDeployment.vue
```diff
@@ -3,32 +3,23 @@
...
-          <div class="page-heading-deployment__run-menu-item">
-            <router-link :to="routes.deploymentFlowRunCreate(deployment.id)">
-              <p-overflow-menu-item>
-                Run
-              </p-overflow-menu-item>
-            </router-link>
-          </div>
+        <template v-if="can.run.deployment && !media.sm" #additional-items>
+          <DeploymentQuickRunOverflowMenuItem :deployment="deployment" />
+          <DeploymentCustomRunOverflowMenuItem :deployment-id="deployment.id" />
```
pleek91, line 11: Can we adopt the `showAll` pattern from FlowRunMenu for the deployment menu as well? That way the DeploymentMenu has all its menu items and there's just a toggle to set if you want the full menu or just the concise menu.

pleek91 (commented):
This is looking really good!

src/components/PageHeadingDeployment.vue
marichka-offen: Totally space out on that, it simplified everything so much!

pleek91: Nice!

src/components/DeploymentQuickRunOverflowMenuItem.vue
```diff
@@ -0,0 +1,41 @@
...
+          type: 'scheduled',
+          message: 'Run from the Prefect UI with defaults',
+        },
+      })
+
+      const toastMessage = h(ToastFlowRunCreate, { flowRun, flowRunRoute: routes.flowRun, router, immediate: true })
+      showToast(toastMessage, 'success')
+    } catch (error) {
+      showToast(localization.error.scheduleFlowRun, 'error')
+      console.warn(error)
```
pleek91, line 38: We should always log errors as errors to the console. Warnings should be reserved for something that didn't fail per say but might not be what the developer intended. 

```suggestion
      console.error(error)
```

pleek91 (approved):
One last thing then this LGTM

PR #905, opened 2022-12-09
Enhancement: Add ConcurrencyLimitMenu to ConcurrencyLimitsTable

marichka-offen:


src/components/ConcurrencyLimitMenu.vue
```diff
@@ -12,27 +12,27 @@
...
-    name: 'ConcurrencyLimitMenu',
+  import { PIconButtonMenu, POverflowMenuItem } from '@prefecthq/prefect-design'
+  import { defineComponent } from 'vue'
+
+  export default defineComponent({
+    name: 'ConcurrencyLimitsMenu',
     expose: [],
     inheritAttrs: false,
-  }
+  })
```
pleek91, line 22: I believe this was correct before? We don't usually use define component but rather just export and object.

stackoverfloweth (approved):

src/components/ConcurrencyLimitMenu.vue
marichka-offen: I can make a PR to remove them all

Screenshot 2022-12-09 at 5 12 51 PM

stackoverfloweth: no you're not @pleek91, we moved away from `defineComponent`, must have just missed a few

PR #906, opened 2022-12-09
Chore: Remove defineComponent in ..Menu components

marichka-offen:


stackoverfloweth (approved):
@marichka-offen based on your other comment, I expected 6 components that needed updating?

marichka-offen:
@stackoverfloweth The other two are complex like this, it's been a while since I dealt with class components. Can I remove it safely here as well??
Screenshot 2022-12-09 at 5 21 46 PM

stackoverfloweth:
@marichka-offen, those are fine as is. Thanks for explaining

PR #907, opened 2022-12-10
Enhancement: Add the concurrency limit empty state component to concurrency limits table

zhen0:
Add the concurrency limit empty state component to concurrency limits table


src/components/ConcurrencyLimitsTable.vue
```diff
@@ -25,6 +26,7 @@
 <script lang="ts" setup>
   import { useSubscription } from '@prefecthq/vue-compositions'
   import { computed } from 'vue'
+  import ConcurrencyLimitsPageEmptyState from './ConcurrencyLimitsPageEmptyState.vue'
```
znicholasbrown, line 29: Can you add this to the import list on line 30 for clarity?

znicholasbrown:
1 comment @zhen0, otherwise LGTM

src/components/ConcurrencyLimitsTable.vue
```diff
@@ -1,5 +1,6 @@
 <template>
-  <p-table class="concurrency-limits-table__table" :columns="columns" :data="concurrencyLimits">
+  <ConcurrencyLimitsPageEmptyState v-if="empty" />
```
pleek91, line 2: Interesting to see the empty state in the orion-design table component. Usually orion-ui and nebula-ui implement them both separately. Is this a conscious choice?

src/components/ConcurrencyLimitsTable.vue
```diff
@@ -49,5 +51,6 @@
   ]
 
   const concurrencyLimitSubscription = useSubscription(api.concurrencyLimits.getConcurrencyLimits)
-  const concurrencyLimits = computed(() => concurrencyLimitSubscription.response ?? [])
+  const concurrencyLimits = computed(() => concurrencyLimitSubscription.response)
+  const empty = computed(()=> concurrencyLimits.value && !concurrencyLimits.value.length)
```
pleek91, line 55: Usually we check for `loading` and `empty. See [WorkspaceFlows](https://github.com/PrefectHQ/nebula-ui/blob/main/src/pages/WorkspaceFlows.vue#L34)

src/components/ConcurrencyLimitsTable.vue
zhen0: I'm open to either but since the subscription is called here it seemed more logical to add the empty state and associated logic here. I think the pattern of putting the empty state in Nebula-ui and Orion-ui came about because we were doing most api calls there, which increasingly isn't necessary.

pleek91: Yep makes sense

znicholasbrown (approved):

PR #917, opened 2022-12-14
Flow run timeline

brandonreid:
This contains the progress I've made so far. It needs to be broken up, etc. At this stage, @znicholasbrown and I will be putting this in the [graphs repo](https://github.com/PrefectHQ/graphs) and continuing on.


src/components/FlowRunGraph.vue
```diff
@@ -0,0 +1,382 @@
...
+    shoveUntilLayer: (
+      sourceLayerIndex: number,
+      sourceRow: number,
+    ) => void,
+    shoveLayer: (
+      sourceLayerIndex: number,
+      sourceRow: number,
+    ) => void,
+  }
+  const layersEngine: LayersEngine = {
```
brandonreid, line 90: Note that layer related stuff is not used here. It was created for use in the DAG view.

zhen0:
Can we close this one now @brandonreid?

PR #920, opened 2022-12-14
Feature: Create models and API services for WorkerPool

marichka-offen:
This PR adds models and API services for WorkerPool


src/maps/workerFlowRun.ts
```diff
@@ -0,0 +1,20 @@
...
+    flowRun: this.map('FlowRunResponse', source.flow_run, 'FlowRun'),
+  }
+}
+
+export const mapWorkerFlowRunsToWorkerFlowRunsRequest: MapFunction<WorkerFlowRuns, WorkerFlowRunsRequest> = function(source: WorkerFlowRuns): WorkerFlowRunsRequest {
+  return {
+    worker_pool_queue_names: source.workerPoolQueueNames,
+    scheduled_before: source.scheduledBefore,
+    scheduled_after: source.scheduledAfter,
+    limit: source.limit,
```
stackoverfloweth, line 18: This allows you to get around eslint:camelcase without an eslint-ignore. Which I think is preferred because at the file level, it could hide other instances that are avoidable. Usually I find most of my maps that deal with pythonic keys put all the python keys as strings. 

```suggestion
    'worker_pool_queue_names': source.workerPoolQueueNames,
    'scheduled_before': source.scheduledBefore,
    'scheduled_after': source.scheduledAfter,
    'limit': source.limit,
```

src/models/WorkerPoolFilter.ts
```diff
@@ -0,0 +1,13 @@
+type any_ = { any_?: string[] }
+type operator_ = 'and_' | 'or_'
```
stackoverfloweth, line 2: I'm surprised we don't have these defined more globally?

stackoverfloweth: @pleek91, I know you've done a lot with filter types?

src/models/WorkerPoolResponse.ts
```diff
@@ -0,0 +1,13 @@
+import { DateString } from '@/types/dates'
+
+export type WorkerPoolResponse = {
+  id: string,
+  created: DateString,
+  updated: DateString,
+  name: string,
+  description: string | null,
+  type: string |null,
```
pleek91, line 9:
```suggestion
  type: string | null,
```
@stackoverfloweth maybe we can lint this as well?

src/models/WorkerPoolResponse.ts
```diff
@@ -0,0 +1,13 @@
+import { DateString } from '@/types/dates'
+
+export type WorkerPoolResponse = {
```
stackoverfloweth, line 3: this type should probably be in `src/models/api`?

src/services/WorkspaceWorkerPoolApi.ts
```diff
@@ -0,0 +1,52 @@
...
+  getWorkerPoolByName: (workerPoolName: string) => Promise<WorkerPool>,
+  getWorkerPools: (filter: WorkerPoolFilter) => Promise<WorkerPool[]>,
+  updateWorkerPool: (workerPoolName: string, request: WorkerPoolEdit) => Promise<void>,
+  deleteWorkerPool: (workerPoolName: string) => Promise<void>,
+  getWorkerPoolRuns: (workerPoolName: string, request: WorkerFlowRuns) => Promise<WorkerFlowRun[]>,
+}
+
+export class WorkspaceWorkerPoolApi extends WorkspaceApi implements IWorkspaceWorkerPoolApi {
+
+  protected override routePrefix = '/experimental/worker_pools/'
```
pleek91, line 15: Is the `experimental` prefix something specific to worker pool or something we might be seeing in other features?

src/models/WorkerPoolRequest.ts
```diff
@@ -0,0 +1,13 @@
+export type WorkerPoolCreateRequest = Partial<{
```
stackoverfloweth, line 1: these types should probably be in `src/models/api`?

pleek91 (approved):

src/models/WorkerFlowRunsRequest.ts
```diff
@@ -0,0 +1,6 @@
+export type WorkerFlowRunsRequest = Partial<{
```
stackoverfloweth, line 1: this type should probably be in `src/models/api`?

src/models/WorkerFlowRunResponse.ts
```diff
@@ -0,0 +1,7 @@
+import { FlowRunResponse } from '@/models'
+
+export type WorkerFlowRunResponse = {
```
stackoverfloweth, line 3: this type should probably be in `src/models/api`?

src/services/WorkspaceWorkerPoolApi.ts
marichka-offen: It's a worker pool thing for now
Screenshot 2022-12-19 at 9 54 37 AM

src/models/WorkerPoolResponse.ts
stackoverfloweth: https://github.com/PrefectHQ/eslint-config/issues/20

src/models/WorkerPoolFilter.ts
marichka-offen: We have them defined in Union Filter, but they are part of that filter, so exporting them to totally separate filter might be odd. I can extract them in separate PR, they are very useful standalone

src/maps/workerFlowRun.ts
marichka-offen: Copy that! Thanks for pointing that out, I saw both options around and was wondering which one is preferable

src/models/WorkerPoolFilter.ts
pleek91: I believe these exist in the `UnionFilters` file but they are not exported. With the context of a branch I have that redoes all the filter types, I think adding these here self contained is preferred rather than trying to export something from union filters (which is going to be deleted in the near future hopefully).

PR #922, opened 2022-12-15
Chore: fixing bad import from nebula

stackoverfloweth:
image


pleek91:
Yeah.... the import was the way it was to avoid the naming conflict.

stackoverfloweth:
@pleek91 what if we just export the 1 type we need?

pleek91:
Id rather fix the naming collision. Having selective exports in a barrel is confusing and likely someone else will fall into the same issue. But I don't feel strongly.  I don't feel like the import being fixed here is "bad". Though I agree its a sign of an underlying issue. I'll leave it up to you

stackoverfloweth:
@pleek91 I reverted to export *, and fixed the naming collision as we discussed. I will ensure that orion-ui and nebula-ui get updated with this name change as well

SchemaProperty => SchemaPropertyKeyValue
SchemaProperties => SchemaPropertiesKeyValues

PR #928, opened 2022-12-15
Enhancement: Consistently use state names rather than state types

pleek91:
# Description
Converts everything that used state types over to using state names. EXCEPT for the state change stuff. Those specifically use the type only terminal types. Everything else type vs name didn't really matter. 

Visible changes
- Organized the states in the StateNameSelect (which is now used everywhere except the state change modal). 
- Added the TimedOut and Resuming state names to the StateNameSelect

**before**
image

**after**
image

Closes https://github.com/PrefectHQ/prefect/issues/6238


pleek91:
@zhen0 I could use some help QA'ing saved searches with this PR. I need to make sure existing user saved searches still work correctly.

zhen0:
LGTM @pleek91 - I tried creating a saved search in staging and then viewed it through this branch's deploy preview and everything worked as expected.

pleek91:
@zhen0 do you have a workspace that has some saved custom filters? Those are the ones I'm concerned about.

zhen0:
Yep I think my message was unclear. I created saved filters in Eng-and-product in staging (so the workspace did not have these changes). Then viewed that workspace locally using this branch (so did have these changes). The filters worked as expected. Does that make sense?

pleek91:
@zhen0 yes perfect! Also, smart testing. I totally coulda done that myself 😂 I was trying to find a workspace that already had saved filters 🙄

src/maps/savedSearch.ts
```diff
@@ -1,3 +1,4 @@
+import { asArray, mapStateTypeOrNameToStateName } from '..'
```
marichka-offen, line 1: I've never seen an import like this 😯

pleek91: vscode 🤦 fixed!

PR #933, opened 2022-12-19
Feature: WorkerPool create form

marichka-offen:
This PR adds create form to create Worker pools

Screenshot 2022-12-21 at 3 55 52 PM


src/components/WorkerPoolCreateForm.vue
```diff
@@ -0,0 +1,120 @@
...
+  const router = useRouter()
+  const routes = useWorkspaceRoutes()
+
+  const { validate, pending } = useValidationObserver()
+
+  const name = ref('')
+  const description = ref()
+  const type = ref('')
+  const concurrencyLimit = ref()
+  const isPaused = ref()
```
stackoverfloweth, line 64: why do some have default `''` and other `undefined`?

stackoverfloweth: also, the ones that are just `ref()` will be type `Ref<any>`. can you narrow those? `const description = ref<string>()`

src/components/WorkerPoolCreateForm.vue
```diff
@@ -0,0 +1,120 @@
...
+  const isActive = computed({
+    get() {
+      return !isPaused.value
+    },
+    set() {
+      isPaused.value = !isPaused.value
+    },
+  })
```
stackoverfloweth, line 72: this is cleaver, but couldn't you just change your ref above from `isPaused` to `isActive`, and then throw `!` in front of it where you pass to server?

src/components/WorkerPoolCreateForm.vue
```diff
@@ -0,0 +1,120 @@
...
+    router.push(routes.workerPools())
+  }
+  const submit = async (): Promise<void> => {
+    const valid = await validate()
+    if (valid) {
+      const values = {
+        name: name.value,
+        description: description.value,
+        type: type.value,
+        isPaused: isPaused.value,
```
stackoverfloweth, line 94: here is where I'm suggesting `isPaused: !isActive.value`

src/components/WorkerPoolCreateForm.vue
```diff
@@ -0,0 +1,120 @@
...
+  import { useRouter } from 'vue-router'
+  import { SubmitButton, WorkerPoolTypeSelect } from '@/components'
+  import { useWorkspaceApi, useWorkspaceRoutes } from '@/compositions'
+  import { localization } from '@/localization'
+
+  const api = useWorkspaceApi()
+  const router = useRouter()
+  const routes = useWorkspaceRoutes()
+
+  const { validate, pending } = useValidationObserver()
```
stackoverfloweth, line 58: ❤️

src/components/WorkerPoolCreateForm.vue
```diff
@@ -0,0 +1,120 @@
...
+  const rules: Record<string, ValidationRule<string>[]> = {
+    name: [isRequired],
+    type: [isRequired],
+  }
+
+  const { error: nameErrorMessage, state: nameState } = useValidation(name, 'Name', rules.name)
+  const { error: typeErrorMessage, state: typeState } = useValidation(type, 'Type', rules.type)
+
+  function cancel(): void {
+    router.push(routes.workerPools())
```
stackoverfloweth, line 85: I feel like we usually opt for `router.back()` on cancel?

src/components/WorkerPoolCreateForm.vue
```diff
@@ -0,0 +1,120 @@
...
+        type: type.value,
+        isPaused: isPaused.value,
+        concurrencyLimit: concurrencyLimit.value,
+      }
+
+      try {
+        const { name } = await api.workerPools.createWorkerPool(values)
+        showToast(localization.success.createWorkerPool, 'success')
+
+        router.push(routes.workerPool(name))
```
stackoverfloweth, line 102: I'm surprised to see `name` as the prop here not an id?

src/components/WorkerPoolCreateForm.vue
```diff
@@ -0,0 +1,120 @@
...
+        console.error(error)
+      }
+    }
+  }
+</script>
+
+<style>
+.worker-pool-create-form {
+@apply
+  border-[1px]
```
stackoverfloweth, line 114: pretty sure this is the same as 
```suggestion
  border
```

src/components/WorkerPoolTypeSelect.vue
```diff
@@ -0,0 +1,48 @@
...
+</script>
+
+<!--
+  ECS
+  Kubernetes
+  GCP Cloud Run
+  Azure Container
+  Docker Container
+  Process
+-->
```
stackoverfloweth, line 48: intentional?

src/components/WorkerPoolCreateForm.vue
marichka-offen: workerPools do all actions based on their name 🤷‍♀️

src/components/WorkerPoolTypeSelect.vue
marichka-offen: Nope, forgot to remove it

src/components/WorkerPoolCreateForm.vue
marichka-offen: `''` needed for validation to work when we click Create button without touching any fields

stackoverfloweth: if the value was `undefined` it would pass validation?

src/components/WorkerPoolCreateForm.vue
```diff
@@ -0,0 +1,120 @@
...
+      </p-label>
+
+      <p-label label="Type" :state="typeState" :message="typeErrorMessage">
+        <WorkerPoolTypeSelect v-model:selected="type" :state="typeState" />
+      </p-label>
+
+      <p-label label="Status (Optional)">
+        <p-toggle v-model="isActive">
+          <template #append>
+            <div>
```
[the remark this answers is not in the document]
marichka-offen: I copied it from WorkQueuesCreateForm and WorkQueuesEditForm, I'll update it there too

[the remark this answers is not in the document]
marichka-offen: Although we might be getting rid of WorkQueues soon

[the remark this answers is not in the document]
stackoverfloweth: thanks! appreciate you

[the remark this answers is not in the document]
stackoverfloweth: So worker pool is the new work queue??

src/components/WorkerPoolCreateForm.vue
marichka-offen: It was like that, yes, but once I added types to refs just now I get another error. If I don't have `''` in  `const name = ref<string>()`, I get this

Screenshot 2022-12-21 at 4 22 06 PM

src/components/WorkerPoolCreateForm.vue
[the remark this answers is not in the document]
marichka-offen: The worker pool is a container that will have workerPoolQueues in it and will take care of stuff like concurrency setting, health check, and activating stuff, and workerPoolQueues will be like workQueues but simplified afaik

src/components/WorkerPoolCreateForm.vue
stackoverfloweth: those errors are accurate, the only reason you didn't see them before is because the underlying type `any` was swallowing errors. Theoretically you could have assigned a rule that only accepts `Date` for example. Those rules should probably accept `string | undefined` since that's what the values actually are

marichka-offen: I'm not sure what change I need to make here?

src/components/WorkerPoolCreateForm.vue
```diff
@@ -0,0 +1,110 @@
...
+  const name = ref<string>()
+  const description = ref<string>()
+  // Once types feature is implemented by backend, this should be changed to
+  // const type = ref<string>()
+  const type = ref<string>('prefect-agent')
+  const concurrencyLimit = ref<number>()
+  const isActive = ref<boolean>(true)
+  const isActiveLabel = computed(() => isActive.value ? 'Active' : 'Paused')
+
+  const isRequired: ValidationRule<string | undefined> = (value) => value!.trim().length > 0
```
stackoverfloweth, line 64: if value is undefined, this will error. 
```suggestion
  const isRequired: ValidationRule<string | undefined> = (value) => value !== undefined && value.trim().length > 0
```

znicholasbrown (approved):

src/components/WorkerPoolCreateForm.vue
```diff
@@ -0,0 +1,110 @@
+<template>
+  <p-form class="worker-pool-create-form" @submit="submit">
+    <p-content>
+      <p-label label="Name " :message="nameErrorMessage" :state="nameState">
```
pleek91, line 4:
```suggestion
      <p-label label="Name" :message="nameErrorMessage" :state="nameState">
```

src/components/WorkerPoolCreateForm.vue
```diff
@@ -0,0 +1,110 @@
...
+    type: [isRequired],
+  }
+
+  const { error: nameErrorMessage, state: nameState } = useValidation(name, 'Name', rules.name)
+  const { error: typeErrorMessage, state: typeState } = useValidation(type, 'Type', rules.type)
+
+  function cancel(): void {
+    router.back()
+  }
+  const submit = async (): Promise<void> => {
```
pleek91, line 77:
```suggestion
  }

  const submit = async (): Promise<void> => {
```

marichka-offen: Doesn't seem like it, I manually fix linting warnings all the time that are only revealed here in github

pleek91 (approved):

PR #934, opened 2022-12-19
create FlowRunTimeline wrapper component

brandonreid:
Create a component that gets graph data for the timeline graph and renders it.


src/components/FlowRunTimeline.vue
```diff
@@ -0,0 +1,81 @@
...
+    startTime: Date,
+  }
+
+  const graphData = computed(() => {
+    const items: TimelineNodeData[] = (graphSubscription.response ?? [])
+      .filter((node): node is GraphDataTypeGuard => node.startTime !== null)
+      .map(node => {
+        const nodeData = {
+          id: node.id,
+          label: node.id,
```
brandonreid, line 39: I might hold off on merging this until the `label` field has been added to nodes, in which case I'll update this PR.

package.json
```diff
@@ -67,6 +67,7 @@
     "dependencies": {
         "@fontsource/inconsolata": "^4.5.9",
         "@fontsource/inter": "4.5.14",
+        "@prefecthq/graphs": "^0.1.3",
```
pleek91, line 70: Just a note. If you have the `^` you gotta makes sure you stick to semver pretty strictly.

src/components/FlowRunTimeline.vue
```diff
@@ -0,0 +1,81 @@
...
+    flowRun: FlowRun,
+  }>()
+
+  const api = useWorkspaceApi()
+
+  const graphSubscription = useSubscription(api.flowRuns.getFlowRunsGraph, [props.flowRun.id], { interval: 1000 })
+
+  interface GraphDataTypeGuard extends GraphNode {
+    startTime: Date,
+  }
```
pleek91, line 31: You might as well actually use a type guard for this. (untested)

```typescript
function isGraphData<T>(value: T): T & { startTime: Date } {
  return typeof value === 'object' && value !== null && 'startTime' in value && value.startTime instanceof Date
}
```

src/components/FlowRunTimeline.vue
```diff
@@ -0,0 +1,81 @@
...
+
+  const graphData = computed(() => {
```
pleek91, line 33: At some point I'd recommend writing a mapper for this. That way the logic for converting from the api response to a `TimelineNodeData` is abstracted out of this component. 

Either I or @stackoverfloweth would be happy to give you a quick demo of how the mapper works.

src/components/FlowRunTimeline.vue
```diff
@@ -0,0 +1,81 @@
...
+      .map(node => {
+        const nodeData = {
+          id: node.id,
+          label: node.id,
+          start: node.startTime,
+          end: node.endTime,
+          state: node.state?.name.toLowerCase() ?? 'pending',
+        }
+        return nodeData
+      })
```
pleek91, line 45: little more concise and readable IMO

```suggestion
      .map(node => ({
        ...node,
        state: node.state?.name.toLowerCase() ?? 'pending',
      }))
```

src/components/FlowRunTimeline.vue
stackoverfloweth: +1

src/components/FlowRunTimeline.vue
znicholasbrown: https://github.com/PrefectHQ/prefect/pull/7951 - adds names to the `/graph` route, assuming that's what you were looking for

src/components/FlowRunTimeline.vue
```diff
@@ -0,0 +1,81 @@
...
+  import { useWorkspaceApi } from '@/compositions'
+  import { FlowRun, GraphNode } from '@/models'
+
+  const props = defineProps<{
+    flowRun: FlowRun,
+  }>()
+
+  const api = useWorkspaceApi()
+
+  const graphSubscription = useSubscription(api.flowRuns.getFlowRunsGraph, [props.flowRun.id], { interval: 1000 })
```
znicholasbrown, line 27:
```suggestion
  const graphSubscription = useSubscription(api.flowRuns.getFlowRunsGraph, [props.flowRun.id], { interval: 5000 })
```

Polling for this every second is extremely aggressive; I'd recommend you bump this up to at least 5 seconds.

src/components/FlowRunTimeline.vue
```diff
@@ -0,0 +1,81 @@
+<template>
+  <div class="flow-run-timeline-container">
+    <span class="flow-run-timeline-container__beta-badge">
+      Beta
+    </span>
```
znicholasbrown, line 5: @pleek91 or @marichka-offen I thought I saw one of you create a beta badge somewhat recently; was that a one-off? If not we should (in another PR) componentize this so we're not repeating this markup.

pleek91: I created on in nebula. That whole component should probably get copied to orion-design and then removed from nebula.

demo/services/mockWorkspaceFlowRunsApi.ts
```diff
@@ -1,5 +1,5 @@
-import { MockApi } from './MockApi'
-import { FlowRun, GraphNode, RunHistory, StateUpdate } from '@/models'
+import { MockApi } from '@/../demo/services/MockApi'
```
pleek91, line 1: @stackoverfloweth side effect of the new linting rule?

demo/services/mockWorkspaceFlowRunsApi.ts
```diff
@@ -62,4 +66,4 @@ export class MockWorkspaceFlowRunsApi extends MockApi implements IWorkspaceFlowR
     return await this.flowRuns.delete(flowRunId)
   }
 
-}
\ No newline at end of file
+}
```
pleek91, line 69: @stackoverfloweth I feel like I've been seeing this diff come up a lot. Guessing there's a linting rule or a vscode workspace setting for newlines at end of file?

src/components/FlowRunTimeline.vue
pleek91: @brandonreid can you copy the badge from nebula into orion-design and use that here? Just so we can start using it consistently.

znicholasbrown (approved):
Free to merge after resolving any of @pleek91's comments

src/components/FlowRunTimeline.vue
```diff
@@ -0,0 +1,77 @@
...
+    api.flowRuns.getFlowRunsTimeline,
+    [props.flowRun.id],
+    { interval: isRunning.value ? 5000 : undefined },
+  )
+
+  function isValidTimelineNode(value: TimelineNode): value is TimelineNodeData {
+    return typeof value === 'object'
+      && 'startTime' in value
+      && value.startTime instanceof Date
+  }
```
pleek91, line 41: Should have mentioned this earlier. Generally type guards live next to their types. So sometime you might want to move this to wherever TimelineNode is defined.

demo/services/mockWorkspaceFlowRunsApi.ts
stackoverfloweth: 🤷 if there is, it has to be from @typescript-eslint/recommended

src/services/WorkspaceFlowRunsApi.ts
```diff
@@ -55,6 +56,12 @@ export class WorkspaceFlowRunsApi extends WorkspaceApi implements IWorkspaceFlow
     return mapper.map('FlowRunGraphResponse', data, 'GraphNode')
   }
 
+  public async getFlowRunsTimeline(id: string): Promise<TimelineNode[]> {
+    const { data } = await this.get<FlowRunGraphResponse[]>(`/${id}/graph`)
+
+    return mapper.map('FlowRunGraphResponse', data, 'TimelineNode')
+  }
```
pleek91, line 63:
```suggestion
  public async getFlowRunsTimeline(flowRunId: string): Promise<TimelineNode[]> {
    const { data } = await this.get<FlowRunGraphResponse[]>(`/${flowRunId}/graph`)

    return mapper.map('FlowRunGraphResponse', data, 'TimelineNode')
  }
```

demo/services/mockWorkspaceFlowRunsApi.ts
stackoverfloweth: probably. these demos are a bit strange that they have to go up from `@` (src) to get to the demo folder. It's unfortunate but only in demo code and the benefit to production code outweighs IMO

pleek91 (approved):
Approving. Feel free to resolve comments in future PRs.

demo/services/mockWorkspaceFlowRunsApi.ts
stackoverfloweth: we could also disable that rule with a nested .eslintrc in the demo folder

pleek91: might be a good idea. not worried about it, just curious

brandonreid: That might be me actually. My last spot we always had newlines at the end of files, so it's a habit of mine now

pleek91: I don't really care which, but it needs to be configured for everyone via a workspace setting or a linting rule so that way we're not seeing these in PRs. 

Something we can figure out. Don't worry about it.

src/components/FlowRunTimeline.vue
brandonreid: Of course, my bad missen that

PR #944, opened 2022-12-21
Feature: new createCan approach

stackoverfloweth:
this updated approach replaces `getAppPermissions` with `createCan`. The idea is the same that the method will take an argument like `(permission:string) => boolean` and return a `Can` object that looks like `can.create.work_queue`, however the new version uses proxies rather than having physical boolean properties that must be reactive through `watch` and `Object.assign`. The new version also is more dynamic by accepting the permission strings as an argument. This means that, similar to how mapper/mocker work, orion-design can define "root" permissions/flags and nebula/orion can extend those permissions to include additional permissions/flags that only the end application needs to know about. 

this will be used for feature flags, which currently only exist in nebula, but in the near future will also exist in orion.

here is an example of how orion-ui can extend `can` with it's own list of feature flags

```ts
import { Can, createCan as createCanOrionDesign, PermissionCheck, permissions as designPermissions } from '@prefecthq/orion-design'
import { InjectionKey } from 'vue'

const flags = [
  'access:magic',
  'access:wizardry',
  'access:alchemy',
  'access:witchcraft',
] as const

export type Flag = typeof flags[number]
export function isFlag(value: unknown): value is Flag {
  return flags.includes(value as Flag)
}

export const permissions = [...designPermissions, ...flags] as const
export type Permissions = typeof permissions[number]

export function createCan(permissionCheck: PermissionCheck<Permissions>): Can<Permissions> {
  return createCanOrionDesign(permissions, permissionCheck)
}

export const canKey: InjectionKey<Can<Permissions>> = Symbol('canInjectionKey')
```

which can later be provided to application like so

```ts
const can = createCan((permissionString) => {
  if (isFlag(permissionString)) {
    return flags.includes(permissionString)
  }

  // orion-ui still has no concept of permissions, so if it's not a feature flag, it should be enabled
  return true
})

provide(canKey, can)
```

this approach should be completely backwards compatible with previous syntax


src/services/can.ts
```diff
@@ -0,0 +1,142 @@
...
+          const permissionString = `${verb.toString()}:${key.toString()}` as T
+          if (permissions.includes(permissionString)) {
+            return permissionCheck(permissionString)
+          }
+
+          return false
+        },
+      })
+    },
+  })
```
pleek91, line 138: will this be reactive? I thought the benefit of the previous version was that permissions could be loaded async and can would be reactive to that.

stackoverfloweth:
example of implementation in nebula-ui
https://github.com/PrefectHQ/nebula-ui/pull/2136

src/services/can.ts
```diff
@@ -0,0 +1,142 @@
+import { InjectionKey } from 'vue'
+
+const accountPermissions = [
```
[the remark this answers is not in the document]
stackoverfloweth: dude. didn't even think of that. nice

src/services/can.ts
stackoverfloweth: yes? The boolean value is reactive because it's never "set". It's like moving from reactive booleans to a computed getter for each

stackoverfloweth: @pleek91 and I discussed this sync, made some changes, proved that values in can are reactive in nebula when underlying subscriptions update

demo/.eslintrc.json
```diff
@@ -0,0 +1,5 @@
+{
+  "rules": {
+    "no-relative-import-paths/no-relative-import-paths": "off"
```
pleek91, line 3: this shouldn't be here should it?

stackoverfloweth: I can pull it off onto separate PR if you desire

pleek91: I just thought you already merged it. Must have been a different PR. I've looked at to many PRs today lol.

stackoverfloweth: yeah, that was the same change but for prefect-design
https://github.com/PrefectHQ/prefect-design/pull/564

pleek91 (approved):

PR #959, opened 2022-12-22
Feature: Create WorkerPool edit form

marichka-offen:
This PR adds Worker Pool edit header and form

Screenshot 2022-12-22 at 1 08 55 PM


src/components/WorkerPoolEditForm.vue
```diff
@@ -0,0 +1,94 @@
...
+  import { mapProcessTypeValueToProcessTypeLabel } from '@/utilities'
+
+  const props = defineProps<{
+    workerPool: WorkerPool,
+  }>()
+
+  const api = useWorkspaceApi()
+  const router = useRouter()
+  const routes = useWorkspaceRoutes()
+  const { validate, pending } = useValidationObserver()
```
marichka-offen, line 56: Even though we don't use validation here right away, I'm almost certain in the future we'll be able to update `name` which will require validation, so added it in anticipation of a future update

src/components/WorkerPoolEditForm.vue
```diff
@@ -0,0 +1,87 @@
...
+
+  const description = ref<string | null | undefined>(props.workerPool.description)
+  const type = ref<string>(props.workerPool.type!)
```
pleek91, line 59: why the `!`?

src/components/WorkerPoolEditForm.vue
```diff
@@ -0,0 +1,87 @@
...
+  const typeLabel = computed(() => mapProcessTypeValueToProcessTypeLabel(type.value))
```
pleek91, line 60: You might want to add a `typeLabel` getter to your worker pool model that way you don't have to keep computing the label everywhere you use it.

src/components/WorkerPoolEditForm.vue
```diff
@@ -0,0 +1,87 @@
...
+  const concurrencyLimit = ref<number | null | undefined>(props.workerPool.concurrencyLimit)
+  const isActive = ref<boolean | undefined>(!props.workerPool.isPaused)
+  const isActiveLabel = computed(() => isActive.value ? 'Active' : 'Paused')
```
pleek91, line 63: This could be part of the worker pool model as well.

src/components/WorkerPoolEditForm.vue
```diff
@@ -0,0 +1,87 @@
...
+  const api = useWorkspaceApi()
+  const router = useRouter()
+  const routes = useWorkspaceRoutes()
+  const { validate, pending } = useValidationObserver()
+
+  const description = ref<string | null | undefined>(props.workerPool.description)
+  const type = ref<string>(props.workerPool.type!)
+  const typeLabel = computed(() => mapProcessTypeValueToProcessTypeLabel(type.value))
+  const concurrencyLimit = ref<number | null | undefined>(props.workerPool.concurrencyLimit)
+  const isActive = ref<boolean | undefined>(!props.workerPool.isPaused)
```
pleek91, line 62: Same with this. You can make a get/set on the model itself.

src/components/WorkerPoolEditForm.vue
marichka-offen: I'm not sure how these two can be part of model. ` const isActive = ref<boolean | undefined>(props.workerPool.isActive)` just adds to the code but not saves anything 🤔 I need the label to be able change dynamically

stackoverfloweth (approved):

src/components/WorkerPoolEditForm.vue
pleek91: I could be wrong, so don't spend much time trying to get that to work.

PR #962, opened 2022-12-23
Feature: OTable

znicholasbrown:
This PR introduces a new table-wrapping component that incorporates some of the common patterns I've seen floating around the library. It should allow us to cut down on a ton of boilerplate when creating new tables.

The `OTable` component is used in almost the exact same ways as `PTable`, with the except that it requires column headers to be passed. All slots and styling is passed to the wrapped `PTable` component, and controls exist to extend, override, disable, and augment functionality of results, empty state, search, etc.

Note: the demo could use a little work, I only included some pretty basic use cases for the moment.

cc @marichka-offen

### DIFF
--- a/src/components/OTable.vue
+++ b/src/components/OTable.vue
@@ -1,0 +1,222 @@
+<template>
+  <div class="o-table">
+    <div class="o-table__controls-header">
+      <slot name="controls-header">
+        <div class="o-table__controls-header__section o-table__controls-header__section-start">
+          <slot name="controls-header__start">
+            <slot name="controls-header__start-before" />
+
+            <slot name="controls-header__counts">
+              <ResultsCount v-if="_selectedRows.length === 0" :label="rowsCountLabel" :count="_data.length" />
+              <SelectedCount v-else :count="_selectedRows.length" />
+            </slot>
+
+            <slot name="controls-header__start-after" />
+          </slot>
+        </div>
+
+        <div class="o-table__controls-header__section o-table__controls-header__section-end">
+          <slot name="controls__header__end">
+            <slot name="controls-header__end-before" />
+
+
+            <template v-if="!hideSearch">
+              <slot name="controls-header__search">
+                <SearchInput v-model="search" :disabled="disableSearch" :placeholder="searchPlaceholder" :label="searchLabel" />
+              </slot>
+            </template>
+
+            <slot name="controls-header__end-after" />
+          </slot>
+        </div>
+      </slot>
+    </div>
+
+    <p-table v-bind="attrs" :data="_data" :columns="_columns">
+      <template #select-heading>
+        <p-checkbox v-model="select" :disabled="disableSelect" />
+      </template>
+
+      <template #actions-heading>
+        <span />
+      </template>
+
+      <template #select="{ row }">
+        <p-checkbox v-model="_selectedRows" :value="row" :disabled="disableSelect" />
+      </template>
+
+      <slot />
+
+      <template #actions="{ row }">
+        <div class="o-table__row__actions">
+          <slot name="actions" :row="row" />
+        </div>
+      </template>
+
+
+      <template #empty-state>
+        <p-empty-results>
+          <template #message>
+            No {{ emptyResultsLabel }}
+          </template>
+          <template #actions>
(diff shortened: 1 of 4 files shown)


src/components/OTable.vue
```diff
@@ -0,0 +1,245 @@
...
+  const attrs = useAttrs()
+
+  const props = withDefaults(defineProps<{
+    selectedRows?: TableData[],
+    columns: TableColumn[],
+    data: TableData[],
+    emptyResultsLabel?: string,
+    searchPlaceholder?: string,
+    searchLabel?: string,
+    rowsCountLabel?: string,
```
pleek91, line 112: Aren't these likely to be the same? "No things" "Search things" "10 things" etc

pleek91 (commented):
Few thoughts. Overall I like the idea behind unifying and abstracting some of the boilerplate

src/components/OTable.vue
```diff
@@ -0,0 +1,245 @@
...
+    disableSelect?: boolean,
+    disableSearch?: boolean,
+    hideSearch?: boolean,
+    hideSelectColumn?: boolean,
+    hideActionsColumn?: boolean,
+    search?: TableDataFilter,
```
stackoverfloweth, line 118: pleek and I just spent some time talking over something very similar in p-select. We came to the conclusion that passing in `data`, and a method to filter that `data` was not ideal, and that the parent component should do the filtering of `data`. I'm not 100% sure the implications of that change here, maybe this component should v-model the search value or emit "search"?

here's the related PR that removes filter prop from p-select
https://github.com/PrefectHQ/prefect-design/pull/561

src/components/OTable.vue
```diff
@@ -0,0 +1,245 @@
...
+              <ResultsCount v-if="_selectedRows.length === 0" :label="rowsCountLabel" :count="_data.length" />
+              <SelectedCount v-else :count="_selectedRows.length" />
+            </slot>
+
+            <slot name="controls-header__start-after" />
+          </slot>
+        </div>
+
+        <div class="o-table__controls__section o-table__controls__section-end">
+          <slot name="controls__header__end">
```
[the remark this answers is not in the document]
pleek91: I missed that first time through. That's an interesting idea.

src/components/OTable.vue
pleek91: +1 here. I'd rather not "configure" a component but rather express things as code. 

I think search needs to be a v-model anyway to support api search.

src/components/OTable.vue
```diff
@@ -0,0 +1,245 @@
...
+
+    <div class="o-table__controls-footer">
+      <slot name="controls__footer">
+        <div class="o-table__controls__section o-table__controls__section-start">
+          <slot name="controls__footer__start" />
+        </div>
+
+        <div class="o-table__controls__section o-table__controls__section-end">
+          <slot name="controls__footer__end" />
+        </div>
```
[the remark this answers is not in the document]
znicholasbrown: thanks 🤦

src/components/OTable.vue
znicholasbrown: Agreed with this - updated this component to use a model value for search and data to allow API searching/filtering and to emit a `clear` event when the `Clear filters` button is clicked; that will allow consumers to add as many filters as needed and to handle those external to the component.

src/components/OTable.vue
znicholasbrown: Yeah agreed - I consolidated all of these down to a `label` prop except for the search placeholder, which is distinct

src/components/OTable.vue
```diff
@@ -0,0 +1,245 @@
...
+      </slot>
+    </div>
+  </div>
+</template>
+
+<script lang="ts">
+  export default {
+    name: 'OTable',
```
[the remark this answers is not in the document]
znicholasbrown: I'm not against this in the future but I'm sort of on the other side where this isn't sufficiently generic enough for prefect-design and is mostly just a view-layer composition... let's see how this component evolves especially around pagination?

znicholasbrown:
Closing in favor of https://github.com/PrefectHQ/prefect-design/pull/572 and https://github.com/PrefectHQ/prefect-design/pull/573

PR #970, opened 2022-12-27
Feature: Create Workers table

marichka-offen:
This PR adds WorkersTable component and demo to the components library

Screenshot 2022-12-27 at 11 14 32 AM


demo/compositions/useWorkersMock.ts
```diff
@@ -0,0 +1,38 @@
...
+  useSeeds({
+    workerPoolWorkers: [worker],
+    workerPools: [workerPool],
+  })
+
+  return worker
+}
+
+export function useWorkersMock(count: number, override?: () => Partial<WorkerPoolWorker>): WorkerPoolWorker[] {
+  console.log('here')
```
pleek91, line 21:
```suggestion
```

demo/compositions/useWorkersMock.ts
```diff
@@ -0,0 +1,38 @@
...
+    return mocker.create('worker', [
+      {
+        workerPoolId: workerPool.id,
+        ...override?.(),
+      },
+    ])
+  })
+
+  useSeeds({
+    workerPoolWorkers,
```
pleek91, line 34: I think you need to add the `workerPool` here to your seeds.

demo/sections/workers/WorkersTable.vue
```diff
@@ -0,0 +1,18 @@
...
+  import { WorkersTable } from '@/components'
+  import ComponentPage from '@/demo/components/ComponentPage.vue'
+  import { useWorkerPoolMock } from '@/demo/compositions/useWorkerPoolMock'
+  import { useWorkersMock } from '@/demo/compositions/useWorkersMock'
+
+  const workerPool = useWorkerPoolMock()
+
+  useWorkersMock(5, () => ({
+    workerPoolId: workerPool.id,
+  }))
```
pleek91, line 17: I think it be better here to use a `useWorkerPoolMock` which automatically creates some related workers. That way no overrides are needed.

src/components/WorkersTable.vue
```diff
@@ -0,0 +1,121 @@
...
+  const props = defineProps<{
+    workerPoolName: string,
+  }>()
+
+  const api = useWorkspaceApi()
+  const subscriptionOptions = {
+    interval: 30000,
+  }
+
+  const workerPoolNameRef = computed(() => props.workerPoolName)
```
pleek91, line 62: @stackoverfloweth has started using `toRefs` rather than computed props and I kinda like that. You don't have to change this but just raising awareness. 

```typescript
const { workerPoolName } = toRefs(props)
```

src/components/WorkersTable.vue
```diff
@@ -0,0 +1,121 @@
...
+
+  const workerPoolWorkersSubscription = useSubscription(api.workerPoolWorkers.getWorkers, [workerPoolNameRef.value, {}], subscriptionOptions)
```
pleek91, line 64: Is the empty `{}` in the args a filter? If that's the case, we should default that filter to an empty object so that it doesn't have to be passed in everywhere a filter isn't needed.

pleek91 (approved):
Looks pretty good! Couple small comments. Approving as you can resolve them how you see fit.

demo/sections/workers/WorkersTable.vue
marichka-offen: `useWorkerPoolMock` doesn't create workers. I can look into it deeper in separate PR for docs improvement, but for rn purposes I thinks this is fine

src/components/WorkersTable.vue
marichka-offen: Is there a benefite of one over the other?

demo/sections/workers/WorkersTable.vue
pleek91: Yep this works. And being able to do stuff like this makes me feel like the api mocks and compositions are working well. 

Just generally we want to mock the thing the component needs (the worker pool) and have the composition generate any relational data that thing could have. 

No change needed

src/components/WorkersTable.vue
pleek91: No not really. I'd say toRefs is a little more concise especially when there are multiple props. But they both accomplish the same thing. I don't think we have to pick one or the other.

src/components/WorkersTable.vue
```diff
@@ -0,0 +1,121 @@
...
+    return !!searchValue.value
+  })
+
+  const columns = [
+    {
+      property: 'name',
+      label: 'Name',
+    },
+    {
+      label: 'Last Polled',
```
stackoverfloweth, line 88: not sure if you're aware, certainly falls under nit or 2 ways to accomplish the same thing. But FTR, you can have a property that's different from the label and then use the property in the template

```suggestion
      property: 'lastHeartbeatTime',
      label: 'Last Polled',
```

then in the template 
```html
<template #last-polled="{ value }">
  <span>{{ formatDateTimeNumeric(value) }}</span>
</template>
```

src/components/WorkersTable.vue
stackoverfloweth: +1

PR #971, opened 2022-12-27
Enhancement: Update UnionFilter and FlowRunFilteredList to filter by workPool and workPoolQueue

marichka-offen:
This PR adds an upgrade to the UnionFilter to filter runs by by workPool and workPoolQueue


stackoverfloweth (commented):
seems fine to me, but worth getting @pleek91's review since he's changing filters a bit

pleek91 (approved):

PR #976, opened 2022-12-27
Feature: WorkerPoolQueue create form & create page header

marichka-offen:
This PR add workerPoolQueue create form and header for create queue page

Screenshot 2022-12-27 at 4 41 54 PM


src/components/WorkerPoolQueueCreateForm.vue
```diff
@@ -0,0 +1,114 @@
...
+  const { error: nameErrorMessage, state: nameState } = useValidation(name, 'Name', [isRequired])
+  const { error: queuePriorityErrorMessage, state: queuePriorityState } = useValidation(queuePriority, 'Priority', [isGreaterThanZero])
+
+  function cancel(): void {
+    router.back()
+  }
+
+  const submit = async (): Promise<void> => {
+    const valid = await validate()
+    if (valid) {
```
stackoverfloweth, line 83: rather than nesting all of these lines under `if (valid)`, can we just do the inverse? `if (!valid) { return }

stackoverfloweth (approved):

PR #980, opened 2022-12-28
Feature: WorkerPoolQueue page heading

marichka-offen:


src/services/WorkspaceWorkerPoolQueuesApi.ts
```diff
@@ -28,7 +30,7 @@ export class WorkspaceWorkerPoolQueuesApi extends WorkspaceApi implements IWorks
   }
 
   public async getWorkerPoolQueueByName(workerPoolName: string, queueName: string): Promise<WorkerPoolQueue> {
-    const { data } = await this.get<WorkerPoolQueueResponse>(`/${workerPoolName}/queues/${queueName}`)
+    const { data } = await this.get<WorkerPoolQueueResponse>(`/${workerPoolName}/queues/${queueName}/`)
```
marichka-offen, line 33:
```suggestion
    const { data } = await this.get<WorkerPoolQueueResponse>(`/${workerPoolName}/queues/${queueName}`)
```

demo/compositions/useWorkerPoolQueueMock.ts
```diff
@@ -16,6 +17,22 @@ export function useWorkerPoolQueueMock(override?: Partial<WorkerPoolQueue>): Wor
   return workerPoolQueue
 }
 
-export function useWorkerPoolQueuesMock(count: number, override?: Partial<WorkerPoolQueue>): WorkerPoolQueue[] {
-  return repeat(count, () => useWorkerPoolQueueMock(override))
+export function useWorkerPoolQueuesMock(count: number, override?: () => Partial<WorkerPoolQueue>): WorkerPoolQueue[] {
+  const workerPool = useWorkerPoolMock()
```
pleek91, line 21: You probably want to just use the mocker directly here rather than the composition. That composition already inserts a worker pool into the mock database. And the useWorkerPoolMock might have more seeds in the future as well which aren't needed here.

src/components/WorkerPoolQueueToggle.vue
```diff
@@ -0,0 +1,62 @@
...
+  const internalValue = computed({
+    get() {
+      return !props.workerPoolQueue.isPaused
+    },
+    set(value: boolean) {
+      toggle(value)
+    },
+  })
+
+  const state = reactive({ pending: false } as State)
```
pleek91, line 36: Didn't we remove the pending state from another toggle component because it just flashed?

pleek91 (approved):
This looks pretty good to me!

src/components/WorkerPoolQueueToggle.vue
marichka-offen: We did it in some other button, but I agree it's not needed here

PR #981, opened 2022-12-28
Feature: Worker pool queue edit form and edit page header

marichka-offen:
Screenshot 2022-12-28 at 3 54 02 PM


src/components/WorkerPoolQueueEditForm.vue
```diff
@@ -0,0 +1,111 @@
+<template>
+  <p-form class="worker-pool-queue-edit-form" @submit="submit">
+    <p-content>
+      <p-label label="Name" :message="nameErrorMessage" :state="nameState">
+        <p-text-input v-model="name" :state="nameState" />
+      </p-label>
```
stackoverfloweth, line 6: this is a problem in most places, `p-label` needs to tie `id` to the control so that clicking the label focuses the control. 

the easiest way to accomplish this is using the #default template which provides a random `id`

```suggestion
      <p-label label="Name" :message="nameErrorMessage" :state="nameState">
        <template #default="{ id }">
          <p-text-input :id="id" v-model="name" :state="nameState" />
        </template>
      </p-label>
```

stackoverfloweth: this just utilizes base HTML label for attribute
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label#attr-for

src/components/WorkerPoolQueueEditForm.vue
```diff
@@ -0,0 +1,111 @@
...
+    workerPoolName: string,
+    workerPoolQueue: WorkerPoolQueue,
+  }>()
+
+  const api = useWorkspaceApi()
+  const router = useRouter()
+  const routes = useWorkspaceRoutes()
+  const { validate, pending } = useValidationObserver()
+
+
```
stackoverfloweth, line 54:
```suggestion
```

src/components/WorkerPoolQueueEditForm.vue
marichka-offen: Valid observation! I went ahead and added it to all fields in all forms I created for worker feature pages

stackoverfloweth (approved):

PR #983, opened 2022-12-28
Feature: `WorkerPoolQueuesTable`

znicholasbrown:
This PR introduces 2 new components: `WorkerPoolQueuesTable` and `WorkerPoolQueuesDeleteButton. 

It also augments functionality of the `WorkerPoolQueueMenu` component to bar deleting the default queue.


src/components/WorkerPoolQueuesTable.vue
```diff
@@ -0,0 +1,90 @@
...
+    },
+    {
+      label: 'Actions',
+      width: '42px',
+    },
+  ]
+
+  const handleDelete = async (): Promise<void> => {
+    await workerPoolQueuesSubscription.refresh()
+    selected.value = selected.value?.filter(queue => workerPoolQueues.value.find(_queue => _queue.id === queue.id))
```
stackoverfloweth, line 100: not a fan of the underscores, can we do this instead?
```suggestion
    selected.value = selected.value?.filter(queue => workerPoolQueues.value.find(({id}) => id === queue.id))
```

src/components/WorkerPoolQueuesTable.vue
```diff
@@ -0,0 +1,90 @@
...
+  const workerPoolQueues = computed(() => workerPoolQueuesSubscription.response ?? [])
+  const filteredWorkerPoolQueues = computed(() => {
+    if (search.value.length == 0) {
+      return workerPoolQueues.value
+    }
+
+    return workerPoolQueues.value.filter(queue => {
+      const values = Object.values(queue).map(value => value.toString().toLowerCase()).join('')
+      return values.includes(search.value.toLowerCase())
+    })
```
stackoverfloweth: do we not have a service for doing such things?

src/utilities/delete.ts
```diff
@@ -1,11 +1,17 @@
 import { showToast } from '@prefecthq/prefect-design'
 import { localization } from '@/localization'
 
-export type itemType = 'Flow' | 'Deployment' | 'Flow run' | 'Work queue' | 'Block' | 'Notification' | 'Task run' | 'Concurrency Limit' | 'Worker pool'
+export type itemType = 'Flow' | 'Deployment' | 'Flow run' | 'Work queue' | 'Block' | 'Notification' | 'Task run' | 'Concurrency Limit' | 'Worker pool' | 'Worker pool queue'
```
stackoverfloweth, line 4:
```suggestion
export type ItemType = 'Flow' | 'Deployment' | 'Flow run' | 'Work queue' | 'Block' | 'Notification' | 'Task run' | 'Concurrency Limit' | 'Worker pool' | 'Worker pool queue'
```

src/utilities/delete.ts
```diff
@@ -1,11 +1,17 @@
...
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type DeleteItemEndpointArgs = any[] | any
+export type DeleteItemEndpoint = (...args: DeleteItemEndpointArgs[]) => Promise<void>
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export async function deleteItem(args: DeleteItemEndpointArgs, endpoint: DeleteItemEndpoint, type: itemType): Promise<void> {
   try {
-    await endpoint(id)
+    const endpointArgs = Array.isArray(args) ? args : [args]
+    await endpoint(...endpointArgs)
```
znicholasbrown, line 14: This is a small change that should be backward compatible that will allow us to use `deleteItem` for async functions that take more than 1 argument

demo/services/mockWorkspaceWorkerPoolQueuesApi.ts
```diff
@@ -49,8 +49,7 @@ export class MockWorkspaceWorkerPoolQueuesApi extends MockApi implements IWorksp
   public async deleteWorkerPoolQueue(workerPoolName: string, queueName: string): Promise<void> {
     const workerPool = this.workerPools.find(workerPool => workerPool.name === workerPoolName)!
     const workerPoolQueue = this.workerPoolQueues.find(workerPoolQueue => workerPoolQueue.name === queueName && workerPoolQueue.workerPoolId === workerPool.id)!
-
-    return await this.workerPoolQueues.delete(workerPoolQueue.name)
+    return await this.workerPoolQueues.delete(workerPoolQueue.id)
```
znicholasbrown, line 52: This ensures the delete is actually registered by the data service; even though our API deletes by name, the data service uses ids

demo/compositions/useWorkerPoolQueueMock.ts
```diff
@@ -5,12 +5,23 @@ import { repeat } from '@/utilities'
...
+    workerPoolId = override.workerPoolId
+  } else {
+    const workerPool = mocker.create('workerPool')
+    workerPoolId = workerPool.id
+    useSeeds({
+      workerPools: [workerPool],
+    })
+  }
+
+  const workerPoolQueue = mocker.create('workerPoolQueue', [{ ...override, workerPoolId }])
```
znicholasbrown, line 21: This ensures we can associate mock worker pool queues with actual objects when ids are overridden

demo/sections/workers/WorkerPoolTables.vue
```diff
@@ -0,0 +1,25 @@
...
+  import { DemoSection } from '@/demo/types/demoSection'
+
+  const demos: DemoSection[] = [{ title: 'Queues Table' }]
+  const workerPool = useWorkerPoolMock()
+  const workerPools = useWorkerPoolQueuesMock(15, { workerPoolId: workerPool.id })
+
+  workerPool.defaultQueueId = workerPools[0].id
+  useSeeds({
+    workerPools: [workerPool],
+  })
```
znicholasbrown, line 24: This is a little wonky but is needed to ensure the default queue id is associated with an actual object in the data store.

src/utilities/delete.ts
stackoverfloweth: honestly really not a fan of this abstraction even before this change.  Assuming we're gonna keep it, we can make the types a bit more sophisticated

we can accept generic `T` that extends `Action` (from vue-compositions, basically any function), then use Parameters so we only accept the actual params for the function, nothing more nothing less.

```ts
export async function deleteItem<T extends Action>(args: Parameters<T>, endpoint: T, type: ItemType): Promise<void> {
```

if we don't want to have to use an array for args when the action has only a single argument, we can use something like this

```ts
type MaybeSingleParam<T extends Action, Params = Parameters<T>> = Params extends [unknown] ? Params[0] | Params : Params
```

image

here's the whole file
```ts
import { showToast } from '@prefecthq/prefect-design'
import { Action } from '@prefecthq/vue-compositions'
import { asArray } from '@/utilities/arrays'

export type ItemType = 'Flow' | 'Deployment' | 'Flow run' | 'Work queue' | 'Block' | 'Notification' | 'Task run' | 'Concurrency Limit' | 'Worker pool' | 'Worker pool queue'
type MaybeSingleParam<T extends Action, Params = Parameters<T>> = Params extends [unknown] ? Params[0] | Params : Params

export async function deleteItem<T extends Action>(args: MaybeSingleParam<T>, endpoint: T, type: ItemType): Promise<void> {
  try {
    const endpointArgs = asArray(args)
    await endpoint(...endpointArgs)

    showToast(localization.success.delete(type), 'success')
  } catch (error) {
    showToast(localization.error.delete(type.toLowerCase()), 'error')

    console.error(error)
  }
}
```

stackoverfloweth: we could also easily support return types by changing the return type from `Promise<void>` to `Promise<ReturnType<T>>` if you so desire

pleek91:
I think there's a misunderstanding on how the seed compositions are designed thats causing some issues here. A seed composition can seed both the primary entity(ies) and any relational data that a component might need. 

So the `useWorkerPoolMock`  should mock a pool, and then mock some queues. Insert both the pool and the queues as seeds and then return the pool.

The `useWorkerPoolsMock` would probably just repeat the `useWorkerPoolMock`

The `useWorkerPoolQueueMock` would be the opposite. It should mock both a pool and a queue and return the queue. 

The `useWorkerPoolQueuesMock` could mock a small number of pools, and then mock a larger number of queues while picking a random pool id for each queue (from the pools that were just mocked). 

That way when you have a component that accepts a pool or a pool queue. You can call a single composition without having to override anything.

src/components/WorkerPoolQueuesDeleteButton.vue
```diff
@@ -0,0 +1,65 @@
...
+  }>()
+
+  const { showModal, open, close } = useShowModal()
+
+  const api = useWorkspaceApi()
+
+  const deleteWorkerPoolQueues = async (workerPoolQueues: WorkerPoolQueue[]): Promise<void> => {
+    const toastMessage = computed(() => {
+      if (workerPoolQueues.length === 1) {
+        return localization.success.delete('Work queue')
```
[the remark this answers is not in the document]
marichka-offen: No, "work queue" is terminology we're going to remove. "worker pool queue(s)" is the correct text

marichka-offen:
Is it an easy fix to unselect multiselector after everything is deleted?


https://user-images.githubusercontent.com/40467112/209971977-9e078be7-3912-4484-a855-dbbcfdaf8fbd.mov

src/components/WorkerPoolQueuesTable.vue
```diff
@@ -0,0 +1,102 @@
...
+      return workerPoolQueuesData.value
+    }
+
+    return workerPoolQueuesData.value.filter(queue => {
+      const values = Object.values(queue).map(value => value.toString().toLowerCase()).join('')
+      return values.includes(search.value.toLowerCase())
+    })
+  })
+
+  const selected = ref<WorkerPoolQueue[] | undefined>(can.update.worker_pool_queue ? [] : undefined)
```
marichka-offen, line 78: How is the column with selection checkboxes hidden if the person doesn't have delete permissions??

src/components/WorkerPoolQueuesDeleteButton.vue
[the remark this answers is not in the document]
znicholasbrown: Yeah bad copy on my part, fixed that 👍🏻

src/components/WorkerPoolQueuesTable.vue
znicholasbrown: Not as far as I'm aware but I can create something generic if we're doing this elsewhere

src/components/WorkerPoolQueuesTable.vue
znicholasbrown: This works similarly to how our `PSelect` models work; if an array ref is passed, multiselect is enabled, if it isn't, multiselect is hidden.

znicholasbrown:
Looking into that @marichka-offen, that's an issue in prefect-design

src/components/WorkerPoolQueuesTable.vue
znicholasbrown: Added a `hasString` object utility that does this generically, lmk if you had something else in mind

znicholasbrown:
> I think there's a misunderstanding on how the seed compositions are designed thats causing some issues here. A seed composition can seed both the primary entity(ies) and any relational data that a component might need.
> 
> So the `useWorkerPoolMock` should mock a pool, and then mock some queues. Insert both the pool and the queues as seeds and then return the pool.
> 
> The `useWorkerPoolsMock` would probably just repeat the `useWorkerPoolMock`
> 
> The `useWorkerPoolQueueMock` would be the opposite. It should mock both a pool and a queue and return the queue.
> 
> The `useWorkerPoolQueuesMock` could mock a small number of pools, and then mock a larger number of queues while picking a random pool id for each queue (from the pools that were just mocked).
> 
> That way when you have a component that accepts a pool or a pool queue. You can call a single composition without having to override anything.

Thanks for the explanaation @pleek91; given how low impact the subpar usage of those compositions is I'd like us to tackle those updates in a later PR since the compositions existed well before this

pleek91:
> Thanks for the explanaation @pleek91; given how low impact the subpar usage of those compositions is I'd like us to tackle those updates in a later PR since the compositions existed well before this

Totally not blocking. Just wanted to explain how those could work since it seemed like the mock api stuff was slowing you down a bit.

marichka-offen (approved):

PR #988, opened 2022-12-29
Feature: using mapper to translate api filter models to new client filter models

stackoverfloweth:
created client-side models for our filter objects and added mapping profiles to go from our models to api models.

moved api models from `types` to `models/api`

fixed left over exports with previous move of some types into `models/api`

everything that uses filters still uses the api version, as to make this change less risky. But any new reference to filters should be able to construct and pass around nicer looking models like 

```
const filter = {
   id: {any: ['1', '2'],
   expectedStartTime: {after: new Date()},
   ...,
}
```

vs before

```
const filter = {
   id: {any_: ['1', '2'],
   // eslint-ignore
   expected_start_time: {after_: new Date().toISOString()},
   ...,
}
```


znicholasbrown (commented):
This is nice work; if you're in a rush to get this merged I'm happy to approve but do wanna look at it a little more if not

pleek91:
This is going to conflict with https://github.com/PrefectHQ/orion-design/tree/filter-types-and-maps

stackoverfloweth:
closing to avoid conflicts with @pleek91

PR #989, opened 2022-12-29
Feature: Create WorkerPoolQueueUpcomingFlowRunsList component

marichka-offen:
This PR adds a component to render Upcoming flowRuns that belong to certain workerPoolQueue


src/components/WorkerPoolQueueUpcomingFlowRunsList.vue
```diff
@@ -0,0 +1,45 @@
...
+
+  const props = defineProps<{
+    workerPoolName: string,
+    workerPoolQueue: WorkerPoolQueue,
+  }>()
+
+  const api = useWorkspaceApi()
+
+  const { workerPoolName } = toRefs(props)
+  const { workerPoolQueue } = toRefs(props)
```
znicholasbrown, line 32: Couldn't we do:
```suggestion
  const { workerPoolName, workerPoolQueue } = toRefs(props)
```

src/components/WorkerPoolQueueUpcomingFlowRunsList.vue
```diff
@@ -0,0 +1,45 @@
...
+
+  const workerPoolScheduledRunsSubscription = useSubscription(api.workerPools.getWorkerPoolScheduledRuns, [workerPoolName.value, { workerPoolQueueNames: [workerPoolQueue.value.name] }], { interval: 30000 })
```
znicholasbrown, line 34: This interval might be a little long for this - it'd be good to at least match the polling interval, which is 10 seconds

znicholasbrown: Feel free to disregard ^ if that doesn't make sense for worker pools though

src/components/WorkerPoolQueueUpcomingFlowRunsList.vue
```diff
@@ -0,0 +1,45 @@
+<template>
+  <div class="worker-pool-queue-upcoming-flow-runs-list">
+    <FlowRunList :selected="[]" :flow-runs="filteredFlowRuns" disabled />
```
znicholasbrown, line 3: Is it intentional to pass an unreferenced array to the selected attr?

marichka-offen: For whatever reason, we're doing it everywhere we use FlowRunList disabled, and `selected` is a required prop

src/components/WorkerPoolQueueUpcomingFlowRunsList.vue
marichka-offen: Is polling interval always 10 seconds?

src/components/WorkerPoolQueueUpcomingFlowRunsList.vue
znicholasbrown: Hm ok, no reason to block then

znicholasbrown (approved):

PR #993, opened 2023-01-03
Feature: ready to support flags defined in orion-design

stackoverfloweth:
this puts down the framework to define flags in orion-design (as well as in nebula/orion-ui still).

flags can be defined in orion-design if they are planned to be implemented in both nebula and orion-ui

defining a can in nebula will be the combination of

- orion-design "WorkspacePermission"
- orion-design "FeatureFlags"
- nebula-ui "AccountPermission"
- nebula-ui "FeatureFlags"

defining a can in orion-ui will be the combination of 

- orion-design "WorkspacePermission"
- orion-design "FeatureFlags"
- orion-ui "FeatureFlags"

### DIFF
--- a/src/services/can.ts
+++ b/src/services/can.ts
@@ -1,8 +1,8 @@
 import { InjectionKey, ref } from 'vue'
 import { MaybeRef } from '@/types/reactivity'
 
-export const featureFlags = [] as const
-export type FeatureFlags = typeof featureFlags[number]
+export const workspaceFeatureFlags = [] as const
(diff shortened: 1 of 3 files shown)


src/services/can.ts
```diff
@@ -1,6 +1,9 @@
 import { InjectionKey, ref } from 'vue'
 import { MaybeRef } from '@/types/reactivity'
 
+export const featureFlags = [] as const
```
pleek91, line 4: to be consistent with `workspacePermissions` should these all be `workspaceFeatureFlags`?

pleek91 (approved):
Nice

PR #994, opened 2023-01-03
Feature: Flow run history inside the date range picker

stackoverfloweth:
This PR introduces a visual indication of flow run states per date visible on date range picker.

image

the query to flow-run-history is based on the range visible in p-calendar (thanks to exposing viewing-date).  Other filter items, such as state are passed into history query, which effectively also changes which states we expect to see in the calendar.

here's the same snippet with only "pending" state
image

Note, if a given date ever nets >2 state types we slice down to the 2 with the highest `countRuns`. The idea being that even if you have 1 outlier "crashed", if most states are "completed" or "cancelled", you'll only see the green and gray dots for the date. Showing more than the top 2 quickly becomes overwhelming. If the user wants to find outlier "crashed", they should use the states filter.


src/components/DateRangeInputWithFlowRunHistory.vue
```diff
@@ -0,0 +1,110 @@
...
+
+  const internalStartDate = ref<Date | null>(startDate.value)
+  const internalEndDate = ref<Date | null>(endDate.value)
+  const viewingDate = ref<Date>()
+
+  watch([internalStartDate, internalEndDate], ([startDate, endDate]) => {
+    if (startDate && endDate) {
+      updateFilters({ startDate, endDate })
+    }
+  })
```
pleek91, line 46: This is such a nice example of watching an array of values, array destructuring, and object shorthand ❤️

src/components/DateRangeInputWithFlowRunHistory.vue
```diff
@@ -0,0 +1,110 @@
...
+
+  const flowRunsHistorySubscriptionArgs = computed<Parameters<typeof api.flowRuns.getFlowRunsHistory> | null>(() => {
+    if (viewingDate.value === undefined) {
+      return null
+    }
+
+    const monthStart = dateFunctions.startOfMonth(viewingDate.value)
+    const start = dateFunctions.startOfWeek(monthStart)
+    const monthEnd = dateFunctions.endOfMonth(viewingDate.value)
+    const end = dateFunctions.endOfWeek(monthEnd)
```
pleek91, line 56: Seeing this makes me feel like the date picker should v-model both of these rather than a single date. No change necessary. Just thoughts.

src/components/FlowRunsFilterGroup.vue
```diff
@@ -68,4 +58,32 @@
...
+  rounded-full
+}
+
+.flow-runs-filter-group__date--with-completed .flow-runs-filter-group__date-indicator--completed { @apply
+  bg-green-400
+}
+
+.flow-runs-filter-group__date--with-failed .flow-runs-filter-group__date-indicator--failed { @apply
+  bg-red-400
+}
```
pleek91, line 88: Shouldn't these be in the DateRangeInputWithFlowRunHistory component?

src/models/FlowRunHistoryMap.ts
```diff
@@ -0,0 +1,23 @@
+import { RunHistory } from '@/models/RunHistory'
+import { StateHistory } from '@/models/StateHistory'
+import { mapper } from '@/services/Mapper'
+import { DateString } from '@/types/dates'
+import { dateFunctions } from '@/utilities/timezone'
+
+export class FlowRunHistoryMap {
```
pleek91, line 7: This is cool. Maybe something we could make generic in the future. A composition that accepts a reactive array and a method for how to group it that returns a `reactive<Map>`

demo/services/mockWorkspaceFlowRunsApi.ts
```diff
@@ -26,11 +27,27 @@ export class MockWorkspaceFlowRunsApi extends MockApi implements IWorkspaceFlowR
   }
 
   public getFlowRunsHistory(filter: FlowRunsHistoryFilter): Promise<RunHistory[]> {
-    if (Object.keys(filter).length) {
-      console.warn('MockWorkspaceFlowRunsApi has not implemented the filter argument of the getFlowRunsCount method')
+    let start = mapper.map('string', filter.history_start, 'Date')
+    const end = mapper.map('string', filter.history_end, 'Date')
+    const runHistory: RunHistory[] = []
+
+    while (dateFunctions.isBefore(start, end)) {
```
znicholasbrown, line 34: Does `isBefore` also guarantee that passed arguments are both defined? I would assume so

stackoverfloweth: Well the [underlying date-fns method](https://date-fns.org/v2.29.3/docs/isBefore) does not accept `null` or `undefined`, so I imagine there would be an error. I would probably be more concerned with handling that if this weren't in the demo

znicholasbrown: Ah yeah fair enough

src/components/DateRangeInputWithFlowRunHistory.vue
```diff
@@ -0,0 +1,110 @@
...
+      .slice(0, 2)
+  }
+
+  function getRunHistoryKey(date: Date, state: StateType): string {
+    return `${date.toISOString()}-${state}`
+  }
+</script>
+
+<style>
+.date-range-input-with-flow-run-history__date { @apply
```
znicholasbrown, line 89: I wonder if we need a new naming convention lol

znicholasbrown:
Code looks good @stackoverfloweth, this is the only concerning bit:

https://user-images.githubusercontent.com/27291717/210584176-d039be1e-320a-4a34-8c4e-72353647f09e.mov

When a date doesn't have any runs, date numbers become misaligned and move around

src/components/DateRangeInputWithFlowRunHistory.vue
stackoverfloweth: haha, open to suggestions. I generally choose wordy but clear > concise and possibly misunderstood.

stackoverfloweth:
@znicholasbrown looks like just setting an explicit height on each indicators parent fixes it, and has the benefit of pulling in a bit of extra space

https://user-images.githubusercontent.com/6098901/210592975-7e303ab9-5e12-4f01-b48f-3138da93864a.mov

znicholasbrown (approved):
nice @stackoverfloweth 🙏🏻

PR #1000, opened 2023-01-04
Enhancement: Add optional position to schema properties, sort schema form fields

znicholasbrown:
Decided against doing this in the mapper since it would fundamentally change the data being passed around and instead added sorting to the form fields component. 

Resolves: https://github.com/PrefectHQ/prefect/issues/7508

cc @j-tr

### DIFF
--- a/src/components/ParametersTable.vue
+++ b/src/components/ParametersTable.vue
@@ -43,17 +43,20 @@
   const properties = computed(() => props.deployment.parameterOpenApiSchema.properties ?? {})
 
   const data = computed(() => {
-    return Object.entries(properties.value).map(([key, value]) => {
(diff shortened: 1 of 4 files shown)


pleek91:
This LGTM. But this is only the form side. Thoughts on implementing this on the viewing side as well? Specifically `SchemaPropertiesKeyValues`. And possibly for the `ParametersTable`?

znicholasbrown:
Mm good call - I can implement in those as well, give me a sec

znicholasbrown:
Updated - noticed that the typing of the `ParametersTable` component is a little wonky cause of the mapping hack. I'll open a ticket to address that later.

pleek91 (approved):

PR #1007, opened 2023-01-05
Release version 1.1.51

brandonreid:


src/types/index.ts
```diff
@@ -4,7 +4,6 @@ export * from './classes'
 export * from './ColorMode'
 export * from './css'
 export * from './dates'
-export * from '../models/api/LogsRequestFilter'
```
pleek91, line 7: What is this change for?

stackoverfloweth: I asked him to fix duplicate exports. I recognize `LogsRequestFilter`, which I moved out of types and into models/api. Looks like vscode automatically "fixed" this import

stackoverfloweth (approved):

PR #1010, opened 2023-01-05
Enhancement: Add workerPool IconText to flowRun card

marichka-offen:
This PR adds WorkerPool name to the flowRun card. The goal is to show WorkQueue if there's no WorkerPool, and show WorkerPool even if WorkQueue exists

Screenshot 2023-01-05 at 11 27 29 AM


pleek91 (approved):
This all looks good to me. But I feel like we should start breaking this list item into more components. No action needed right this minute. But something we should start considering. This component it's template is starting to get a little hard to read.

PR #1011, opened 2023-01-05
Feature: BlockCapabilityBlockDocumentInput component

pleek91:
# Description
Adds a new component for selecting a block document from a list of block documents with the same capability. Almost identical in appearance and functionality to the BlockDocumentInput component except it accepts a capability rather than a block type. 

image


stackoverfloweth (approved):
can't wait to see better looking filter typesss

pleek91:
> Reference in new

+1

PR #1015, opened 2023-01-05
Enhancement: Improve layout of the BlockCapabilityBlockDocumentInput

pleek91:
# Description
Moves the logo inside the select
Exports the component

image

### DIFF
--- a/src/components/BlockCapabilityBlockDocumentInput.vue
+++ b/src/components/BlockCapabilityBlockDocumentInput.vue
@@ -1,8 +1,15 @@
 <template>
-  <div class="block-capability-block-document-input">
-    <BlockTypeLogo v-if="blockDocument" :block-type="blockDocument.blockType" />
-
-    <BlockDocumentsSelect v-model:selected="internalModalValue" class="block-capability-block-document-input__select" :block-documents="blockDocuments" required />
+  <div class="block-capability-block-document-input" :class="classes" :style="styles" v-bind="listeners">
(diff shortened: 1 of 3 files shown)


marichka-offen (approved):

src/components/BlockCapabilityBlockDocumentInput.vue
```diff
@@ -1,16 +1,32 @@
 <template>
-  <div class="block-capability-block-document-input">
-    <BlockTypeLogo v-if="blockDocument" :block-type="blockDocument.blockType" />
-
-    <BlockDocumentsSelect v-model:selected="internalModalValue" class="block-capability-block-document-input__select" :block-documents="blockDocuments" required />
+  <div class="block-capability-block-document-input" :class="classes" :style="styles" v-bind="listeners">
```
znicholasbrown, line 2:
```suggestion
  <div class="block-capability-block-document-input" :class="classes" :style="styles" v-bind="listeners">
```
Given that this is the root element I'm surprised to see this is necessary; is there something about the `useAttrsStylesClassesAndListeners` component that doesn't allow you just to use/bind the `attrs` down below?

pleek91: We use this when we basically want to pass props/attrs but not styles, classes, or listeners. Specifically used here so `state` can be bound to the select.

znicholasbrown: I missed an extremely critical line of your PR, sorry lol:
```
    inheritAttrs: false,
```

pleek91: lol no worries!

PR #1016, opened 2023-01-06
Enhancement: Add suggestions to tag filter

zhen0:
Uses a combobox for the tag filter to give a list of suggested tags based on the runs that are currently showing (but users can also add their own tags if the list doesn't include the tag they want). 

https://www.loom.com/share/804d038b5bc34a2c85b500655221b4e8

The max number of runs that we're getting tags from here is 200 which means there may be tags that the user does not see suggested.  They can still enter these in the combobox.


src/components/TagCombobox.vue
```diff
@@ -0,0 +1,55 @@
...
+    },
+  })
+
+  const api = useWorkspaceApi()
+  const { filter } = useFlowRunFilterFromRoute()
+
+  const flowRunsSubscription = useSubscription(api.flowRuns.getFlowRuns, [filter])
+  const flowRuns = computed(() => flowRunsSubscription.response ?? [])
+  const tagList = computed(() => flowRuns.value.flatMap(run => run.tags ?? []))
+  const options = computed(() => [...new Set(tagList.value)])
```
pleek91, line 54: Nice use of flatMap and set

stackoverfloweth (approved):
this seems great to me 👍 thanks @zhen0

src/components/TagCombobox.vue
```diff
@@ -0,0 +1,55 @@
...
+      } else if (multiple.value) {
+        emits('update:selected', Array.isArray(value) ? value : [value])
+      } else {
+        emits('update:selected', value)
+      }
+    },
+  })
+
+  const api = useWorkspaceApi()
+  const { filter } = useFlowRunFilterFromRoute()
```
pleek91, line 49: If we make this filter a prop I think there's a few other places we could use this component.

src/components/TagCombobox.vue
```diff
@@ -0,0 +1,55 @@
...
+    set(value: string | string[] | null | undefined) {
+      if (!value) {
+        emits('update:selected', null)
+      } else if (multiple.value) {
+        emits('update:selected', Array.isArray(value) ? value : [value])
+      } else {
+        emits('update:selected', value)
+      }
+    },
+  })
```
pleek91, line 46: I could be wrong but I don't think we actually need this internalModel. Since the combobox is the root element the v-model can just be inherited.

src/components/TagCombobox.vue
```diff
@@ -0,0 +1,55 @@
...
+    <template #default="scope">
+      <slot v-bind="scope" />
+    </template>
+    <template #option="{ option }">
+      <slot name="option" :option="option" />
+    </template>
+  </p-combobox>
+</template>
+
+<script lang="ts" setup>
```
pleek91, line 15: Can we rename this to FlowRunTagCombobox? That way we can also have have tag components for other things.

pleek91 (commented):
Super cool!

zhen0:
Thanks for the comments @pleek91 - are you happy for this to be merged?

PR #1022, opened 2023-01-06
Chore: Updates and clean up to workerPool and poolQueues

marichka-offen:
This PR updates a few features in workerPools and poolQueues:
- adds a toggle queue to the `workPoolQueues` table
- makes late indicator universal for workerPools and poolQueues and renames it `WorkersLateIndicator`
- adds a late indicator to the `workPoolQueues` table
- creates and adds priority description modal to createQueue form priority label, editQueue form priority label, and queue table priority label
- updates of workerPools

Screenshot 2023-01-09 at 9 38 27 AM
Screenshot 2023-01-09 at 9 38 33 AM


src/components/PriorityDescriptionModal.vue
```diff
@@ -0,0 +1,35 @@
...
+    inheritAttrs: false,
+  }
+</script>
+
+<script lang="ts" setup>
+  import { ref, useAttrs } from 'vue'
+
+  const attrs = useAttrs()
+
+  const show = ref(false)
```
stackoverfloweth, line 25: for consistency sake, I'd like to see us use `useShowModal` composition here

src/components/PriorityDescriptionModal.vue
```diff
@@ -0,0 +1,35 @@
+<template>
+  <p-icon class="priority-description-modal__icon" v-bind="attrs" icon="QuestionMarkCircleIcon" solid @click.stop.prevent="show = true" />
```
stackoverfloweth, line 2: idk how aggressively we want to embrace this pattern, but there is undoubtably other places in the app that could use this pattern. What do you think of abstracting this into something like HelpIconModal that takes icon, title, and text description for modal?

src/components/WorkerPoolQueuesTable.vue
```diff
@@ -19,6 +19,13 @@
       </template>
 
       <p-table v-model:selected="selected" :data="filteredWorkerPoolQueues" :columns="columns">
+        <template #priority-heading>
+          <div class="worker-pool-queues-table__priority-heading">
+            Priority
+            <PriorityDescriptionModal />
+          </div>
```
stackoverfloweth, line 26: I see something like this duplicated in a few places, should there just be a `WorkerPoolPriority` component? Might make even more sense if you end up needing a place to implement a more generic `HelpIconModal` component

pleek91: +1 mostly because there's some css that could get removed from these components into the PriorityHeading component.

src/components/PriorityDescriptionModal.vue
marichka-offen: Should the description in `HelpIconModal` be a prop or a slot as well?

marichka-offen: I think I'll turn it into slot in case we want to add styling (bolding some words for example) in the description

src/components/HelpIconModal.vue
```diff
@@ -0,0 +1,54 @@
+<template>
+  <div class="help-icon-modal">
+    {{ title }}
+    <p-icon class="help-icon-modal__icon" v-bind="attrs" :icon="icon" solid @click.stop.prevent="open" />
+
+    <p-modal v-model:show-modal="showModal" :title="title">
+      <slot name="description" />
+
+      <slot name="content" />
```
pleek91, line 9: These can probably both be the default slot

```suggestion
      <slot />
```

src/components/WorkerPoolQueueCreateForm.vue
```diff
@@ -20,6 +20,13 @@
       </p-label>
 
       <p-label label="Priority" :message="queuePriorityErrorMessage" :state="queuePriorityState">
+        <template #label>
+          <HelpIconModal title="Priority" icon="QuestionMarkCircleIcon">
+            <template #description>
+              Priority specifies how Prefect prioritizes the delivery of work — more precisely, execution of flow runs among worker pools. Priority must be a unique, positive integer. Lower numbers indicate higher pool priorities. For example, 1 is the highest priority queue and its flow run execution takes precedence over any lower-priority pool. 10 is a lower priority queue than 1, and 100 is lower priority than both 1 and 10. If you specify a priority that already exists, the new priority supersedes any previously set priority on a worker pool.
+            </template>
+          </HelpIconModal>
```
pleek91, line 28: So I would still create a single component for this called `WorkQueuePriorityLabel` so that we don't duplicate the content in multiple places.

src/components/WorkersLateIndicator.vue
```diff
@@ -0,0 +1,42 @@
...
+
+  const workPoolQueueNames = computed(() => props.workPoolQueueNames ?? [])
+
+  const api = useWorkspaceApi()
+  const { workPoolName } = toRefs(props)
+
+  const workPoolScheduledRunsSubscription = useSubscription(api.workPools.getWorkPoolScheduledRuns, [workPoolName.value, { workPoolQueueNames: workPoolQueueNames.value }], { interval: 30000 })
+  const workPoolScheduledRuns = computed(() => workPoolScheduledRunsSubscription.response ?? [])
+
+  const lateRunsCount = computed(() => workPoolScheduledRuns.value.filter(run => run.flowRun.stateName === 'Late').length)
```
pleek91, line 34: Nit: I'd add a service method called `getWorkPoolLateRuns` that calls the `getWorkPoolScheduledRuns` and filters to late and just returns the late runs. Just to move this logic out of the component.

pleek91 (approved):
One thought but approving as is

PR #1026, opened 2023-01-09
Chore: useFlowRunGraphMock, useTaskRunMock

znicholasbrown:
^Also adds a flow run radar demo

### DIFF
--- a/src/mocks/flowRunGraph.ts
+++ b/src/mocks/flowRunGraph.ts
@@ -7,6 +7,7 @@ export type GraphOptions = {
   shape?: Shape,
   size?: number,
   fanMultiplier?: number,
+  ids?: string[],
 }
 
 export type UpstreamReference = {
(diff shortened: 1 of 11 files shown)


pleek91:
Seeing something in this PR that we haven't done before. Not necessarily against it but wanna make sure I understand what's happening.  You added a `FlowRunGraphMock` type and use that for the seed type. And then in the mock api you use the `get` method to grab an entire graph. 

Assuming I understand that correctly. Is that needed for anything other than convenience/performance of getting a graph? Or is that also because there's not a way to associate graph nodes together?

src/mocks/flowRunGraph.ts
```diff
@@ -15,29 +16,30 @@ export type UpstreamReference = {
...
 
   // Create nodes
+  let i = 0
   while (nodes.length < size) {
-    const target: GraphNode = this.create('graphNode')
+    const target: GraphNode = this.create('graphNode', [{ id: ids[i] }])
 
     const proxy = new Proxy(target, {})
     nodes.push(proxy)
+    ++i
```
pleek91, line 42: What is the proxy for?

znicholasbrown:
> Assuming I understand that correctly. Is that needed for anything other than convenience/performance of getting a graph? Or is that also because there's not a way to associate graph nodes together?

Yup it's exactly the latter; in the API the `/graphs` route returns no associating information at the top level so we'd need to implicitly rely on the shape of its items in a serialization mock like this.

pleek91 (approved):
LGTM

PR #1028, opened 2023-01-09
Feature: hide timestamp from TimezoneSelect by default

stackoverfloweth:
the timezone select uses the append slot to display an example of what the current time would be in the selected timezone. upon closer review, this doesn't make sense for situations where the timezone select is right next to a datepicker. The user would expect to A: be able to edit that time from here. B: expect the time to be the one selected from the other component, not the current time.

before:
image

after: 
image


pleek91 (approved):

zhen0:
Thanks for the quick turnaround on this!

PR #1030, opened 2023-01-10
Chore: CSS cleanup + dark mode support 

znicholasbrown:
For real this time; requires https://github.com/PrefectHQ/prefect-design/pull/600

Note: this will fail build until the next prefect-design release


demo/components/PContextAccordionItem.vue
```diff
@@ -1,153 +0,0 @@
-<template>
-  <div class="p-context-accordion-item">
```
pleek91, line 2: Why was this component deleted?

demo/components/ResizableSection.vue
```diff
@@ -21,32 +21,46 @@
...
+
+
   const classes = computed(() => ({
+    content: {
+      [getColorModeClass(activeColorMode.value)]: true,
+      'dark': colorTheme.value === 'dark',
+      'light': colorTheme.value === 'light',
+      'bg-background-700': colorTheme.value === 'light',
+      'bg-background-400': colorTheme.value === 'dark',
+    },
```
pleek91, line 49:
```suggestion
    content: [getColorModeClass(activeColorMode.value, {
      'dark': colorTheme.value === 'dark',
      'light': colorTheme.value === 'light',
      'bg-background-700': colorTheme.value === 'light',
      'bg-background-400': colorTheme.value === 'dark',
    }],
```

demo/utilities/colorMode.ts
```diff
@@ -0,0 +1,24 @@
+import { useLocalStorage } from '@prefecthq/vue-compositions'
```
pleek91, line 1: This file would make a nice composition. Don't bother but would be cool at some point.

src/components/CodeSnippet.vue
```diff
@@ -38,7 +38,6 @@
 
 <style>
 .code-snippet { @apply
-  bg-slate-700
```
pleek91, line 41: does a code snippet really not need a background?

src/components/FlowRunCreateForm.vue
```diff
@@ -23,13 +23,7 @@
...
 
       <p-label label="Tags (optional)">
-        <p-tags-input v-model="tags" :options="deploymentTags">
-          <template #default="{ selectedOption }">
-            <p-tag :dismissible="!selectedOption.disabled">
-              {{ selectedOption.label }}
-            </p-tag>
-          </template>
-        </p-tags-input>
+        <p-tags-input v-model="tags" :options="deploymentTags" />
```
pleek91, line 26: Ah is this what the update to `p-tags-input` was for?

src/components/FlowsTable.vue
```diff
@@ -197,7 +188,7 @@
 }
 
 .flows-table__activity-chart { @apply
-  h-12
+  !h-12
```
pleek91, line 191: Seems suspicious. Why the `!`?

src/components/FourZeroFour.vue
```diff
@@ -1,38 +1,35 @@
...
....1-3.4v-6.8h-2.6v-2h2.6v-4.4h2.2v4.4h4.8v2h-4.8v6.8c0 .9.2 1.5.6 2 .4.3 1 .5 1.7.5l1-.1a8 8 0 0 0 1.2-.3l.5 1.9a8 8 0 0 1-2.9.5Zm20.1-.2v-17.5H351v2h-9.2v6.8h7.6v2h-7.6v6.7h-2.2Zm21 .2c-1.7 0-3-.4-4-1.2-1-.8-1.5-2-1.5-3.7v-3.6c0-1.7.5-3 1.5-3.8a6 6 0 0 1 4-1.2 6 6 0 0 1 3.9 1.2c1 .9 1.5 2.1 1.5 3.8v3.6c0 1.7-.5 2.9-1.5 3.7-1 .8-2.3 1.2-4 1.2Zm0-2c1 0 1.9-.2 2.4-.6.5-.5.8-1.3.8-2.3v-3.6c0-1-.3-1.8-.8-2.3-.5-.5-1.4-.7-2.5-.7-1 0-1.9.2-2.4.7-.6.5-.8 1.3-.8 2.3v3.6c0 1 .2 1.8.8 2.3.5.4 1.4.7 2.4.7Zm14.2 2a4 4 0 0 1-3-1.2 4.5 4.5 0 0 1-1.2-3.3V492h2.2v8.4c0 1 .2 1.6.6 2.1.4.5 1 .7 2 .7a4 4 0 0 0 2-.5c.6-.4 1.2-1 1.9-1.7v-9h2.2V505h-2.2v-2.2h-.7c-.3.7-.8 1.3-1.5 1.8-.7.4-1.4.6-2.3.6Zm11.3-.2v-13h2.1v2h.7c.3-.6.8-1.2 1.5-1.7.7-.4 1.5-.6 2.3-.6a4 4 0 0 1 3.1 1.2c.8.8 1.2 1.9 1.2 3.3v8.8h-2.2v-8.4c0-1-.3-1.7-.8-2.1-.4-.5-1-.7-1.8-.7s-1.6.2-2.2.6c-.6.4-1.2 1-1.8 1.7v8.9H386Zm20.2-1.9c.8 0 1.5-.1 2-.5.7-.4 1.3-1 2-1.7v-6.5l-1.5-.6a6 6 0 0 0-1.6-.2c-1.1 0-2 .3-2.6.8-.6.5-.9 1.3-.9 2.3v3.5c0 1.2.3 2 .7 2.4a3 3 0 0 0 2 .5Zm3.3-.3c-.4.8-.9 1.4-1.5 1.8-.7.4-1.4.6-2.3.6a4 4 0 0 1-3.1-1.3 5.4 5.4 0 0 1-1.2-3.7v-3.5c0-1.5.5-2.8 1.5-3.7a6 6 0 0 1 4.2-1.3 7.6 7.6 0 0 1 3 .6v-4.8h2.2V505h-2.1v-2.2h-.7Z"
```
pleek91, line 11: lol what was this path? Something you cannot even see?

src/components/FourZeroFour.vue
```diff
@@ -332,32 +329,43 @@
 </script>
 
 <style>
-.not-found {
+.four-zero-four {
   display: flex;
   align-items: center;
   justify-content: center;
-  background:#0a2053;
+}
```
pleek91, line 336: This could be tailwind and combined with the next selector right?

src/components/JsonInput.vue
```diff
@@ -90,7 +90,7 @@
   w-full
   h-full
   min-h-[inherit]
-  bg-slate-700
+  bg-slate-800 /* This should match the background color of the base input component in its darkest mode */
```
pleek91, line 93: Is this needed because the JsonInput is basically always in dark mode?

src/components/LogsContainer.vue
```diff
@@ -62,7 +62,7 @@
   flex
   justify-center
   w-full
-  bg-slate-700
+  bg-slate-800
```
pleek91, line 65: Should the logs container and the JsonInput components have a "lite" version so that way they can just inherit the default light and dark mode colors? Rather than always being "dark"

src/components/RadarNodePlaceholder.vue
```diff
@@ -30,17 +44,18 @@
...
 
 .radar-node-placeholder::after {
   animation: node-placeholder-animation 2s cubic-bezier(0, 0.35, 1, 0.35) infinite;
   background: linear-gradient(
     90deg,
-    rgba(255, 255, 255, 0) 0%,
-    rgba(255, 255, 255, 0.2) 20%,
-    rgba(255, 255, 255, 0.5) 60%,
-    rgba(255, 255, 255, 0.1)
+    hsla(v-bind('style.placeholder'), 0) 0%,
```
pleek91, line 54: the fact that this works is amazing

src/components/TimezoneSelect.vue
```diff
@@ -14,6 +14,7 @@
 
   const props = defineProps<{
     modelValue: string | null,
+    showTimestamp?: boolean,
```
pleek91, line 17: seems like a pretty unrelated change. Was this a bug you stumbled across?

src/components/WorkPoolQueueEditForm.vue
```diff
@@ -107,7 +107,7 @@
 <style>
 .work-pool-queue-edit-form { @apply
   border
-  border-gray-300
+  dark:border-background-600
```
pleek91, line 110: if this was wrapped in a card would this just work?

tsconfig.json
```diff
@@ -28,7 +28,11 @@
     "skipLibCheck": true,
     "sourceMap": true,
     "strict": true,
-    "target": "esnext"
+    "target": "esnext",
+    "types": [
+      "vite-svg-loader",
+      "vite/client"
```
pleek91, line 34: what did adding these fix?

demo/components/PContextAccordionItem.vue
znicholasbrown: It's no longer needed since the fix from https://github.com/PrefectHQ/prefect-design/pull/588 went live; we can use the `PContextAccordionItem` component from `prefect-design` instead of duplicating it now.

src/components/CodeSnippet.vue
znicholasbrown: This is a wrapper for lang-specific code views; those have their own backgrounds already for syntax highlight contrast so this was a duplicate (and subtly incorrect) bg color.

demo/components/PContextAccordionItem.vue
pleek91: ah nice, forgot it was duplicated.

src/components/FlowRunCreateForm.vue
znicholasbrown: Yeah exactly haha

src/components/FlowsTable.vue
znicholasbrown: Honestly might not be needed, the chart component doesn't do a good job growing to allowed space so this was a way to force it to do so. Maybe higher CSS specificity is needed?

src/components/FourZeroFour.vue
znicholasbrown: Nah, this was the text glyphs converted to a single path, which is why this component never responded to global font changes

src/components/FlowsTable.vue
pleek91: Not a big deal. But if there's a specificity issue that would be something to fix possibly. We shouldn't have to worry about specificity especially for something like height.

src/components/FourZeroFour.vue
pleek91: 😂 nice

src/components/FourZeroFour.vue
znicholasbrown: Yeah I think so? I'm not entirely sure why these were written this way, it's possible there's an SVG style scoping issue? I'd rather not dig into it in this PR tbh

src/components/JsonInput.vue
znicholasbrown: Correct - all our custom code snippets are dark themed already, this just brings the styling in line with other components (I've added this comment anywhere I needed to add an explicit bg color)

src/components/LogsContainer.vue
znicholasbrown: I thought about that but since both of those are pretty heavy on the custom styling I figure if there's enough demand we can tackle those later. Personally I kinda like the visual contrast in the light theme

src/components/RadarNodePlaceholder.vue
znicholasbrown: I KNOW. It feels like magic!!

src/components/TimezoneSelect.vue
znicholasbrown: Ah no, it looks like git didn't do a great job fast forwarding this branch on [this commit](https://github.com/PrefectHQ/orion-design/commit/1773abab0ab1431dfcaabc1be3e5f39bc10ed57c)

znicholasbrown: I think it's probably ok though because @stackoverfloweth is still listed as the co-author so it'll just get fast forwarded when we merge

src/components/LogsContainer.vue
pleek91: I do to but only because I want everything dark lol. It would just be nice if these didn't have special styling.

src/components/WorkPoolQueueEditForm.vue
znicholasbrown: Yeah I think there's a small-ish ticket to create a common interface for full-page forms

tsconfig.json
znicholasbrown: Just TS being big mad when working in the vite config and refusing to do any type validation and also allows us to use `import.meta.hot` without adding it to our own meta definition

src/components/FlowsTable.vue
znicholasbrown: Agreed _usually_; these canvas and svg charts are a little weird when it comes to responsiveness

demo/utilities/colorMode.ts
znicholasbrown: Agreed, I think I've created this or something like it in a few places now

demo/components/ResizableSection.vue
znicholasbrown: Ah actually `p-frame` is lazily looking for `Record<string, unknown>` as a `bodyClass` type which means we can't use this syntax

pleek91: 🤔  these should be the same?  
```
{ 
  [getColorModeClass(activeColorMode.value)]: true
}
```
```
[getColorModeClass(activeColorMode.value]
```

pleek91: oh I get your comment now. Looking at that

znicholasbrown: Yeah it was probably cause I wrote the prop that way tbh but I think there's a good reason to; passing around the vue class and style types is really slow for TS cause they're huge

pleek91: Vue has some helper types for this sort of this. Anyway, carry on. I'll create an issue in prefect-design to support vue style dynamic classes

pleek91: https://github.com/PrefectHQ/prefect-design/issues/603

pleek91 (approved):

PR #1033, opened 2023-01-11
bump graphs version and tweek timeline

brandonreid:


src/components/FlowRunTimeline.vue
```diff
@@ -35,7 +35,7 @@
   const graphSubscription = useSubscription(
     api.flowRuns.getFlowRunsTimeline,
     [props.flowRun.id],
-    { interval: isRunning.value ? 5000 : undefined },
+    { interval: 5000 },
   )
```
pleek91, line 39: Why an interval even if it's not running?

pleek91 (approved):

src/components/FlowRunTimeline.vue
brandonreid: The flow run can go from `!isRunning` to `isRunning`, like when it's scheduled and starts running. I could probably do something in the implementation to say "only show if it's running or has ran", but this keeps the implementation simple and sets us up for features we hope to add, such as showing flow run state changes (so you might see the playhead running before the flow runs as it leads up to the scheduled start time).

PR #1041, opened 2023-01-17
Enhancement: Group blocks by type in the BlockCapabilityBlockDocumentInput

pleek91:
# Description
Group the block documents by their block type
image

### DIFF
--- a/src/components/BlockCapabilityBlockDocumentInput.vue
+++ b/src/components/BlockCapabilityBlockDocumentInput.vue
@@ -1,6 +1,6 @@
 <template>
   <div class="block-capability-block-document-input" :class="classes" :style="styles" v-bind="listeners">
-    <BlockDocumentsSelect v-model:selected="internalModalValue" class="block-capability-block-document-input__select" :block-documents="blockDocuments" v-bind="attrs" required>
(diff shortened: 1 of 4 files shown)


stackoverfloweth (approved):

src/components/BlockCapabilityBlockDocumentInput.vue
```diff
@@ -1,6 +1,6 @@
 <template>
   <div class="block-capability-block-document-input" :class="classes" :style="styles" v-bind="listeners">
-    <BlockDocumentsSelect v-model:selected="internalModalValue" class="block-capability-block-document-input__select" :block-documents="blockDocuments" v-bind="attrs" required>
```
znicholasbrown, line 3: Is there a reason not to update the underlying component? just curious

znicholasbrown (approved):

src/components/BlockCapabilityBlockDocumentInput.vue
pleek91: BlockDocumentsSelect is more generic. And currently its only used with a single block type. Technically it could be used with multiple block types. But didn't seem like the complexity was necessary. We can always add it later if we want.

znicholasbrown: fair enough 🤷🏻

PR #1042, opened 2023-01-17
Enhancement: Support displaying an anonymous block in the BlockCapabilityBlockDocumentInput

pleek91:
# Description
Solves an issue with migrating notifications over to automations where notifications use anonymous blocks. We don't want anonymous blocks to show up in the list but if the selected value is an anonymous block we can display it.


znicholasbrown (approved):
lgtm

PR #1044, opened 2023-01-18
BugFix: Hide browser default option when editing deployment schedules

zhen0:
Current behavior:
If I create an interval schedule and select "Use browser default" the timezone is sent as null, which the api interprets as UTC.  The timezone that is sent back is shown as UTC. 

Fix:
This adds the local timezone string to send the correct value to the api. 

I'd appreciate eyes on from @stackoverfloweth here as I know he did a lot of timezone work and I don't want this to mess any of that up!

### DIFF
--- a/src/components/DeploymentDetails.vue
+++ b/src/components/DeploymentDetails.vue
@@ -132,6 +132,7 @@
     updateScheduleLoading.value = true
 
     try {
+      console.log('schedule', schedule)
       await api.deployments.updateDeployment(props.deployment.id, { schedule })
 
       emit('update')
--- a/src/components/DeploymentDetails.vue
+++ b/src/components/DeploymentDetails.vue
@@ -132,7 +132,6 @@
     updateScheduleLoading.value = true
 
     try {
-      console.log('schedule', schedule)
       await api.deployments.updateDeployment(props.deployment.id, { schedule })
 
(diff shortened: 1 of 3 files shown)


src/components/DeploymentDetails.vue
```diff
@@ -132,6 +132,7 @@
     updateScheduleLoading.value = true
 
     try {
+      console.log('schedule', schedule)
```
stackoverfloweth, line 135:
```suggestion
```

stackoverfloweth (commented):
I worry about this change. I'm guessing my example will be true for deployments too, but for sure with user profile settings, if the user chooses "use browser default", they probably will (and should) assume that if they set it to that in New York, then fly to LA the times should all update accordingly. That was true when the value is `null`, but now it's actually setting a specific timezone that happens to be the browsers default at this moment. Happy to chat this through if that isn't making sense.

zhen0:
Thanks for chatting through @stackoverfloweth.  Your point was a good one as we use timezone select in other areas.  For deployments, we decided the option to use browser default was not necessary and was potentially misleading so we've added a prop to hide that option.

stackoverfloweth (approved):
looks good to me 👍 note that this won't be reactive, but that's probably fine for now

PR #1046, opened 2023-01-19
Feature: added line numbers, moved some styles

stackoverfloweth:
the styles I moved (background and padding) are to support prefect-design components like p-code-file exposing a default slot that we can fill with <PythonView> to get syntax highlighting

I also carried over the line-numbers (shout out to @nicolascribbles). Requires passing `show-line-numbers`, currently off by default

### Should `show-line-numbers` be on be default for JsonView? PythonView?

```ts
<CodeSnippet language="python" :snippet="python" show-line-numbers />
```

image

works for JSON as well, though I assume it will be used a lot less often


src/components/JsonView.vue
```diff
@@ -12,11 +14,16 @@
...
+    showLineNumbers?: boolean,
   }>()
 
-  const innerHtml = computed(() => {
-    return highlight(props.value ?? '', languages.json, 'json')
+  const lines = computed(() => {
+    return highlight(props.value ?? '', languages.json, 'json').split('\n')
   })
+
+  const classes = computed(() => ({
```
space-age-pete, line 24: does this need to be a computed?

src/components/PythonView.vue
```diff
@@ -12,11 +14,16 @@
...
+    showLineNumbers?: boolean,
   }>()
 
-  const innerHtml = computed(() => {
-    return highlight(props.value ?? '', languages.python, 'python')
+  const lines = computed(() => {
+    return highlight(props.value ?? '', languages.python, 'python').split('\n')
   })
+
+  const classes = computed(() => ({
```
space-age-pete, line 24: same question as above, does this need to be a computed? does it need to be reactive/need to ever re-evaluate?

space-age-pete (commented):
Looks good!

src/components/JsonView.vue
stackoverfloweth: Probably unlikely but I wouldn't want developers to pass in a reactive boolean and not get the result they expected.

src/components/JsonView.vue
```diff
@@ -35,10 +42,29 @@
...
+  w-full
+  block
+}
+
+.json-view--with-line-numbers .json-view__line:before { @apply
+  inline-block
+  px-3
+  mr-5
+  border-r-[1px]
+  border-slate-200
```
znicholasbrown, line 64:
```suggestion
  border-slate-400
```
Can we make the number border dimmer? It's a tad bright atm (not sure if 400 is the way to go)

src/components/PythonView.vue
```diff
@@ -35,10 +42,29 @@
...
+
+.python-view--with-line-numbers .python-view__line:before { @apply
+  inline-block
+  px-3
+  mr-5
+  border-r-[1px]
+  border-slate-200
+  text-slate-400;
+  counter-increment: line;
+  content: counter(line);
```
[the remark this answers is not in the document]
stackoverfloweth: I like that 👍 esp if we add more "Language"Views in the future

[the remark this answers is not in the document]
pleek91: There's very little difference between PythonView and JsonView other that what language definition they use to highlight. They could probably be combined into a single component without increasing the complexity more than a single "language" prop. And if that's the case then the line numbers wouldn't need their own component. Food for thought.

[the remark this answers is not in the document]
pleek91: Right now the css is really the only reason to have a separate component. But the language css could easily be moved into a separate file.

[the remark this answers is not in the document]
stackoverfloweth: I like this change. Take a look and let me know what you think. cc @znicholasbrown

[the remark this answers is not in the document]
stackoverfloweth: could use someones help to verify that block fields work as expected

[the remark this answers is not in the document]
znicholasbrown: Love it

src/components/CodeHighlighting.vue
```diff
@@ -0,0 +1,58 @@
...
+<style>
+.code-highlighting {
+  -moz-tab-size: 4;
+  -o-tab-size: 4;
+  tab-size: 4;
+
+  -webkit-hyphens: none;
+  -moz-hyphens: none;
+  -ms-hyphens: none;
+  hyphens: none;
```
pleek91, line 49: Quick double check on if we actually need these vendor prefixes.

pleek91 (commented):
This looks so good. Love the change. But how I cannot help but think the PythonInput and JsonInput are almost identical as well lol.

src/components/CodeHighlighting.vue
stackoverfloweth: I'll need @znicholasbrown to help with this one

znicholasbrown: Lol https://developer.mozilla.org/en-US/docs/Web/CSS/hyphens#browser_compatibility

pleek91: Seems like `hyphens` needs the prefixes but I don't think `tab-size` does according to mdn. 
https://developer.mozilla.org/en-US/docs/Web/CSS/tab-size
https://developer.mozilla.org/en-US/docs/Web/CSS/hyphens

znicholasbrown: the mdn hyphen support is nightmarish cause of localization but i think @pleek91 is right that tab size probably doesn't need it (also the vendor prefixes might be added by auto-prefixer regardless so it all might be moot)

pleek91 (approved):

PR #1052, opened 2023-01-21
Update some of the empty state copy

cicdw:
I've been poking around the UI and noticed some of these could use some minor adjustments.


src/utilities/savedFilters.ts
```diff
@@ -56,12 +56,12 @@ export const oneWeekSavedSearch = new SavedSearch({
 
 export const excludeScheduledSavedSearch = new SavedSearch({
   id: null,
-  name: 'No scheduled',
+  name: 'Hide Scheduled Runs',
```
brandonreid, line 59: I could be mistaken, but I believe we use sentence case for all the filters
```suggestion
  name: 'Hide scheduled runs',
```

brandonreid (approved):
One comment but looks a lot better. 💯

stackoverfloweth (approved):

src/components/NotificationsPageEmptyState.vue
```diff
@@ -5,11 +5,11 @@
     </template>
 
     <template #heading>
-      Create a notification to get started
+      Create a notification to begin monitoring your flows
     </template>
 
     <template #description>
-      Send an alert when any flow enters any state you specify.
+      Send an alert when a flow enters a specific state.
```
zhen0, line 12: Send an alert when a run enters a specific state.

zhen0 (commented):
One small suggestion. Thanks for the edits.

src/components/NotificationsPageEmptyState.vue
cicdw: good call, will update

PR #1053, opened 2023-01-22
Work pool tweaks

cicdw:
This PR makes a few minor adjustments to the (still experimental) work pool displays:
- fixes an issue where work pool queues were not displaying at all
- allows for optional queue priority (the API does not require this to be provided)

Note that for optional priority, I tried to adjust the various types to `number | null` similar to concurrency limit, but I couldn't sort it out - note that the API will _always_ return a non-null priority, but you don't need to provide one when creating or editing a work pool queue.


src/components/WorkPoolQueueCreateForm.vue
```diff
@@ -63,7 +63,7 @@
   const isRequired: ValidationRule<string | undefined> = (value) => value !== undefined && value.trim().length > 0
 
   const isGreaterThanZero: ValidationRule<number | undefined> = (value, name) => {
-    if (value && value > 0) {
+    if (value == null || value > 0) {
```
stackoverfloweth, line 66: It looks like this field also is using `isRequired` rule? These seem to conflict and will probably lead to unexpected results. 

I would also like to see a change to rule name to match updated logic.

cicdw: I changed the name - this field didn't use the `isRequired` rule, only the `name` field was using that

stackoverfloweth (approved):

PR #1060, opened 2023-01-24
Chore: Updates permissions for work pool queue components and pages

desertaxle:
The work pool queue components and pages will use the same permissions that work queues currently use. This PR is a small update to use the existing work queue permission for work pool queue pages and components.

### Verification

Verified work pool queue creation, modification, and deletion with `nebula-ui` running against the staging API.


zhen0:
@marichka-offen - I know you're out today so don't feel you need to respond but tagging you here for reference.

stackoverfloweth (approved):
seems fine to me, I'm curious if we plan to remove `work_pool_queue` from permissions entirely? Nebula-ui still maps scopes to `see`, `write`, and `manage_work_pool_queues`

desertaxle:
> I'm curious if we plan to remove `work_pool_queue` from permissions entirely? Nebula-ui still maps scopes to `see`, `write`, and `manage_work_pool_queues`

I'm planning on submitting a PR to `nebula-ui` also to remove the work_pool_queue permissions since they won't be used.

PR #1061, opened 2023-01-24
BugFix: Cannot create block that has a union property of multiple blocks

pleek91:
Fixes an issue with how references for blocks were being guessed based on the the value. Update the ui to always keep the block id and the block type slug together so anywhere in code we always know what type the block is. 

Before
image

After
image

Close https://github.com/PrefectHQ/orion-design/issues/979

### DIFF
--- a/src/utilities/cache.ts
+++ b/src/utilities/cache.ts
@@ -1,5 +1,5 @@
 // this can be used to cache bust between releases. Incrementing this will remove all caches using the previous version
-const cacheVersion = 1
+const cacheVersion = 2
 const cachePrefix = 'cache-key'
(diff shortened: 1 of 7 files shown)


stackoverfloweth (approved):

src/utilities/cache.ts
```diff
@@ -1,5 +1,5 @@
 // this can be used to cache bust between releases. Incrementing this will remove all caches using the previous version
-const cacheVersion = 1
+const cacheVersion = 2
```
znicholasbrown, line 2: at last

znicholasbrown (approved):
lgtm

PR #1063, opened 2023-01-25
Chore: replace DocumentionButton "topic" prop with "to" prop

space-age-pete:
Makes `DocumentationButton` component more generic by allowing string URLs to be passed in rather than keys of the `localization.docs` object. 

Updates all instances of `DocumentationButton` to use the new "to" prop, and pass in the appropriate string const from `localization.docs`

This contains no breaking changes for Orion UI, but does contain breaking changes for Nebula UI

### DIFF
--- a/src/components/BlocksPageEmptyState.vue
+++ b/src/components/BlocksPageEmptyState.vue
@@ -14,7 +14,7 @@
     </template>
 
     <template #actions>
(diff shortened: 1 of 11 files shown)


pleek91 (approved):
Thanks @space-age-pete !

PR #1065, opened 2023-01-26
Enhancement: Add local timezone suggestion when unset and add groups to timezone select

zhen0:
Current: 
Need to know timezone string to set local timezone
image
New: 
Adds a suggested timezone group and all timezone group and adds localtimezone  as a string when unset is passed as a prop. 
image

Closes https://github.com/PrefectHQ/orion-design/issues/1062

### DIFF
--- a/src/components/TimezoneSelect.vue
+++ b/src/components/TimezoneSelect.vue
@@ -49,21 +49,23 @@
   onUnmounted(() => clearTimeout(timeout))
 
   const timezones = Intl.supportedValuesOf('timeZone').map(timezone => ({ label: timezone, value: timezone }))
-  const options: SelectOptionGroup[] = [
-    {
-      label: 'Suggested timezones',
-      options: [{ label: 'UTC', value: utcTimezone }],
-    },
-    {
-      label: 'All timezones',
-      options: [...timezones],
-    },
-  ]
+  const options = computed<SelectOptionGroup[]>(() => {
+    return [
+      {
+        label: 'Suggested timezones',
+        options: [{ label: 'UTC', value: utcTimezone }],
+      },
+      {
+        label: 'All timezones',
+        options: [...timezones],
+      },
+    ]
+  })
 
   const localTimezone = Intl.DateTimeFormat().resolvedOptions().timeZone
   if (!props.hideUnset) {
-    options[0].options.unshift({ label: 'Use browser default', value: null })
+    options.value[0].options.unshift({ label: 'Use browser default', value: null })
   } else {
-    options[0].options.unshift({ label: localTimezone, value: localTimezone })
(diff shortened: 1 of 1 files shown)


src/components/TimezoneSelect.vue
```diff
@@ -49,12 +49,21 @@
...
+    },
   ]
 
+  const localTimezone = Intl.DateTimeFormat().resolvedOptions().timeZone
   if (!props.hideUnset) {
-    options.unshift({ label: 'Use browser default', value: null })
+    options[0].options.unshift({ label: 'Use browser default', value: null })
+  } else {
+    options[0].options.unshift({ label: localTimezone, value: localTimezone })
   }
```
pleek91, line 68: This isn't going to be reactive to `props.hideUnset` and its also a little hard to read. I think making `options` a computed would solve both.

pleek91 (changes requested):

znicholasbrown:
Mostly unrelated to this PR but is there a reason we need to keep machine-readable labels in the time select component? Would be nice to display "America / New York" instead of "America/New_York"

zhen0:
> keep machine-readable labels in the timezone select component? Would be nice to display "America / New York" instead of "America/New_York"

https://github.com/PrefectHQ/orion-design/issues/1066

zhen0:
Updated to a computed (though I'm not sure we'd want this to be reactive) and renamed options to timezoneOptions since the select options group also as an options.

pleek91:
> Updated to a computed (though I'm not sure we'd want this to be reactive) and renamed options to timezoneOptions since the select options group also as an options.

Since `hideUnset` is a prop we should respect that it might change after the component has been mounted. Took some time to write this out. One slight difference is it always has the local timezone as a suggestion. I don't see a reason why we wouldn't want that suggested even if browser default is enabled. 

```typescript
const timezones = Intl.supportedValuesOf('timeZone').map(timezone => ({ label: timezone, value: timezone }))
const localTimezone = Intl.DateTimeFormat().resolvedOptions().timeZone

const timezoneOptions = computed<SelectOptionGroup[]>(() => {
  const suggested: SelectOption[] = [
    { label: localTimezone, value: localTimezone },
    { label: 'UTC', value: utcTimezone },
  ]

  if (!props.hideUnset) {
    suggested.unshift({ label: 'Use browser default', value: null })
  }

  return [
    {
      label: 'Suggested timezones',
      options: suggested,
    },
    {
      label: 'All timezones',
      options: timezones,
    },
  ]
})
```

zhen0:
Ok!  That makes more sense on computed.  I can see value in giving both local timezone and browser default.  Let me update that.  Thanks @pleek91

zhen0:
Ready for review @pleek91

pleek91 (approved):

PR #1067, opened 2023-01-26
Chore: Filter types, maps, and compositions

pleek91:
# Description
Refactor and of all filter models and compositions. 

## Motivation
Working with filters is tedious. Existing UnionFilter types are not quite correct (missing and incorrect properties). And our filter compositions are incomplete (missing filter properties, incorrect filters inherited from UnionFilters, missing compositions, etc).

## Solution
This PR implements the following
- New filter request models to match the API verbatim 
- New filter models that match naming conventions and reduce filter nesting
- New mappers for filter models
- Update service methods to accept new filter models and have default values where missing
- New compositions for each filter. Each filter has a generic filter composition and a composition that automatically stores values in the url query.


zhen0:
Would this close https://github.com/orgs/PrefectHQ/projects/11/views/10?pane=issue&itemId=12924584?

pleek91:
> Would this close https://github.com/orgs/PrefectHQ/projects/11/views/10?pane=issue&itemId=12924584?

It will! Good catch. 

Closes https://github.com/PrefectHQ/orion-design/issues/690

stackoverfloweth (approved):

PR #1068, opened 2023-01-27
BugFix: Add support for withControlled from vee-validate@4.7.0 

zhen0:
Vee validate introduced a method property on handleSubmit which we are wrapping in our useForm composition. This caused a type error because we didn't implement the same type.

Added support for withControlled which fixes the type error.

Replaces #701

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -41,7 +41,7 @@
             "peerDependencies": {
                 "@prefecthq/prefect-design": "^1.2.1",
                 "@prefecthq/vue-compositions": "^1.2.3",
-                "vee-validate": "^4.6.5",
+                "vee-validate": "^4.7.0",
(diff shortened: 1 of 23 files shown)


stackoverfloweth (approved):
@zhen0 so the purpose of this PR is only to support updating to latest vee-validate? I don't see any issue with it, but I wanted to throw out there that we want to remove vee-validate as a dependency soon

pleek91 (approved):

pleek91:
@stackoverfloweth yesg it's just keeping things up to date so dependabot can do it's thing

zhen0:
@stackoverfloweth - yes it's an old PR that was sitting there unreviewed.  Totally fine if we remove vee validate soon.

PR #1069, opened 2023-01-27
Chore: Remove workPool types

marichka-offen:
This PR removes workPool types since we don't want to introduce them to users just yet


pleek91:
@marichka-offen looks like theres some unused variables that eslint is warning about. Can you remove those? Otherwise looks good

marichka-offen:
@pleek91 good catch, removed!

PR #1073, opened 2023-01-28
Enhancement: Add a small message about collections to the blocks page

zhen0:
Adds a link to the collections catalog docs on the blocks catalog page. 

Closes https://github.com/PrefectHQ/prefect/issues/8211
cc @ahuang11 

image


stackoverfloweth (approved):
Thanks @zhen0 this is a great short term solution while we work on a better collections integration

PR #1074, opened 2023-01-28
BugFix: line numbers when > 9 lines

stackoverfloweth:
using v-bind to allocate sufficient space for each digit place in `lines.length`
tested up to 1000 lines, theoretically should work for any size

open to more elegant css solutions if you know of one

before:
Screen Shot 2023-01-28 at 3 03 25 PM

after:
Screen Shot 2023-01-28 at 3 20 13 PM

### DIFF
--- a/src/utilities/math.ts
+++ b/src/utilities/math.ts
@@ -16,6 +16,10 @@ const weightedNumber = (): number => {
   return choice(range(101, 1000))
 }
 
+export function countDigits(value: number): number {
+  return Math.max(Math.floor(Math.log10(Math.abs(value))), 0) + 1
+}
+
 export {
   uniform,
   coinflip,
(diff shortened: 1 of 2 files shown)


src/utilities/math.ts
```diff
@@ -16,6 +16,10 @@ const weightedNumber = (): number => {
   return choice(range(101, 1000))
 }
 
+export function countDigits(value: number): number {
+  return Math.max(Math.floor(Math.log10(Math.abs(value))), 0) + 1
+}
```
pleek91, line 21: This seems overly complex. Wouldn't this do the same?
```
`${value}`.length
```

stackoverfloweth: nice. I like that

pleek91 (approved):

PR #1077, opened 2023-01-30
Feature: Support empirical policy

stackoverfloweth:
added support for Deployment create_flow_run sending empirical policy

image

closes https://github.com/PrefectHQ/prefect/issues/8273


src/components/FlowRunCreateForm.vue
```diff
@@ -151,6 +168,13 @@
 
     emit('submit', resolvedValues)
   })
+
+  watch(addRetries, value => {
+    if (!value) {
+      retries.value = null
+      retryDelay.value = null
+    }
+  })
```
brandonreid, line 177: I wonder if you could just clear the retries, if they have a retry delay on submit just toss it out. That way if I'm fiddling with the retries it doesn't clear my delay I might want to keep. nbd

src/maps/empiricalPolicy.ts
```diff
@@ -11,11 +12,20 @@ export const mapEmpiricalPolicyResponseToEmpiricalPolicy: MapFunction<EmpiricalP
...
     'retry_delay': source.retryDelay,
     'max_retries': source.maxRetries,
     'retry_delay_seconds': source.retryDelaySeconds,
   }
+}
+
+export const mapEmpiricalPolicyToEmpiricalPolicyRequest: MapFunction<EmpiricalPolicy, EmpiricalPolicyRequest> = function(source) {
+  return {
+    'retries': source.retries,
+    'retry_delay': source.retryDelay,
```
brandonreid, line 27: Then this could be like `source.retries ? source.retryDelay : null`

brandonreid (approved):
Couple lint warnings in the diff, one two-part comment there, g2g though

src/maps/empiricalPolicy.ts
stackoverfloweth: interestingly it seems that the API does support sending a `retryDelay` even when `retries` is `null`.. I can't think of a good reason for it to work that way, but I'm tempted to leave as is in case there's some situation where maybe retries is set in code and we just want to set `retryDelay` in the frontend to be something different?

brandonreid: Oh dang, good on ya for investigating that. Yeah that's an interesting sort of developer principle to work out.. Like default to flexible/open, or default to some sort of sanitized outcome. Idk

src/components/FlowRunCreateForm.vue
```diff
@@ -26,6 +26,20 @@
         <p-tags-input v-model="tags" :options="deploymentTags" />
       </p-label>
 
+      <p-label label="Add Retries">
+        <p-toggle v-model="addRetries" />
+      </p-label>
```
pleek91, line 31: Personally I would remove this toggle and just show the retries all the time. I might flex the retry fields so they are formatted together as one thing. 

They're optional and having a toggle to show them doesn't provide much value imo.

src/components/FlowRunCreateForm.vue
pleek91: Removing the toggle means this isn't needed anymore. Simpler might be better

src/maps/deployment.ts
```diff
@@ -47,9 +47,9 @@ export const mapDeploymentUpdateToDeploymentUpdateRequest: MapFunction<Deploymen
   return mapped
 }
 
-export const mapDeploymentFlowRunCreateToDeploymentFlowRunRequest: MapFunction<DeploymentFlowRunCreate, DeploymentFlowRunRequest> = function(source: DeploymentFlowRunCreate): DeploymentFlowRunRequest {
-  const { parameters, state, schema, ...rest } = source
-  const mapped = mapCamelToSnakeCase<DeploymentFlowRunRequest>(rest)
+export const mapDeploymentFlowRunCreateToDeploymentFlowRunRequest: MapFunction<DeploymentFlowRunCreate, DeploymentFlowRunRequest> = function(source) {
+  const { parameters, state, schema, empiricalPolicy, ...rest } = source
+  const mapped = mapCamelToSnakeCase(rest) as DeploymentFlowRunRequest
```
pleek91, line 52: You mentioned wanting to be more explicit in mappers. Feel free to remove this helper in favor of an explicit object map.

src/maps/empiricalPolicy.ts
pleek91: Yeah I'd support what the api supports. Map them both individually.

src/components/FlowRunCreateForm.vue
stackoverfloweth: The code would be simpler but I think the UX is better with the toggle. What do you guys think?

brandonreid: Below it you have the "Start [now | later]" toggle, which I assume shows more form if you select "later", the toggle feels a part of that theme.

It is a little strange to opt in, then have optional fields, at that point those fields are required for the toggle to be truthful. I also wonder if it should be separated out like "Start" is, instead of an "Add Retries" label, make a "Retry on Fail" (or something) header in it's own section. I like what Craig mentioned about making these fields align horizontally, since they both accept a pretty short number.

The Retries and Start parts of the form might be a bit pre-optimized, idk (not sure how much form is behind the Start toggle). I don't feel strongly, but if you made me pick, I'd probably roll with continuing reveal pattern with the design nudges noted above here.

pleek91: The start now/later is a choice the user has to make. If they pick later there are required fields to fill in. 

This toggle is purely for display purposes (the form acts the same with or without it) which is why I'd vote to remove it. 

But I don't feel strongly.

src/maps/deployment.ts
stackoverfloweth: I will do a follow-up PR that removes `mapCamelToSnakeCase`, there are only 6 profiles that use it

stackoverfloweth: https://github.com/PrefectHQ/orion-design/pull/1087

pleek91 (approved):
Nice, looks great

PR #1085, opened 2023-01-30
Chore: sorting tags displayed by p-tags

stackoverfloweth:
as part of [this change](https://github.com/PrefectHQ/prefect-design/pull/623), prefect-design no longer will sort tags for us.


pleek91 (changes requested):
Can we do this in the mapper? So future devs don't have to worry about this.

pleek91 (approved):

PR #1087, opened 2023-01-31
Chore: Removed mapping shortcuts

stackoverfloweth:
Removes `mapCamelToSnakeCase` and `mapSnakeToCamelCase` in favor of explicit mapping of each property.

We have found situations where using this shortcut causes errors

Will follow-up nebula-ui which has 9 files that use these utilities


pleek91:
Looks like this PR has changes from another branch. 

Also, probably want to give everyone a heads up that you're removing these. Just so everyone's aware and does go and try to use them and wonder.

stackoverfloweth:
@pleek91 good catch, that was to try and save the empirical policy in the deployment profile but that can be cherry-picked into the branch that was previously the parent to this one. Should be fixed

znicholasbrown (approved):

PR #1091, opened 2023-01-31
Chore: Cleaned up types

stackoverfloweth:
regex replace
```
MapFunction<(.*?)> = function\(source(:.+)\)(:.+) \{$
MapFunction<$1> = function(source) {
```


space-age-pete:
Is the idea here just that this typing was redundant?

stackoverfloweth:
@space-age-pete exactly, just adds noise

pleek91:
@space-age-pete I believe at one point either because of typescript or eslint (maybe both) the types were necessary. But something got updated and now they're not necessary so nice to be able to remove them. 

thanks @stackoverfloweth !

pleek91 (approved):

PR #1098, opened 2023-02-01
Enhancement: Add an extra info modal component

zhen0:
Screenshot 2023-02-01 at 12 00 33 PM
Screenshot 2023-02-01 at 12 00 24 PM
Uses the code form the RBAC info modal to create a general extra info modal component.  To be used in https://github.com/PrefectHQ/prefect/pull/8295 and https://github.com/PrefectHQ/nebula-ui/pull/2308


src/components/ExtraInfoModal.vue
```diff
@@ -0,0 +1,34 @@
+<template>
+  <button v-bind="attrs" type="button" @click.stop.prevent="show = true">
+    <p-icon icon="QuestionMarkCircleIcon" solid />
```
brandonreid, line 3: A little padding would let the text and ? icon breath, maybe just a `pl-1` or `pl-2` class

brandonreid (approved):
One design nit.

Would be nice to make a tooltip component for this in the future, so that the information is in close proximity to what it's related to (pretty common pattern too)

PR #1108, opened 2023-02-02
Enhancement: Add ability to create a new flow run from the UI with custom default parameter values taken from a previous run

marichka-offen:
This PR adds new functionality to the flowRun page, where we can create even more custom run with already preset parameters of the previous run. It will require some additional changes in orion-ui/nebula-ui, but final result will look like this

**NB!** No changes to `JsonInput` needed 

Screenshot 2023-02-06 at 3 31 31 PM

Addresses https://github.com/PrefectHQ/prefect/issues/7808

Special thanks to @znicholasbrown for showing me how to work with queryParams and @pleek91 for helping with queryParams formatting


src/components/JsonInput.vue
```diff
@@ -28,33 +28,55 @@
...
+          internalValue.value = undefined
+          return
+        }
+
+        internalValue.value = JSON.parse(val)
+      } catch {
+        // do nothing
+      }
+    },
+  })
```
pleek91, line 79: In my experience doing json stringify/parse on a computed like this means that the value you're typing gets formatted as you type causing the cursor to lose its place. There isn't a demo for this component so I cannot verify.

src/components/JsonInput.vue
```diff
@@ -28,33 +28,55 @@
   import { useScrollLinking } from '@/compositions'
   import { stringify } from '@/utilities/json'
 
+  type ModelValue = string | number | boolean | undefined | null
+  type ExtendedModelValue = ModelValue | ModelValue[] | Record<string, ModelValue>
+
   const props = defineProps<{
-    modelValue: string | undefined,
+    modelValue: ExtendedModelValue,
```
pleek91, line 35: Curious why the more complex model? JSON must be a string so I think its understandable for this component to only accept a string. Love to hear your thoughts on this.

src/components/FlowRunCreateForm.vue
```diff
@@ -116,14 +117,22 @@
     start: fieldRules('Start date', isRequiredIf(() => when.value === 'later')),
   }
 
+  const parameters = computed(() => {
+    if (props.flowRun) {
+      return { ...props.deployment.parameters, ...props.flowRun.parameters }
+    }
+
+    return props.deployment.parameters
+  })
```
pleek91, line 126: You can spread `undefined` into an object so you can simplify this
```suggestion
  const parameters = computed(() => {
    return { ...props.deployment.parameters, ...props.flowRun.parameters }
  })
```

pleek91 (commented):
JsonInput is about to get deprecated and replaced by p-code-input. I'm fine with making these changes but I'm curious if anything you ran into here needs consideration in p-code-input. Love to hear what problem you were solving with your changes to the JsonInput component.

src/components/FlowRunMenu.vue
```diff
@@ -1,6 +1,10 @@
 <template>
   <p-icon-button-menu>
     <template #default>
+      <router-link :to="routes.customFlowRunCreate(flowRunId)">
+        <p-overflow-menu-item label="Custom run" />
```
zhen0, line 5: I'd call this "Copy run" or "Duplicate run"

zhen0 (commented):
So fast! Thanks @marichka-offen 

One small naming suggestion.

marichka-offen:
@pleek91 I honestly don't have a good answer to all your questions because I don't fully understand this component, `JsonInput` was giving me trouble while working on a different thing, and @znicholasbrown and @space-age-pete helped me immensely to solve that trouble. They have better understanding to answer why all the changes to `JsonInput`

src/components/FlowRunMenu.vue
pleek91: Or maybe "Run Again"?

pleek91:
> @pleek91 I honestly don't have a good answer to all your questions because I don't fully understand this component, `JsonInput` was giving me trouble while working on a different thing, and @znicholasbrown and @space-age-pete helped me immensely to solve that trouble. They have better understanding to answer why all the changes to `JsonInput`

@znicholasbrown or @space-age-pete can you give some insight on the more complex model type for JsonInput? My concern is this component is going to be deprecated in favor of the p-code-input component and pretty sure that only accepts a string.

src/components/FlowRunMenu.vue
marichka-offen: Both suggestions make sense. I'm trying to think if there are one or two words that can describe running the same thing with different parameters

marichka-offen: ChatGPT suggested 'Parametric execution/run' 😄

zhen0: Can you ask it to try again but give something simpler?

marichka-offen: It's not giving very good ideas.. It also suggested `Adjusted run`

space-age-pete:
@pleek91  I'm not 100% sure. From my recollection it had something to do with having one type for the raw JSON and another for the stringified version of the same value?

pleek91:
> @pleek91 I'm not 100% sure. From my recollection it had something to do with having one type for the raw JSON and another for the stringified version of the same value?

I'd expect this component to not care and only work with strings? Idk, I just wanna make sure we're not doing something that we'll later have to fix when the p-code-input makes it way here.

src/components/FlowRunMenu.vue
zhen0: I vote we go with "Run Again"

marichka-offen: Run Again and Retry next to each other might be confusing together. I chose Custom Run because it matches what deployment does, and technically it takes you to the same page as customDeploymentRun, but with pre-filled params

Screenshot 2023-02-03 at 10 45 48 AM

zhen0: Yeah I get that.  But custom run doesn't explain what it's doing.  How about Duplicate run?

Naming is hard and I think any of these are fine so I'll let you make the final decision here and we can go with it.

pleek91: Its basically the same as "Quick Run" or "Custom Run" on the deployments page. What if we copied the same ui? Made the Retry button have a dropdown same as the run button on a deployment.

image

zhen0: I think we'd have other naming questions then... I vote we keep it simple for now.

zhen0: But again, Marichka can have decision power here.

znicholasbrown:
Sorry for the slow response here, didn't notice the ping! 

I think I misunderstood the context of this change @marichka-offen, my apologies - this doesn't look quite right. Where are we populating the flow run object here? Is that an app-side change? I'm not sure that even makes sense to do (we don't have any checks that it's a flow run owned by or even compatible with that deployment?) conceptually but I would guess the weird behavior we were observing before @marichka-offen has more to do with the pattern you were using to pass data (I don't see any of that in this PR).

More broadly I intended this to be a shim that would allow us to use the `JSONInput` component more flexibly and in line with its name - it's called JSON input but it can only ever return a string, which is only only one of the JSON interfaces.  Given how frequently we use this component I think we do need an abstraction that allows a developer to `v-model` any JSON-compliant object while accounting for user input.

marichka-offen:
@znicholasbrown, why doesn't it look right? Each flow run has `deploymentId`, so we're for sure working with the deployment that created the original flow run. Then based on URL params in orion-ui or nebula-ui, we're coming back to the custom deployment page and concatenate the params of existing deployment and original flow run like this `{ ...props.deployment.parameters, ...props.flowRun.parameters }` in the createCustomRun component. The solution might need to be clarified from orion-design alone, but in conjunction with other repos, it makes sense and works. I'm happy to demo how it looks in both repos simultaneously, and I welcome other solutions for this problem.
The question is, what to do with `JsonInput` since concerns around it are the main blocker for this PR

src/components/FlowRunCreateForm.vue
```diff
@@ -91,12 +91,13 @@
...
   import TimezoneSelect from '@/components/TimezoneSelect.vue'
   import { useForm } from '@/compositions/useForm'
-  import { Deployment, DeploymentFlowRunCreate } from '@/models'
+  import { Deployment, DeploymentFlowRunCreate, FlowRun } from '@/models'
   import { mocker } from '@/services'
   import { fieldRules, isRequiredIf } from '@/utilities/validation'
 
   const props = defineProps<{
     deployment: Deployment,
+    flowRun?: FlowRun,
```
znicholasbrown, line 100:
```suggestion
    parameters?: Deployment['parameters'],
```

src/router/routes.ts
```diff
@@ -88,6 +89,11 @@ export function createWorkspaceRouteRecords(components: Partial<WorkspaceRouteCo
           path: 'task-run/:taskRunId',
           component: components.taskRun,
         },
+        {
+          name: 'workspace.flow-runs.custom-flow-run',
+          path: 'flow-run/:flowRunId/custom',
+          component: components.customFlowRunCreate,
+        },
```
znicholasbrown, line 96:
```suggestion
```

src/router/routes.ts
```diff
@@ -15,6 +15,7 @@ export function createWorkspaceRoutes(config?: CreateWorkspaceRoutesConfig) {
     flowRun: (flowRunId: string) => ({ name: 'workspace.flow-runs.flow-run', params: { flowRunId, ...config } }) as const,
     flowRunRadar: (flowRunId: string) => ({ name: 'workspace.flow-runs.flow-run-radar', params: { flowRunId, ...config } }) as const,
     taskRun: (taskRunId: string) => ({ name: 'workspace.flow-runs.task-run', params: { taskRunId, ...config } }) as const,
+    customFlowRunCreate: (flowRunId: string) => ({ name: 'workspace.flow-runs.custom-flow-run', params: { flowRunId, ...config } }) as const,
```
znicholasbrown, line 18:
```suggestion
deploymentFlowRunCreate: (deploymentId: string, parameters: Record<string, unknown>) => ({ name: 'workspace.deployments.deployment-flow-run-create', params: { deploymentId, ...config }, query: { parameters } }) as const,
```

src/components/FlowRunCreateForm.vue
```diff
@@ -116,14 +117,18 @@
     start: fieldRules('Start date', isRequiredIf(() => when.value === 'later')),
   }
 
+  const parameters = computed(() => {
+    return { ...props.deployment.parameters, ...props.flowRun?.parameters }
```
znicholasbrown, line 121:
```suggestion
    return { ...props.deployment.parameters, ...props.parameters }
```

znicholasbrown (changes requested):
@marichka-offen and i discussed this synchronously and went over a pattern that's a little more flexible and agnostic to this specific motivation (while fulfilling it still). This review reflects what we discussed and not actual PR feedback 👍🏻

pleek91 (approved):

znicholasbrown (approved):
this looks fantastically simple, well done @marichka-offen!

PR #1109, opened 2023-02-02
Bugfix: late indicator showing wrong information

marichka-offen:


src/components/WorkersLateIndicator.vue
```diff
@@ -16,20 +16,41 @@
...
+
+
+  const flowRunFilter = computed<UnionFilters>(() => {
+
+    const flowRunFilter: UnionFilters = {
+      'work_pools': { name: { any_: [workPoolName.value] } },
+      'flow_runs': {
+        state: { name: { any_: ['Late'] } },
+      },
+      sort: 'START_TIME_ASC',
```
stackoverfloweth, line 41: can't wait for @pleek91 to give us a better syntax here 😀

stackoverfloweth (approved):

pleek91:
@marichka-offen code wise looks good. I'm curious though what the `getWorkPoolLateRuns` is for though? This change appears to get late runs for a work pool but using the `getFlowRunsCount` endpoint.

src/components/WorkersLateIndicator.vue
pleek91: Me too! 😂

src/components/WorkersLateIndicator.vue
```diff
@@ -16,20 +16,41 @@
...
     workPoolName: string,
     workPoolQueueNames?: string[],
   }>()
 
-  const workPoolQueueNames = computed(() => props.workPoolQueueNames ?? [])
 
   const api = useWorkspaceApi()
+
   const { workPoolName } = toRefs(props)
+  const workPoolQueueNames = computed(() => props.workPoolQueueNames ?? [])
```
brandonreid, line 31: Not sure you need this computed property? Could just use `props.workPoolQueueNames` directly on like 45, no?

marichka-offen:
@pleek91 `getWorkPoolLateRuns` is supposed to get me those late runs, but the endpoint doesn't seem to work, and I need this fixed asap. So it's either a temporary solution, and I'll catch a backend engineer to help me, or we're going to remove `getWorkPoolscheduled/LateRuns` altogether in favor of flowRun filter since the filter there already has search by workPool/poolQueue

src/components/WorkersLateIndicator.vue
marichka-offen: Good catch!

src/components/WorkersLateIndicator.vue
marichka-offen: It's future Craig's responsibilities

pleek91 (approved):

src/components/WorkersLateIndicator.vue
pleek91: that guy

PR #1112, opened 2023-02-02
Bugfix: Fix FlowRunTagCombobox breaking when using custom values

pleek91:
# Description
Not sure why but removing the slots fixed the issue. There was also something weird going on with the custom values not having quotes around them which a combobox should have. Updated the component to use a p-tags-input rather than a combobox which will hopefully avoid that causing problems in the future. 

Before
image

After
image


stackoverfloweth (approved):

stackoverfloweth:
@pleek91 obviously renaming component means we need to follow up this orion-design release with changes to nebula-ui and orion-ui

PR #1127, opened 2023-02-07
Add paused to cancellable states

cicdw:
- Adds `Paused` to the list of cancellable states.  
- Confirmed that this work with no further changes
- some codebase design feedback: calling these states `stuckStateTypes` with the state type models is confusing; "stuck" needs additional context to be a meaningful name.  Maybe something like `cancellableStateTypes` or `nonTerminalStateTypes` that doesn't require explanation

Closes https://github.com/PrefectHQ/prefect/issues/7807

### DIFF
--- a/src/models/StateType.ts
+++ b/src/models/StateType.ts
@@ -30,7 +30,7 @@ export function isTerminalStateType(value: string): value is TerminalStateType {
   return terminalStateType.includes(value as TerminalStateType)
 }
 
-export const stuckStateTypes = ['running', 'scheduled', 'pending']
+export const stuckStateTypes = ['running', 'scheduled', 'pending', 'paused']
 export type StuckStateType = typeof stuckStateTypes[number]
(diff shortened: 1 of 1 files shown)


pleek91 (approved):
+1 to `nonTerminalStateTypes` being a better name for this

znicholasbrown (approved):

zhen0 (changes requested):
@cicdw - did this work for you locally?

Both paused and cancel use the same button space so I'm concerned that on bigger screens we could end up with the cancel button still not showing.  We could add an option to show cancel in the flowRun menu drop down if the flow run state is paused.

cicdw:
Yup sorry @zhen0 I should have included a screenshot!  Here it is for other's reference (taken on a 14" laptop screen):
image

zhen0:
> Yup sorry @zhen0 I should have included a screenshot! Here it is for other's reference (taken on a 14" laptop screen): image

Fab. Thank you!

PR #1128, opened 2023-02-07
Feature: Batch processing of api requests

pleek91:
# Description
In many places we have lists where each item needs to make some individual requests for more details. For example, the FlowRunListItem makes an individual request for the flow, work pool, work queue, and deployment. Thats 4 additional GET requests per flow run. 

We compensate for this already in two ways
- Subscriptions make sure we're not fetching duplicate data. If all the flow runs are from the same flow only 1 flow GET request will be made
- Visibility observers ensure that we're not fetching data the user can't even see

## Proposal
Batching is the concept of taking multiple individual actions and combining them into a single action. By batching multiple individual GET requests into a single request (using the filter endpoint) we can be much more efficient while also worrying less about how many resources we are requesting. 

## Implementation
This PR introduces a new service called `BatchProcessor`. You can give it a callback for how to handle a batch and then call `batcher.batch(value)`. It will automatically queue batch calls and then flush the queue into the provided callback. 

The callback can return either a `Map` that is keyed by the value or a lookup method. 

## Working Example
This PR includes an example of using the batcher to batch `getFlow` requests. 

Without the batcher
image

With the batcher
image

## Additional Notes
This can be used for more than api requests. We could use it for any repetitive task that can happen many times quickly. For example, when we implement websockets we might use this to batch reacting to individual events to we don't process events to quickly.

### DIFF
--- a/src/services/BatchProcessor/BatchProcessor.ts
+++ b/src/services/BatchProcessor/BatchProcessor.ts
@@ -114,18 +114,18 @@ export class BatchProcessor<V, R> {
     try {
       const response = await this.callback(values)
 
-      if (typeof response === 'function') {
+      if (this.isBatchCallbackLookup(response)) {
         return this.resolveBatchUsingLookup(batch, response)
       }
 
-      return this.resolveBatch(batch, response)
+      return this.resolveBatchUsingMap(batch, response)
 
     } catch (error) {
       this.rejectBatch(batch, error)
     }
   }
 
-  private resolveBatch(batch: BatchQueue<V, R>, map: Map<V, R>): void {
+  private resolveBatchUsingMap(batch: BatchQueue<V, R>, map: Map<V, R>): void {
     batch.forEach(({ resolve }, id) => {
       resolve(map.get(id)!)
     })
@@ -139,6 +139,10 @@ export class BatchProcessor<V, R> {
 
   private rejectBatch(batch: BatchQueue<V, R>, error: unknown): void {
     batch.forEach(({ reject }) => reject(error))
+  }
+
+  private isBatchCallbackLookup(value: ReturnType<BatchCallback<V, R>>): value is BatchCallbackLookup<V, R> {
+    return typeof value === 'function'
   }
 
 }


src/utilities/arrays.ts
```diff
@@ -3,13 +3,12 @@ import { floor, random } from '@/utilities/math'
 
 // we really do want any here
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-export function toRecord<T extends any[], K extends keyof T[number]>(source: T, key: K): Record<K, T> {
```
pleek91, line 6: This utility wasn't being used anywhere and I'm also pretty sure the types are incorrect for it anyway.

src/utilities/arrays.ts
```diff
@@ -3,13 +3,12 @@ import { floor, random } from '@/utilities/math'
...
-  return source.reduce((result, item) => {
-    const itemKey = item[key]
-    result[itemKey] = item
+export function toMap<T extends any[], K extends keyof T[number]>(source: T, key: K): Map<T[number][K], T[number]> {
```
znicholasbrown, line 6: can you copy this utility over to every other repo i want to code in

znicholasbrown (commented):
I'll avoid a full review for now (unless you're looking for one) but this feature is great

src/utilities/arrays.ts
pleek91: 😂 I'll get right on that

pleek91:
> I'll avoid a full review for now (unless you're looking for one) but this feature is great

I'd love a full review on the batcher itself and the syntax for implementing it.

src/services/Batcher.ts
```diff
@@ -0,0 +1,102 @@
+import { MaybePromise } from '@/types'
+
+export type BatchCallbackLookup<V, R> = (value: V) => MaybePromise<R>
+export type BatchCallback<V, R> = (values: V[]) => MaybePromise<Map<V, R>> | BatchCallbackLookup<V, R>
```
znicholasbrown, line 4: Just so I understand this properly; the `MaybePromise<Map<V, R>> | BatchCallbackLookup<V, R>` return value is meant to support a many to one AND a many to many model, right? And the many to many lookup is a `<MaybePromise>` just cause why not?

src/services/Batcher.ts
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    const maxBatchSize = this.options.maxBatchSize ?? Infinity
+
+    if (this.queue.size >= maxBatchSize) {
+      this.processQueue()
+      return
+    }
+
+    const maxCycles = this.options.maxCycles ?? 100
```
znicholasbrown, line 51: It'd be good to see this and the `maxBatchSize` defaults defined on the argument instead. Also: any reason for 100 here?

src/services/Batcher.ts
```diff
@@ -0,0 +1,102 @@
...
+
+    return response
+  }
+
+  private requestBatchProcessOnNextCycle(): void {
+    if (this.timeout) {
+      clearTimeout(this.timeout)
+    }
+
+    const maxBatchSize = this.options.maxBatchSize ?? Infinity
```
znicholasbrown, line 44:
```suggestion
    const { maxBatchSize = Infinity } = this.options
```
I'm surprised eslint isn't warning about this. Maybe the coalesce overrides the rule

src/services/Batcher.ts
```diff
@@ -0,0 +1,102 @@
...
+
+  public constructor(callback: BatchCallback<V, R>, options: BatchOptions = {}) {
+    this.callback = callback
+    this.options = options
+  }
+
+  public batch(value: V): Promise<R> {
+
+    const response = new Promise<R>((resolve, reject) => {
+      this.queue.set(value, { resolve, reject })
```
znicholasbrown, line 31: So since `batch` takes just one argument instead of spreading unknown arguments we're encouraging callbacks that require multiple arguments to be passed as arrays. TIL you can use objects as Map keys but is there a better/clearer pattern here? I was struggling to understand the single call arguments => queue key connection when reading through this

src/services/Batcher.ts
```diff
@@ -0,0 +1,102 @@
...
+    const values = Array.from(batch.keys())
+
+    try {
+      const response = await this.callback(values)
+
+      if (typeof response === 'function') {
+        return this.resolveBatchUsingLookup(batch, response)
+      }
+
+      return this.resolveBatch(batch, response)
```
znicholasbrown, line 79: `resolveBatch` and `resolveBatchUsingLookup` are a little tough to make sense of, maybe there are more descriptive or informative names for these 🤔

znicholasbrown (commented):
Some initial thoughts @pleek91

src/services/WorkspaceFlowRunsApi.ts
```diff
@@ -26,10 +28,20 @@ export class WorkspaceFlowRunsApi extends WorkspaceApi implements IWorkspaceFlow
 
   protected routePrefix = '/flow_runs'
 
-  public async getFlowRun(id: string): Promise<FlowRun> {
-    const { data } = await this.get<FlowRunResponse>(`/${id}`)
+  private readonly batcher = new Batcher<string, FlowRun>(async ids => {
+    const flowRuns = await this.getFlowRuns({
```
znicholasbrown, line 32: technically `getFlowRuns` could be making its _own_ batch call under the hood 😮

src/services/Batcher.ts
```diff
@@ -0,0 +1,102 @@
+import { MaybePromise } from '@/types'
+
+export type BatchCallbackLookup<V, R> = (value: V) => MaybePromise<R>
+export type BatchCallback<V, R> = (values: V[]) => MaybePromise<Map<V, R>> | BatchCallbackLookup<V, R>
+export type BatchOptions = {
+  maxBatchSize?: number,
+  maxCycles?: number,
```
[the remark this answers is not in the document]
pleek91: Ooo I like that.

src/services/Batcher.ts
pleek91: So the map or the lookup can both return the correct value. One to many works either way as long as you set up the map or the lookup correctly. A map is more efficient but the lookup is more flexible. Being able to return an arrow function that knows how to get the value means we can support lots of different use cases.

src/services/Batcher.ts
pleek91: No reason. Those settings are unique to each instance. So I'd like to avoid passing them around.

pleek91: This will change to support ms max

src/services/Batcher.ts
znicholasbrown: Makes sense to me, thanks!

src/services/Batcher.ts
pleek91: I wasn't even considering callbacks with multiple arguments. We can update these easily to support multiple args. Good catch

pleek91: The queue is keyed on the value you're batching. Every time you call batch a new key is added to the queue with an unresolved promise and you get that promise back as the return value from the batch call. 

The callback accepts the value (`V`) that was batched and must return `R`. `R` is used to resolve the promise. 

This will all work the same way for multiple args. But I think the queue will need to be keyed with the batch value stringified. Otherwise if you batch the same value twice you might end up with an unresolved promise that just waits forever.

src/services/Batcher.ts
pleek91: Don't disagree. resolveBatch is used if the callback returns a map and resolveBatchLookup is used if the callback returns a lookup function.

src/services/WorkspaceFlowRunsApi.ts
pleek91: Technically yes. But that would be difficult because of filters. I'm not sure how you would decide if two filters could be combined into one request and then the callback for that Hatcher would be really gross lol.

src/services/Batcher.ts
pleek91: @znicholasbrown thinking more about this. I think a single argument is actually preferred. The batch method is for batching operations, not method calls. If you have an operation you want to perform multiple times you can init a batcher with a callback and then call batch with a single argument. If your operation needs multiple values then you're able to pass an object or an array into the batcher. How the callback processes those values is an implementation detail. 

But I do need to make sure if a value already exists in the map that a promise isn't lost.

pleek91:
@znicholasbrown I made some updates based on your comments.

src/services/Batcher.ts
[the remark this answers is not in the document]
pleek91: Implemented

src/services/Batcher.ts
```diff
@@ -0,0 +1,144 @@
+import { MaybePromise } from '@/types'
+
+export type BatchCallbackLookup<V, R> = (value: V) => MaybePromise<R>
+export type BatchCallback<V, R> = (values: V[]) => MaybePromise<Map<V, R>> | BatchCallbackLookup<V, R>
+export type BatchOptions = {
+  maxBatchSize?: number,
+  maxWait?: number,
```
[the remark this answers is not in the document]
pleek91: Its ms not seconds. I'm of the opinion that in javascript time is always ms. But I don't feel strongly about leaving it off. You'd prefer `maxWaitMilliseconds`?

src/services/Batcher.ts
```diff
@@ -0,0 +1,144 @@
...
+      return
+    }
+
+    this.waitToProcessQueue()
+  }
+
+  private waitToProcessQueue(): void {
+    if (this.waitingSince === null) {
+      this.waitingSince = Date.now()
+    }
```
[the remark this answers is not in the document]
pleek91: So if `waitingSince` is not null that means a queue has started but has not been batched yet. But each time this method is called it sets the timeout again, bumping/delaying the processing of the queue. But each time we "wait" we need to 
- clear the timeout (because we're about to create a new timer
- set `waitingSince` IF this is the first time we've waited

src/services/Batcher.ts
```diff
@@ -0,0 +1,144 @@
...
+  private maxWaitReached(): boolean {
+    const { maxWait = Infinity } = this.options
+    const now = Date.now()
+    const since = this.waitingSince ?? 0
+
+    return now - since >= maxWait
+  }
+
+  private getBatchToProcess(): BatchQueue<V, R> {
+    const batch = new Map(this.queue)
```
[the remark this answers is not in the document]
pleek91: This line and the next line. 

First, create a new map of all the values in the queue
```typescript
const batch = new Map(this.queue)
```

Then clear the queue (removing all the items that were just batched)
```typescript
this.queue.clear()
```

src/services/Batcher.ts
```diff
@@ -0,0 +1,144 @@
...
+  private async processQueue(): Promise<void> {
+    const batch = this.getBatchToProcess()
+    const values = Array.from(batch.keys())
+
+    try {
+      const response = await this.callback(values)
+
+      if (typeof response === 'function') {
+        return this.resolveBatchUsingLookup(batch, response)
+      }
```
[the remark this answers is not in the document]
pleek91: A read me is a great idea. But I'll explain briefly here. The return type for a `BatchCallback` is

```typescript
export type BatchCallbackLookup<V, R> = (value: V) => MaybePromise<R>
export type BatchCallback<V, R> = (values: V[]) => MaybePromise<Map<V, R>> | BatchCallbackLookup<V, R>
```

You can return `Map<V, R>` OR a function that accepts V and returns R. This is for flexibility in implementation. 

If a map is returned, its assumed that for every `V` there will be a key in the map with its `R` as the value. 

If a lookup function is returned. Its assumed that it can be called with `V` and `R` will be returned.

src/services/Batcher.ts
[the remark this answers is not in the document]
pleek91: I renamed it to `maxWaitMilliseconds`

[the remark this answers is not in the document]
stackoverfloweth: I think the fact that we had different assumptions is good enough reason to add the unit

src/services/BatchProcessor/README.md
```diff
@@ -0,0 +1,36 @@
...
+The `BatchProcessor` can be used to queue and batch the processing of values using a single callback. This is useful for repetitive tasks but also useful for combining tasks for efficiency. A common use case for this is batching individual calls to an api endpoint into a single api request. 
+
+A `BatchProcessor` can be created by passing a callback to be used for processing multiple values. The `BatchProcessor` will execute the callback automatically, passing one or more values in to be processed. 
+
+A callback can return either `Map<Value, Response>` or `(value: Value) => Response`. If a map is returned from a callback, it assumes that every value will exist as a key in the map and the response can be found using that key. If a lookup function is returned it will call the lookup function for each value being processed to get the response for each value
+
+## Example 
+```typescript
+const batcher = new BatchProcessor<string, User>(async users => {
+  const users: User[] = await getUsersArray(users)
```
stackoverfloweth, line 11: if I have `users` from the callback, why would we need to go `getUsersArray`? maybe the variable names could be more specific? I'm guessing the thing we got from the callback is actually `userIds`?

pleek91: Correct, messed this up a bit. Updated!

stackoverfloweth (approved):

PR #1130, opened 2023-02-08
Bugfix: Deployment form required parameters

### LINKED ISSUE #1125: Bug: Can't save deployments form with required parameters
The deployment edit form doesn't allow saving when using `Custom` parameters if any of the parameters are required but not supplied. This is incorrect because these are _defaults_ but are only required when actually creating a run.

znicholasbrown:
This PR modifies the deployment model to include the raw schema (returned from the API) in addition to the parsed schema. This is necessary so we can create new schemas on the fly with modified values; in this case I've emptied the top-level property requirements to retain field-level validation for everything except for `isRequired`

Resolves #1125

cc @space-age-pete


stackoverfloweth (approved):

brandonreid (approved):

zhen0:
Thanks @znicholasbrown

pleek91 (approved):
Wish there was a better way but this is less of a hack than anything I had in mind. Nice out of the box thinking here

PR #1131, opened 2023-02-08
Fix: WorkQueues fix for deployments and filters

marichka-offen:
This PR:

- Replaces the `workQueue` combobox with the `workPool` combobox on the flowRuns dashboard for improved filtering capabilities
- Updates the `WorkQueueIconText` component to accept either the pool name or ID as a prop, fixing a broken router link and ensuring navigation to the correct work pool
- Expands the deployment model to now include the work pool name
- Enhances the filter functionality to properly use work pool IDs, names, and queue IDs, allowing for their use as query parameters as needed
- As a bonus, the changes to filters have also fixed the flowRuns filtering for the "Upcoming" and "Runs" tabs on the `workPoolQueue` page

**NB!** After these changes, orion-design no longer uses `workQueueCombobox`. I have not removed it though, because I see it's used in `AutomationWorkQueueCombobox` in nebula-ui. Since the entire issue was that `workQueueCombobox` wasn't specific enough and let you select several queues with the same name, I wonder if it's possible in a clean way to replace it with `WorkPoolQueueCombobox` (requires pool name) or update it specifically for automation to base on id rather than the name. cc @pleek91


src/components/DeploymentDetails.vue
```diff
@@ -6,9 +6,15 @@
...
-        <WorkQueueIconText :work-queue-name="deployment.workQueueName!" />
+    <p-key-value v-if="deployment.workPoolName" label="Work Pool" :alternate="alternate">
+      <template #value>
+        <WorkPoolIconText :work-pool-name="deployment.workPoolName!" />
+      </template>
+    </p-key-value>
+
+    <p-key-value v-if="deployment.workQueueName" label="Work Queue" :alternate="alternate">
+      <template #value>
+        <WorkQueueIconText :work-queue-name="deployment.workQueueName!" :work-pool-name="deployment.workPoolName" />
```
znicholasbrown, line 17: Is it possible that both work queue name and work pool name are set? If so, should we show both? I'm not sure I understand the implications of that scenario

src/components/WorkPoolCombobox.vue
```diff
@@ -0,0 +1,67 @@
...
+      value: workPool.name,
+      label: workPool.name,
+    }))
+
+    if (props.allowUnset) {
+      options.unshift({
+        value: null,
+        label: 'None',
+      })
+    }
```
znicholasbrown, line 63: This is clever

src/components/WorkPoolQueueCombobox.vue
```diff
@@ -0,0 +1,77 @@
+<template>
+  <p-combobox v-model="internalValue" :options="options" :empty-message="emptyMessage">
+    <template #combobox-options-empty>
+      No work queues
```
znicholasbrown, line 4:
```suggestion
      No work pool queues
```
Not sure if this is correct

src/components/WorkQueueIconText.vue
```diff
@@ -14,29 +14,48 @@
...
 
-    return [{ 'work_pools': { id: { any_: [workQueue.value.workPoolId] } } }]
+    return [{ 'work_pools': { id: { any_: [props.workPoolId] } } }]
+
   })
 
   const workPoolsSubscription = useSubscriptionWithDependencies(api.workPools.getWorkPools, workPoolArgs)
   const workPools = computed(() => workPoolsSubscription.response ?? [])
-  const workPool = computed(() => workPools.value[0])
+  const workPoolName = computed(() => workPools.value[0].name)
```
space-age-pete, line 41: This could maybe be simplified so that you wouldn't need a separate `workPoolNameRef` variable
```suggestion
  const workPoolName = computed(() => pros.workPoolName ?? workPools.value[0].name)
```

src/components/WorkPoolCombobox.vue
```diff
@@ -0,0 +1,67 @@
...
+    selected: string | string[] | null | undefined,
+    emptyMessage?: string,
+    allowUnset?: boolean,
+  }>()
+
+  const emits = defineEmits<{
+    (event: 'update:selected', value: string | string[] | null): void,
+  }>()
+
+  const multiple = computed(() => Array.isArray(props.selected))
```
znicholasbrown, line 31: The behavior for the `p-select` (which I think is what `p-combobox` uses under the hood) multiple attribute changed a bit recently - is this still necessary? I was under the impression it wasn't but I haven't worked with this much

src/components/DeploymentDetails.vue
```diff
@@ -6,9 +6,15 @@
...
-    <p-key-value label="Work Queue" :alternate="alternate">
-      <template v-if="deployment.workQueueName" #value>
-        <WorkQueueIconText :work-queue-name="deployment.workQueueName!" />
+    <p-key-value v-if="deployment.workPoolName" label="Work Pool" :alternate="alternate">
+      <template #value>
+        <WorkPoolIconText :work-pool-name="deployment.workPoolName!" />
+      </template>
+    </p-key-value>
+
+    <p-key-value v-if="deployment.workQueueName" label="Work Queue" :alternate="alternate">
```
space-age-pete, line 15: I'm not sure if you should add `&& deployment.workPoolName` to this `v-if`, or if it's okay to display the `WorkQueueIconText` without it

space-age-pete (commented):
looks good!

src/components/WorkQueueIconText.vue
```diff
@@ -1,8 +1,8 @@
 <template>
```
znicholasbrown, line 1: This component got a lot more complicated in this PR - maybe we should break the work pool stuff into its own icon-text component instead to simplify and make the work pool name a required prop

znicholasbrown (commented):
Altogether looks good @marichka-offen - a few questions/comments

src/components/FlowRunsFilterGroup.vue
```diff
@@ -15,8 +15,8 @@
       <p-label label="Deployments">
         <DeploymentCombobox v-model:selected="deployments" empty-message="All deployments" />
       </p-label>
-      <p-label label="Work Queues">
-        <WorkQueueCombobox v-model:selected="workQueues" empty-message="All work queues" />
+      <p-label label="Work Pools">
+        <WorkPoolCombobox v-model:selected="workPoolName" empty-message="All pools" />
```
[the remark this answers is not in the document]
marichka-offen: Yes, we'll be only filtering by work pool on this page since there's no mention of work queues there, and it isn't apparent to filter by something we can't see. Here's old screenshot to show why that change is needed, since we use workPools in flowRunListItems now

Screenshot 2023-02-08 at 10 09 47 AM

src/components/WorkPoolCombobox.vue
marichka-offen: I took it from WorkQueueCombobox since we actively used it before this update, so I don't have a good answer to this question. Who could be a good person to shed some light on how it works now?

src/components/WorkPoolQueueCombobox.vue
marichka-offen: I don't think it matters as much since we do use "work queue" in the UI, so people can still think of workPoolQueue as workQueues they are so used to 
Screenshot 2023-02-08 at 1 40 01 PM

src/components/WorkPoolCombobox.vue
znicholasbrown: @stackoverfloweth should know; I _think_ you should be able to ignore the attribute entirely here and let callers pass it themselves as an attr (which the combobox will inherit)

src/components/WorkQueueIconText.vue
marichka-offen: But we don't know if we're going to get workPoolName or workPoolId, we cannot make it required. And depending on what we get, we then do different subscriptions to get data we need. For some reason, some models like deployment return pool name, and some like old workQueues return id, and they cannot be updated because it breaks something on the backend. Also, there's no point in keeping `WorkQueueIconText` since it's not used anywhere else

src/components/DeploymentDetails.vue
marichka-offen: I don't understand the question

marichka-offen: I added both to UI. If we have one of the set, we show one of them, if we have both we show both, if none - none
Screenshot 2023-02-08 at 1 49 19 PM

src/components/DeploymentDetails.vue
marichka-offen: It's good question. For now `work-pool-name` prop is optional, so technically, we don't need to have the pool name to at least show the name.  I wonder if even old deployments will have work pool name added to it right away

src/components/WorkQueueIconText.vue
marichka-offen: Nice, good suggestion! Thanks

src/components/DeploymentDetails.vue
znicholasbrown: i think i misunderstood the queue changes here, you can ignore this

znicholasbrown (approved):

src/components/WorkPoolCombobox.vue
znicholasbrown: ^ that was my hope as well, if so we could simplify the model values on lots of these comboboxes

marichka-offen: Nice! I'll make an issue for that, to go through combo boxes and clean them up

stackoverfloweth: I don't know that *for sure* we have extra complexity, just something that's worth exploring

PR #1135, opened 2023-02-09
add layout and hide edges options to timeline

brandonreid:
Users with large flow runs [have reported](https://prefect-community.slack.com/archives/C0192RWGJQH/p1675430792974719) issues with the timeline slowing down the Flow Run page.

In addition to the latest graphs update having some rendering optimizations, this introduces a couple options for faster/lighter rendering of the timeline:

Screenshot 2023-02-09 at 7 43 43 AM

### Layout options
- Nearest Parent - default
- Waterfall - this is how the timeline used to render nodes, it's much faster to calculate and will be set as the default layout if the flow run has more than the threshold number of nodes (currently set to 100)

### Hide Edges option
Edges are more expensive to render since the paths are all one-off graphics currently, so this option let's you skip rendering them and is automatically set to true if the flow run has more than the threshold number of nodes (currently set to 40). Note that edge paths for selected nodes and their dependencies will still show.


src/components/FlowRunTimelineOptions.vue
```diff
@@ -0,0 +1,91 @@
+<template>
+  <p-pop-over
```
pleek91, line 2: This is almost identical to `PIconButtonMenu`. I think you can use that component here and just pass in the attrs for your button.

src/components/FlowRunTimelineOptions.vue
```diff
@@ -0,0 +1,91 @@
...
+        ref="settingsButton"
+        aria-label="Flow Run Timeline Options"
+        icon="CogIcon"
+        flat
+        @click="toggle"
+      />
+    </template>
+    <div class="flow-run-timeline-options__popover">
+      <p-overflow-menu>
+        <div class="flow-run-timeline-options__popover-option-set">
```
pleek91, line 19: Is there a reason for using this div and custom styling rather than using the `p-overflow-menu-item` component's with its slot?

src/components/FlowRunTimelineOptions.vue
```diff
@@ -0,0 +1,91 @@
...
+    hideEdges: boolean,
+  }>()
+
+  const emit = defineEmits<{
+    (event: 'update:layout', value: TimelineNodesLayoutOptions): void,
+    (event: 'update:hideEdges', value: boolean): void,
+  }>()
+
+  const settingsButton = ref<typeof PButton>()
+  const popOver = ref<typeof PPopOver>()
```
pleek91, line 50: I think this will be solved by using PIconButtonMenu. But I don't think you need these?

src/components/FlowRunTimelineOptions.vue
```diff
@@ -0,0 +1,91 @@
...
+.flow-run-timeline-options__popover { @apply
+  py-1
+}
+
+.flow-run-timeline-options__popover-option-set { @apply
+  p-3
+}
+.flow-run-timeline-options__popover-option-set label { @apply
+  cursor-pointer
+}
```
pleek91, line 90: Why is this necessary? `p-checkbox` already has styles for this?

brandonreid: ah ya know what, it was `p-radio` that didn't. I'll get a prefect-design PR going instead.

pleek91: Ah okay nice. Yeah then these styles can be removed 👍

src/components/FlowRunTimelineOptions.vue
brandonreid: So `p-overflow-menu-item` was wrapping everything in a button. I don't believe having a radio group in a button is valid markup. Also it caused the whole area to have a `cursor: pointer` when I'd prefer it to only apply to the individual radio/checkbox items themselves.

src/components/FlowRunTimelineOptions.vue
brandonreid: You might be right with some additional props on that component. A couple things it came with I didn't want, namely closing the menu on any click within.

src/components/FlowRunTimelineOptions.vue
pleek91: You're correct. Completely forgot that component was a button. Seems like a problem we should solve in prefect-design. I'll open an issue.

pleek91 (approved):

PR #1140, opened 2023-02-10
Chore: Scrub

znicholasbrown:


demo/utilities/colorMode.ts
```diff
@@ -3,7 +3,7 @@ import { computed } from 'vue'
 import { ColorMode } from '@/types/ColorMode'
 import { applyColorModeClass, isColorMode } from '@/utilities/colorMode'
 
-const colorModeLocalStorageKey = 'orion-design-color-mode'
+const colorModeLocalStorageKey = 'prefect-ui-library-color-mode'
```
pleek91, line 6: Are we okay with breaking everyones existing color mode selection?

znicholasbrown: i'm not sure we have much of a choice tbh but that feels like an acceptable level of friction since it impacts only upgrades

pleek91: we could "deprecate" and migrate over the next couple versions. But probably more effort than its worth. Just wanted to call it out so make sure.

pleek91 (approved):

PR #1144, opened 2023-02-10
Feature: Flow collections

stackoverfloweth:
adds routes, models, mocks, maps, and components to support showing collection flows in the UI

image
image


src/maps/collectionFlow.ts
```diff
@@ -0,0 +1,19 @@
+import { CollectionFlow, CollectionFlowResponse } from '@/models'
+import { MapFunction } from '@/services/Mapper'
+
+export const mapCollectionFlowResponseToCollectionFlow: MapFunction<CollectionFlowResponse, CollectionFlow> = function(source) {
+  return {
+    collectionType: 'flow',
```
marichka-offen, line 6: Just curious, why collectionType needs to be specified directly like that?

src/router/router.ts
```diff
@@ -0,0 +1,316 @@
...
+        },
+        {
+          name: 'workspace.flows.flow',
+          path: 'flow/:flowId',
+          component: components.flow,
+        },
+        {
+          path: 'collections',
+          meta: {
+            can: 'access:collections',
```
marichka-offen, line 70: To confirm, collections is something somebody can always access if they have general read:flow permissions, right?

src/services/CollectionsApi.ts
```diff
@@ -0,0 +1,17 @@
...
+import { mapper } from '@/services/Mapper'
+import { mocker } from '@/services/Mocker'
+
+export class CollectionsApi extends Api {
+
+  protected override route: ApiRoute = '/collections'
+
+  public getFlowCollection(): Promise<CollectionFlow[]> {
+    return Promise.resolve(mocker.createMany('collectionFlow', 10))
+    // return this.get<CollectionFlowResponse[]>('views/flow')
```
marichka-offen, line 12: Does this need to be removed?

marichka-offen (approved):
Looks great! Have few questions for my own knowledge

src/models/CollectionFlow.ts
```diff
@@ -0,0 +1,12 @@
+import { CollectionItem } from '@/models/CollectionItem'
+
+export type CollectionFlow = CollectionItem<'flow'> & {
+  returns: string,
+  examples: string[],
+  documentationUrl: string,
+  installCommand: string,
+  parameters: unknown,
```
space-age-pete, line 8: should this be of type `SchemaValues` to match how parameters are typed on Deployments?

src/models/CollectionFlow.ts
```diff
@@ -0,0 +1,12 @@
+import { CollectionItem } from '@/models/CollectionItem'
+
+export type CollectionFlow = CollectionItem<'flow'> & {
```
space-age-pete: I see these fields on the `CollectionFlowResponse` type but not here.
The mapper has `pathContainingFlow` and `repositoryUrl` but not `entrypoint`

```suggestion
export type CollectionFlow = CollectionItem<'flow'> & {
entrypoint: string,
pathContainingFlow: string,
repositoryUrl: string,
```

src/maps/collectionFlow.ts
```diff
@@ -0,0 +1,19 @@
...
+    description: source.description.summary,
+    returns: source.description.returns,
+    examples: source.description.examples,
+    documentationUrl: source.documentation_url,
+    installCommand: source.install_command,
+    logoUrl: source.logo_url,
+    parameters: source.parameters,
+    pathContainingFlow: source.path_containing_flow,
+    repositoryUrl: source.repo_url,
+    slug: source.slug,
```
space-age-pete, line 17:
```suggestion
    slug: source.slug,
    entrypoint: source.entrypoint
```

src/maps/collectionFlow.ts
```diff
@@ -0,0 +1,19 @@
...
+    collectionType: 'flow',
+    name: source.name,
+    description: source.description.summary,
+    returns: source.description.returns,
+    examples: source.description.examples,
+    documentationUrl: source.documentation_url,
+    installCommand: source.install_command,
+    logoUrl: source.logo_url,
+    parameters: source.parameters,
+    pathContainingFlow: source.path_containing_flow,
```
space-age-pete, line 15: I'm using this field for the deployment create POST endpoint, which expects it as simply `'path'`

```suggestion
    path: source.path_containing_flow,
```

space-age-pete (commented):
Looks good! just a couple of notes on the models/maps

src/router/router.ts
stackoverfloweth: yes (and the feature flag), it's actually just a "read-only" type feature that's more like marketing material than an actually impactful change to how things operate.

src/services/CollectionsApi.ts
stackoverfloweth: currently there is no api for this, but yes when [this](https://github.com/PrefectHQ/nebula/pull/3542) gets merged, we will want to swap this out

src/models/CollectionFlow.ts
stackoverfloweth: I purposefully left it `unknown` because it's not something I intend on using and I couldn't confirm that it was the same contract as `SchemaValues`.

src/models/CollectionFlow.ts
stackoverfloweth: I've rolled back some of the complexity of these types. See [here](https://github.com/PrefectHQ/prefect-ui-library/pull/1144#discussion_r1103297542).

Take another look at the latest and let me know if you still have questions

src/maps/collectionFlow.ts
stackoverfloweth: I like that more anyways

src/models/CollectionFlow.ts
marichka-offen: Lots of parameters we use are also Record<string, unknown>

src/models/CollectionFlow.ts
space-age-pete: yep, looks good

src/models/CollectionFlow.ts
space-age-pete: it actually might be the same type as `Schema` rather than `SchemaValue`, not sure without looking further into it. I think keeping it loose should be fine, and if I run into any issues when I use it for creating a deployment, we can revisit then.

src/models/api/CollectionItemResponse.ts
```diff
@@ -0,0 +1,16 @@
+export type CollectionItemResponse = {
+  description: {
+    summary: string,
+    returns: string,
+    examples: string[],
+  },
```
marichka-offen, line 6: Curious why here we provide an entire object, but in response, we map and receive only `description.summary`?

marichka-offen: Oh I see, because it's broken down in the main model. But why can't we use object there too?

stackoverfloweth: we could, I just felt like a flatter model was more typical for us

PR #1146, opened 2023-02-13
BugFix: Flow run logs flashing because of polling and subscription reactivity issues

pleek91:
# Description
The flow run logs use the flow run id in its filter. Since the flow run itself is on a 5 second interval, reactive effects are triggered every 5 seconds. Since the `usePaginatedSubscription ` composition joins multiple calls together, the reactive effect of the filter empties the response array. 

There used to be a check in place that checked for equality of the watcher's new and old values. But the old and new values [will be the same when deep watching](https://vuejs.org/guide/essentials/watchers.html#deep-watchers) an object. So that [check was removed](https://github.com/PrefectHQ/prefect-ui-library/commit/85afa18c369b5a4bdc86d29d56c12b3f8aaccab5) in order to support the latest filters work. Removing that ensured filter objects were accurate, but also caused paginated subscriptions to be be resubscribed even when the actual value itself hadn't changed. 

Putting back an equality check that actually works for both simple and object values in the form of a new utility called `uniqueValueWatcher`. Which is effectively the following but written in a way that supports complex objects an arrays as the watch source. 

```typescript
watch(value, (newValue, oldValue) => {
  if(newValue === oldValue) {
    return
  }
  // do stuff
})
```

### DIFF
--- a/src/utilities/reactivity.ts
+++ b/src/utilities/reactivity.ts
@@ -1,0 +1,45 @@
+import { isEqual } from 'lodash'
+import { isRef, unref, watch } from 'vue'
+import { isArray, isRecord, mapValues } from '@/utilities'
+
+function getRawValue(value: unknown): unknown {
+  if (typeof value === 'object') {
+    if (isRef(value)) {
+      return unref(value)
+    }
+
+    if (isArray(value)) {
+      return getRawArrayValue(value)
+    }
+
+    if (isRecord(value)) {
+      return getRawRecordValue(value)
+    }
+  }
+
+  return value
+}
+
+function getRawArrayValue(values: unknown[]): unknown[] {
+  return values.map(value => getRawValue(value))
+}
+
+function getRawRecordValue(value: Record<PropertyKey, unknown>): Record<PropertyKey, unknown> {
+  return mapValues(value, (key, value) => getRawValue(value))
+}
+
+export function uniqueValueWatcher(...[source, callback, options]: Parameters<typeof watch>): ReturnType<typeof watch> {
+  let previous = getRawValue(source)
+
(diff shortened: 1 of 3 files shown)


zhen0:
Thanks for the speedy turnaround on this @pleek91. I can't check this locally right now. We'll try get this approved this evening but if we can't are you ok if we merge this (after approval) in your absence tomorrow?

src/utilities/reactivity.ts
```diff
@@ -0,0 +1,45 @@
...
+
+function getRawArrayValue(values: unknown[]): unknown[] {
+  return values.map(value => getRawValue(value))
+}
+
+function getRawRecordValue(value: Record<PropertyKey, unknown>): Record<PropertyKey, unknown> {
+  return mapValues(value, (key, value) => getRawValue(value))
+}
+
+export function uniqueValueWatcher(...[source, callback, options]: Parameters<typeof watch>): ReturnType<typeof watch> {
```
stackoverfloweth, line 31: love this abstraction. Might be worth a composition someday in vue-compositions? `useUniqueValueWatcher`?

stackoverfloweth (approved):

src/utilities/reactivity.ts
pleek91: I thought about that, but its a wrapper around `watch` which isn't a composition. Currently vue-compositions doesn't export any global utilities. So this seemed like the best place for now.

pleek91:
@zhen0 I'm going to get this out tonight so no one is blocked by it

zhen0:
> @zhen0 I'm going to get this out tonight so no one is blocked by it

Fantastic. Thank you!

PR #1147, opened 2023-02-13
Feature: Artifacts API

znicholasbrown:
This is most of the services and classes we'll need for the mvp artifacts build out

### DIFF
--- a/src/maps/artifact.ts
+++ b/src/maps/artifact.ts
@@ -1,8 +1,8 @@
-import { MapFunction } from '@/services/Mapper'
 import { ArtifactResponse } from '@/models/api/ArtifactResponse'
 import { Artifact } from '@/models/Artifact'
+import { MapFunction } from '@/services/Mapper'
 
-export const mapArtifactResponseToArtifact: MapFunction<ArtifactResponse, Artifact> = function (source) {
+export const mapArtifactResponseToArtifact: MapFunction<ArtifactResponse, Artifact> = function(source) {
   return new Artifact({
     id: source.id,
     created: new Date(source.created),
--- a/src/maps/artifact.ts
+++ b/src/maps/artifact.ts
@@ -1,0 +1,17 @@
+import { MapFunction } from '@/services/Mapper'
+import { ArtifactResponse } from '@/models/api/ArtifactResponse'
+import { Artifact } from '@/models/Artifact'
+
+export const mapArtifactResponseToArtifact: MapFunction<ArtifactResponse, Artifact> = function (source) {
+  return new Artifact({
+    id: source.id,
+    created: new Date(source.created),
+    updated: new Date(source.updated),
+    key: source.key,
+    type: source.type,
+    data: source.data,
+    metadata: source.metadata_,
+    flowRunId: source.flow_run_id,
+    taskRunId: source.task_run_id,
+  })
+}
--- a/src/maps/artifact.ts
+++ b/src/maps/artifact.ts
@@ -5,8 +5,8 @@ import { MapFunction } from '@/services/Mapper'
 export const mapArtifactResponseToArtifact: MapFunction<ArtifactResponse, Artifact> = function(source) {
   return new Artifact({
     id: source.id,
-    created: new Date(source.created),
-    updated: new Date(source.updated),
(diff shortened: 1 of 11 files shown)


src/maps/artifact.ts
```diff
@@ -0,0 +1,17 @@
+import { MapFunction } from '@/services/Mapper'
+import { ArtifactResponse } from '@/models/api/ArtifactResponse'
+import { Artifact } from '@/models/Artifact'
+
+export const mapArtifactResponseToArtifact: MapFunction<ArtifactResponse, Artifact> = function (source) {
+  return new Artifact({
+    id: source.id,
+    created: new Date(source.created),
+    updated: new Date(source.updated),
```
pleek91, line 9: There's a mapper for converting iso date strings to date objects. I think it's basically what you have here but we should use it still.

src/services/WorkspaceArtifactsApi.ts
```diff
@@ -0,0 +1,37 @@
...
+  deleteArtifact: (id: string) => Promise<void>,
+}
+
+export class WorkspaceArtifactsApi extends WorkspaceApi implements IWorkspaceArtifactsApi {
+
+  protected override routePrefix = '/artifacts'
+
+  public async getArtifact(id: string): Promise<Artifact> {
+    const { data } = await this.get<ArtifactResponse>(`/${id}`)
+    return mapper.map('ArtifactResponse', data, 'Artifact')
```
pleek91, line 19: Should we use the batch processor here?

src/types/SortOptionTypes.ts
```diff
@@ -1,6 +1,17 @@
...
+export const artifactSortValues = ['CREATED_DESC', 'UPDATED_DESC', 'KEY_DESC', 'KEY_ASC', 'ID_ASC', 'ID_DESC'] as const
+export type ArtifactSortValues = typeof artifactSortValues[number]
+export const defaultArtifactSort: FlowSortValues = 'CREATED_DESC'
+
+export function isArtifactSortValue(value: unknown): value is ArtifactSortValues
+export function isArtifactSortValue(value: Ref<unknown>): value is Ref<ArtifactSortValues>
+export function isArtifactSortValue(value: MaybeRef<unknown>): value is MaybeRef<ArtifactSortValues> {
+  const valueRef = ref(value)
+  return artifactSortValues.includes(valueRef.value as ArtifactSortValues)
+}
```
pleek91, line 13: Nice

pleek91 (commented):
Looks really good. Couple little comments

src/maps/artifact.ts
znicholasbrown: Ah good call, updated that. Also revealed that I had a type issue with the `ArtifactResponse` since that was supposed to be a string and not a date... +1 for type safe mappers

src/services/WorkspaceArtifactsApi.ts
znicholasbrown: I don't exactly have a use for it right now but... why not! Will test it a bunch while working on this

pleek91: Good enough reason for me 👍

znicholasbrown: updated - nice suggestion

src/services/WorkspaceArtifactsApi.ts
```diff
@@ -0,0 +1,43 @@
...
+}
+
+export class WorkspaceArtifactsApi extends WorkspaceApi implements IWorkspaceArtifactsApi {
+
+  protected override routePrefix = '/artifacts'
+
+  private readonly batcher = new BatchProcessor<string, Artifact>(async ids => {
+    const artifacts = await this.getArtifacts({ id: ids })
+    return toMap(artifacts, 'id')
+  }, { maxBatchSize: 200 })
```
pleek91, line 22: Love how simple this BatchProcessor is

pleek91 (approved):

src/services/WorkspaceArtifactsApi.ts
znicholasbrown: Agreed, whoever designed the api did a nice job 😄

PR #1153, opened 2023-02-14
Enhancement: deployment flow run form

znicholasbrown:
This PR adds demos for the deployment and flow run create forms and changes the default behavior for the flow run create form to show the custom parameters section if any parameters are passed as overrides (thereby straying from defaults)


demo/sections/deployments/CreateRunForm.vue
```diff
@@ -0,0 +1,48 @@
...
+  const schema = mapper.map('SchemaResponse', rawSchema, 'Schema')
+
+  const overrides: [Partial<Deployment>] = [
+    {
+      parameters: deploymentParameters,
+      parameterOpenApiSchema: schema,
+    },
+  ]
+
+  const deployment = mocker.create('deployment', overrides)
```
znicholasbrown, line 47: Split out these overrides to make testing this more consistent than when using the random mocker, which generates a completely new schema each time

pleek91 (approved):
Nice!

PR #1161, opened 2023-02-16
Feature: new api paradigm

stackoverfloweth:
Sorry for the click-bait title, this is not really a new paradigm so much as it's a consolidation towards the pattern we use (and like) from workspace api.

This pr finally removes the long depreciated API (base) APIRouteParams. That class was being used as a base class for APIs in both cloud and oss that were outside of the "workspace apis". 

In it's place, I've created BaseApi, which supports not only the existing WorkspaceApi (base) but also all of these less predictable apis across cloud/oss. The BaseApi use composition over inheritance, so extending it like I do in WorkspaceApi is very simple and should support any changes we maybe need in the foreseeable future without big rewrites.

This BaseApi notably does not support `withParams` concept that we've been running with for a bit. I opted for simplicity over slightly improved devex. 

This paradigm also assumes we are going to put all of our Apis in a single provided (later injected) object `useWorkspaceApi` in library, `useNebulaApi` in cloud, and `usePrefectApi` in oss.  Developers will have access to the full api everywhere you call `const api = useNebulaApi()`, including workspace apis, which you might not have correct context for. To avoid potential confusion, I have our WorkspaceApi (base) throw an exception if it was provided a `CloudConfig` (with AccountId & WorkspaceId), but they are undefined.

This change was inspired by a new api (collectionsApi), which was not technically a WorkspaceApi in that even when given a CloudConfig, it should not include the context in the baseUrl. So I needed a new BaseApi but also wanted it to be included with other "workspaceApis" that all get provided together, accessed via `api.collections.doSomething()`. Otherwise I would have to create a separate key for this api to provide/inject and use a deprecated base class.

To summarize
- no more mixins in cloud or oss!
- no more deprecated classes
- no more defining `_MyClass`, then export `MyClass = Authorize(_MyClass)` kind of thing
- no more `withParams`
- adds ui & health apis to workspacesApi
- cloud will have a lot of changes (PR ready when we have new prefect-ui-library version). Most of these changes are just syntax changes from `import someApi from '@/services/someApi` to `useNebulaApi()` and `api.someApi.something()`. But also cloud just defined a lot more of it's own Apis compared to oss
   - https://github.com/PrefectHQ/nebula-ui/pull/2402
- oss will have very little changes
   - https://github.com/PrefectHQ/prefect/pull/8564


space-age-pete:
I am a fan of this new paradigm

src/services/BaseApi.ts
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  return value
+}
+
+export class BaseApi<T extends PrefectConfig = PrefectConfig> {
+  protected readonly apiConfig: T
+  protected routePrefix: string | undefined
+  protected getHeaders: GetApiHeaders[]
+  protected getBaseUrl: GetApiBaseUrl
```
pleek91, line 35: I feel like we could make these more flexible with very little effort by updating the types a bit

```typescript
type ApiHeaders = MaybeArray<AxiosRequestHeaders | GetApiHeaders>
type ApiBaseUrl = string | GetApiBaseUrl
```

pleek91: I also think these should drop the `get` prefix and be just `headers` and `baseUrl`. Then rename the `compose` methods to `getHeaders` and `getBaseUrl`. IMO a little more clear as to what its doing.

pleek91: I don't feel strongly about either comment above. More strongly about the naming than the types.

src/services/BaseApi.ts
```diff
@@ -0,0 +1,128 @@
...
+      return {
+        ...headers,
+        ...getHeaders(this.apiConfig),
+      }
+    }, {})
+  }
+
+  protected removeUndefinedPaths(input: string | undefined): input is string {
+    return !!input
+  }
```
pleek91, line 59: This is a better and more clear guard IMO. I'd also put this in the `variables.ts` utility file next to the `isNotNullish` utility.
```suggestion
  protected isDefined<T>(input: T | undefined): input is T {
    return input !== undefined
  }
```

src/services/BaseApi.ts
```diff
@@ -0,0 +1,128 @@
...
+  protected removeLeadingAndTrailingSlashes(input: string): string {
+    return input.replace(/^(\/)|(\/)$/g, '')
+  }
+
+  protected combinePath(route: string | undefined): string {
+    return [this.routePrefix, route]
+      .filter(this.removeUndefinedPaths)
+      .map(this.removeLeadingAndTrailingSlashes)
+      .join('/')
+  }
```
pleek91, line 70: This is just a developer convenience around whether the baseUrl ends with a slash or the route starts with a slash yes?

src/services/HealthApi.ts
```diff
@@ -1,10 +1,17 @@
...
 
-  public getHealth(): Promise<string> {
-    return this.get('').then(({ data }) => data)
+  public async checkHealth(): Promise<boolean> {
+    try {
+      return await this.getHealth()
+    } catch {
+      return false
+    }
   }
```
pleek91, line 16: I'm guessing this existed somewhere else and you moved it into this service? 

```suggestion
  public async isHealthy(): Promise<boolean> {
    try {
      return await this.getHealth()
    } catch {
      return false
    }
  }
```

src/services/WorkspaceApi.ts
```diff
@@ -1,100 +1,33 @@
...
-  }
+  throw new CloudConfigMissingParamsError()
+}
 
-  // axios uses any for the data argument
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  protected put<T, R = AxiosResponse<T>>(url: string, data?: any, config?: AxiosRequestConfig): Promise<R> {
-    return this.instance().put(url, data, config)
-  }
+export class WorkspaceApi extends BaseApi<WorkspaceApiConfig> {
```
pleek91, line 27: Github won't let me do a suggestion for this whole file. But couple things I here. 

1. You've defined and exported `getCloudBaseUrl`. But its also used for OSS so I think `getWorkspaceBaseUrl` is more accurate (also follows the naming convention of this file). I'm guessing that naming is just left over from when the only api like this was the workspace api and not really a conscious decision you made. 
2. You've created a constructor (which you had to type and call super in) just to set `this.getBaseUrl`. Which is `protected` and can just be an override directly in this class (same result).
3. I personally wouldn't abstract or export the `getCloudBaseUrl`. Its not used anywhere else and its pretty specific to this file. If you needed that logic you probably should be extended WorkspaceApi directly. 
```typescript
export class WorkspaceApi extends BaseApi<WorkspaceApiConfig> {

  protected override getBaseUrl: GetApiBaseUrl = (config) => {
    if (!isCloudConfig(config)) {
      return config.baseUrl
    }

    if (config.accountId && config.workspaceId) {
      return `${config.baseUrl}/accounts/${config.accountId}/workspaces/${config.workspaceId}`
    }

    throw new CloudConfigMissingParamsError()
  }

}
```

src/services/BaseApi.ts
```diff
@@ -0,0 +1,128 @@
...
+  const value: AxiosRequestHeaders = {}
+
+  if (config.token) {
+    value.Authorization = `bearer ${config.token}`
+  }
+
+  return value
+}
+
+export class BaseApi<T extends PrefectConfig = PrefectConfig> {
```
pleek91, line 31: Since this is not an abstract class I think we can drop the `Base` prefix. Not a rule but I usually expect `BaseThing` classes to not be functional on their own which I believe this class is.

pleek91 (commented):
Nice work! Really excited to remove all the deprecated services and get rid of the mixins in nebula 🎉

src/services/BaseApi.ts
stackoverfloweth: I agree with both comments

src/services/BaseApi.ts
stackoverfloweth: correct. I tried to make slashes more consistent in these PRs but moving forward, ideally developers wouldn't have to know whether to include leading/trailing slashes on routes or routePrefix

src/services/HealthApi.ts
stackoverfloweth: yeah, both cloud and oss defined basically the same function

src/services/WorkspaceApi.ts
stackoverfloweth: agreed with all of this

src/services/Api.ts
```diff
@@ -28,43 +26,44 @@ export const getAuthorizationHeaders: GetApiHeaders = (config) => {
...
+
+    return this.apiBaseUrl(this.apiConfig)
   }
 
   protected composeHeaders(): AxiosRequestHeaders {
-    return this.getHeaders.reduce((headers, getHeaders) => {
+    const array = asArray(this.apiHeaders)
+
+    return array.reduce<AxiosRequestHeaders>((headers, header) => {
+      const value = typeof header === 'function' ? header(this.apiConfig) : header
```
pleek91, line 51: This is totally random. But an `asFunction` utility could be a cool. 

```typescript
function asFunction<T>(value: T | () => T): () => T {
  if(typeof value === 'function') {
    return value
  }

  return () => value
}
```

pleek91: maybe `asMethod` 🤔

pleek91 (approved):
Nice work!

PR #1163, opened 2023-02-18
Feature: ArtifactDescription and ArtifactSummaryCard

znicholasbrown:
This PR introduces 2 new components, `ArtifactDescription`, which is a card-wrapped component for the prefect-design markdown renderer, specifically for artifacts, and `ArtifactSummaryCard`, which is a flexible component meant to be used in grids and lists on artifact overview pages and as summary content on individual artifact pages and visualizations.

It also:
- Tightens some of the types associated with artifact metadata (this is currently an implicit contract that the UI is holding as explicit until the experimental phase is over)
- Adds some better handling for un-keyed artifacts (results, by default)
- Adds mock generators for a few different markdown items: `markdownCodeBlockString`, `markdownCodeSpanString`, `markdownContentString`, `markdownHeaderString`, and `markdownQuoteString`, and a composable for all of these: `markdownString`; nothing special about these really but we can augment them later if we need more dynamic content (e.g. for images, lists, dynamic html etc etc)
- Adds a seeded mocking composition for artifacts: `useArtifacts`

Note: this requires prefect-design `^1.3.1` so that has also been included.


demo/compositions/useArtifactsMock.ts
```diff
@@ -0,0 +1,45 @@
+import { useSeeds } from './useSeeds'
+import { Artifact } from '@/models'
+import { mocker } from '@/services'
+import { coinflip, repeat } from '@/utilities'
+
+export function useArtifactMock(override?: Partial<Artifact>, useTaskRun: boolean = coinflip(0.5)): Artifact {
```
pleek91, line 6: Am I missing something or is the override not used anywhere

demo/sections/artifacts/ArtifactCards.vue
```diff
@@ -0,0 +1,39 @@
...
+    { title: 'Markdown Description', description: 'A card used to render the internal markdown of an artifact' },
+  ]
+
+  const start = new Date()
+  start.setHours(0)
+
+  const condense = ref(false)
+  const artifacts = useArtifactsMock(1)
+
+  const artifact = computed(() => artifacts[0])
```
pleek91, line 38: Wouldn't it be easier to use the singular mock?

src/components/ArtifactSummaryCard.vue
```diff
@@ -0,0 +1,224 @@
...
+  const props = defineProps<{
+    artifact: Artifact,
+    condense?: boolean,
+  }>()
+
+  const slots = useSlots()
+  const routes = useWorkspaceRoutes()
+
+  const flowRunId = ref(props.artifact.flowRunId)
+  const taskRunId = ref(props.artifact.taskRunId)
```
pleek91, line 67: I think these need to be computed

src/components/ArtifactSummaryCard.vue
```diff
@@ -0,0 +1,224 @@
...
+  const artifactLabel = computed(() => {
+    return props.artifact.key ? localization.info.artifactLabel : localization.info.artifactResultLabel
+  })
+
+  const crumbs = computed<BreadCrumbs>(() => {
+    const internalCrumbs: BreadCrumbs = []
+
+    if (flowRunId.value) {
+      internalCrumbs.push({
+        text: flowRun.value?.name ?? '|',
```
pleek91, line 83: Why the pipe default? Won't there always be a name on a flow run?

src/mocks/markdown.ts
```diff
@@ -0,0 +1,158 @@
...
+    markdownLines.push({
+      type: 'content',
+      content: this.create('markdownContentString'),
+    })
+
+    continue
+  }
+
+  return markdownLines.map(mdl => mdl.content).join('\n\n')
+}
```
pleek91, line 158: Nice

src/utilities/dates.ts
```diff
@@ -1,11 +1,14 @@
...
+import { localization } from '@/localization'
 import { secondsToApproximateString } from '@/utilities/seconds'
 import { dateFunctions, toDate, formatDateInTimezone } from '@/utilities/timezone'
 
 const dateTimeNumericFormat = 'yyyy/MM/dd hh:mm:ss a'
 const timeNumericFormat = 'hh:mm:ss a'
 const timeNumericShortFormat = 'hh:mm a'
 const dateFormat = 'MMM do, yyyy'
+const dateMonthDayFormat = 'MMM d'
+const dateMonthDayYearFormat = 'MMM d, yyyy'
```
pleek91, line 11: This is almost identical to the dateFormat. We should pick one and stick to it.

src/utilities/dates.ts
```diff
@@ -89,3 +92,35 @@ export function formatDateTimeRelative(value: Date | string, comparedTo: Date |
 
   return `in ${formatted}`
 }
+
+export function formatDateTime(value: Date | string, formatTime = timeNumericShortFormat): string {
```
pleek91, line 96: Is this a relative format? Like "today at 6:45 pm"?

src/localization/locale/en.ts
```diff
@@ -86,6 +86,15 @@ export const en = {
...
   info: {
+    today: 'today',
+    yesterday: 'yesterday',
+    tomorrow: 'tomorrow',
+    artifactLabel: 'Artifact',
+    artifactResultLabel: 'Result',
+    artifactCreatedTimestampLabel: 'Created',
+    artifactUpdatedTimestampLabel: 'Last Updated',
+    artifactRunLabel: 'Run',
+    artifactRunNone: 'None',
```
pleek91, line 97: I'd shorten these to basically what text they are. Remove the artifact prefix and label suffix. We can use these for things other than artifacts and label is a little redundant IMO

pleek91:
@znicholasbrown What's the thought behind the condensed version of the card? They're pretty similar. Curious if we even need two versions?

demo/compositions/useArtifactsMock.ts
znicholasbrown: Nah you're right, forgot to pass it to the artifact, nice catch (I'm surprised I didn't get a warning about that, these files might be ignored)

src/components/ArtifactSummaryCard.vue
znicholasbrown: Meant that to be an empty string - there will always be a name on the flow run (if one exists) but the flow run is async so there's no guarantee it's here

src/utilities/dates.ts
znicholasbrown: Going off the designs for this one but it doesn't feel like dropping ordinality is a significant departure? Certainly not all dates should be written as "Feb 21st"

src/utilities/dates.ts
znicholasbrown: Correct, I really want to clean up these utilities and give them some more descriptive names but that's a wider change than feels necessary here. Maybe this could be better named something like `formatDateTimeColloquial` or something? Not sold on that

znicholasbrown:
We probably don't _need_ them but it's a nice (and easy) to have toggle that lets us create both compact and lazier layouts with the same component; this will be important for views that include just top-level artifact keys (a lazy view will likely suffice) AND more functional views that include all results and/or versions of an artifact key. Originally I had those broken into two components but the only difference was some CSS so they felt needlessly redundant.

znicholasbrown:
Here's an example of the views this enables:

Screen Shot 2023-02-21 at 11 35 22 AM
Screen Shot 2023-02-21 at 11 35 27 AM

src/utilities/dates.ts
pleek91: The [style guide](https://www.notion.so/prefect/Formatting-Basics-1d4d83bcb2ba471ead90910aeb5913b2) defines our date formats. 

We deliberately picked two formats for dates/times. One "pretty" (dateFormat) and one more spreadsheet like for lists (dateNumeric). IMO introducing variations just introduces inconsistency and also makes devs/designers have to think about what format they should use. 

The style guide also has some rules for "relative" time though we haven't implemented that yet (I'd like to).

src/utilities/dates.ts
pleek91: realizing its missing but `formatDateTime` would be the dateTime formate of `MMM do, yyyy at hh:mm a`. Can we rename this to formateDateTimeRelative? Which following the naming of our style guide.

pleek91:
@znicholasbrown interesting, thanks for explaining the purpose behind the two formats!

PR #1170, opened 2023-02-22
Feature: changed logic for preventing dupe slashes in api

stackoverfloweth:
cc @pleek91 https://github.com/PrefectHQ/nebula-ui/pull/2435#issuecomment-1439364126

This came to me when thinking about this problem last night. I think this is the best solution because it means developers can continue to not worry about where to put leading/trailing slashes but if we need to make sure a route ends in slash this will not conflict with such logic.

### DIFF
--- a/src/services/Api.ts
+++ b/src/services/Api.ts
@@ -57,15 +57,13 @@ export class Api<T extends PrefectConfig = PrefectConfig> {
     }, {})
   }
 
-  protected removeLeadingAndTrailingSlashes(input: string): string {
-    return input.replace(/^(\/)|(\/)$/g, '')
-  }
-
   protected combinePath(route: string | undefined): string {
+    const repeatingSlashes = /(\/+)/g
+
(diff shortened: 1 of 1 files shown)


pleek91 (approved):

src/services/Api.ts
```diff
@@ -57,15 +57,13 @@ export class Api<T extends PrefectConfig = PrefectConfig> {
     }, {})
   }
 
-  protected removeLeadingAndTrailingSlashes(input: string): string {
-    return input.replace(/^(\/)|(\/)$/g, '')
-  }
-
   protected combinePath(route: string | undefined): string {
+    const repeatingSlashes = /(\/+)/g
```
brandonreid, line 61: Nit: So this will get all slashes eh? Maybe a better name is `slashes` or `findSlashes`, idk.

brandonreid (approved):
Rendsday's paying off.

PR #1171, opened 2023-02-22
Feature: Replace CodeSnippet component (step 1)

stackoverfloweth:
this PR is step 1/2 in replacing CodeSnippet component and embracing new p-code-highlight and p-code-input components from prefect-design.

notable changes
- removes deprecated components `CodeHighlighting` and `CodeLines`, which are [now no longer used in cloud](https://github.com/PrefectHQ/nebula-ui/pull/2404) or oss
- introduces a new component CopyableWrapper, which is meant to replace CodeSnippet since all CodeSnippet does that p-code-highlight doesn't is provide the little "copy" button.
image
- CopyableWrapper is intentionally more flexible and simply exposes a default slot for whatever you want it to wrap (doesn't have to be code)
- CopyableWrapper takes a single prop `text-to-copy`, which does what you'd expect
- updates JsonInput and PythonInput to use the new p-code-input component from prefect-design to drive functionality. We could probably remove these altogether in the near future, but these inputs offer a "format" button, which is not currently supported by p-code-input. So no plans to completely remove those as of right now.

minor changes
- sets background color to not be white [context](https://github.com/PrefectHQ/prefect-design/pull/662#issuecomment-1439188554)
- expands sidebar to allow for lengthy block component names


src/components/CopyableWrapper.vue
```diff
@@ -0,0 +1,46 @@
...
+}
+
+.copyable-wrapper:hover .copyable-wrapper__button { @apply
+  opacity-100
+}
+
+.copyable-wrapper__button { @apply
+  absolute
+  top-0
+  right-0
```
brandonreid, line 36: maybe `top/right-1` or even 2 to give it just a touch of breathing room from the edge?

src/components/CodeSnippet.vue
```diff
@@ -1,6 +1,6 @@
 <template>
   <div class="code-snippet">
-    <CodeHighlighting class="code-snippet__code" :value="snippet" :language="language" :show-line-numbers="showLineNumbers" />
+    <p-code-highlight class="code-snippet__code" :text="snippet" :lang="language" :show-line-numbers="showLineNumbers" />
     <PButton size="sm" class="code-snippet__button" @click="copy">
       Copy <p-icon icon="DuplicateIcon" />
     </PButton>
```
brandonreid, line 6: Should this use the CopyableWrapper component? If not, suggest same style treatments for the button as proposed in that component.

src/components/CopyableWrapper.vue
```diff
@@ -0,0 +1,46 @@
+<template>
+  <div class="copyable-wrapper">
+    <slot />
+    <PButton size="sm" class="copyable-wrapper__button" @click="copy">
```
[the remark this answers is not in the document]
brandonreid: I also do wonder if the "Copy" text is necessary. Could you get away with a floating button here, similar to the treatment on the Flow Run Timeline. Can add a title attribute of "Copy" (I think that's ok for accessibility, not 100% on that) so you get the free tooltip.

src/components/CopyableWrapper.vue
stackoverfloweth: I thought of that, but ultimately went with the assumption that most of these are in p-card, which has plenty of padding. I don't want to put a lot of opinions in a generic component that's meant to just wrap your content

src/components/CopyableWrapper.vue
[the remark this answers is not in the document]
stackoverfloweth: love that
image

src/components/CodeSnippet.vue
stackoverfloweth: this whole component is going to be removed in part 2, I only made this change because this PR removed CodeHighlighting

brandonreid (approved):

PR #1173, opened 2023-02-22
Feature: Workers collection fields in worker pool forms

marichka-offen:


znicholasbrown (approved):

src/components/WorkPoolCreateForm.vue
```diff
@@ -31,26 +39,47 @@
...
+  const baseJobConfigs = computed(() => {
+    return workersCollectionItems.value?.find((item) => item.type === type.value)?.defaultBaseJobConfiguration ?? {}
+  })
+
+  const schema = computed<Schema>(() => mapper.map('SchemaResponse', baseJobConfigs.value.variables ?? {}, 'Schema'))
+  const parameters = ref()
+
+  const testSchema = computed<Schema>(() => {
+    return { ...schema.value, type: 'object' }
+  })
```
pleek91, line 75: @desertaxle just calling out that this `testSchema` was for debugging. I couldn't figure out why the schema wasn't rendering and turned out to be because the schema fields component was expecting the `object` type. Not sure if that's a bug in the schemes themselves being used here or if the component should know that anything with `properties` is an `object`.

PR #1174, opened 2023-02-22
BugFix: Flow run filters null dates and recursive filtering

pleek91:
# Description
Fixed several issues here
1. Make sure the date range doesn't return nulls while picking a new range
2. Make sure the default date ranges apply to all components (not just the filters component)
3. Fix `isRecord` utility incorrectly returning `true` when given a `Date`.

### DIFF
--- a/src/components/FlowRunsFilterGroup.vue
+++ b/src/components/FlowRunsFilterGroup.vue
@@ -2,7 +2,7 @@
   <div class="flow-runs-filter-group">
     <div class="flow-runs-filter-group__row">
       <p-label :label="media.hover ? 'Date Range' : ''">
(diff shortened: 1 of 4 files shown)


stackoverfloweth (approved):

src/components/FlowRunsFilterGroup.vue
```diff
@@ -31,27 +31,32 @@
...
     flowRuns: {
       nameLike: flowRunNameLikeDebounced,
-      expectedStartTimeAfter,
-      expectedStartTimeBefore,
+    },
+  })
+
+  const range = computed<[Date, Date]>({
+    get() {
+      return [filter.flowRuns.expectedStartTimeAfter!, filter.flowRuns.expectedStartTimeBefore!]
```
znicholasbrown, line 55: Why do these need non-null assertions?

PR #1180, opened 2023-02-23
1.1.9

stackoverfloweth:
not sure why going to 1.1.8 failed
image


pleek91:
oh, is this because we went back to 1.0.0 when we changed the name? I bet there are tons of landmine tags that already exist 😬  @znicholasbrown

znicholasbrown:
Oo that's... unfortunate

PR #1190, opened 2023-02-28
Feature: Update work pool edit form to allow updates to the default values for the base job template

desertaxle:
## Summary
Updates `WorkPoolEditForm` to allow changes to the default values for the base job template variables for a work pool. Similar to `WorkPoolCreateForm`, the values from the form are used to set the default values for the `variables` section of the base job template. The starting values for the form fields are taken from the default values in the `variables` section of the base job template for easy editing. The type dropdown for the work pool is displayed, but disabled because a work pool's type cannot be changed after creation.

Note, this PR's base branch is the branch for #1173 due this PR's dependence on work in that PR. I recommend reviewing #1173 first.

## Screenshot
Screenshot 2023-02-28 at 9 18 12 AM

### DIFF
--- a/src/components/WorkPoolEditForm.vue
+++ b/src/components/WorkPoolEditForm.vue
@@ -23,7 +23,7 @@
         <WorkPoolTypeSelect :selected="type" disabled />
       </p-label>
 
-      <p-label v-if="type && schemaHasProperties && can.access.workers" label="Base Job Configuration">
+      <p-label v-if="showSchemaForm" label="Base Job Configuration">
         <SchemaFormFieldsWithValues v-model:values="parameters" :schema="schema" />
       </p-label>
     </p-content>
@@ -43,12 +43,13 @@
   import { useValidationObserver } from '@prefecthq/vue-compositions'
   import { ref, computed } from 'vue'
   import { useRouter } from 'vue-router'
-  import { SubmitButton, WorkPoolTypeSelect, SchemaFormFieldsWithValues } from '@/components'
+  import { SubmitButton, WorkPoolTypeSelect, SchemaFormFieldsWithValues, SchemaPropertiesKeyValues } from '@/components'
   import { useCan, useWorkspaceApi, useWorkspaceRoutes } from '@/compositions'
   import { localization } from '@/localization'
   import { WorkPool, WorkPoolEdit } from '@/models'
   import { mapper } from '@/services/Mapper'
   import { Schema } from '@/types/schemas'
+  import { getSchemaDefaults } from '@/utilities/parameters'
 
   const props = defineProps<{
     workPool: WorkPool,
@@ -65,29 +66,28 @@
   const concurrencyLimit = ref<number | null | undefined>(props.workPool.concurrencyLimit)
   const schema = computed<Schema>(() => mapper.map('SchemaResponse', props.workPool.baseJobTemplate.variables ?? {}, 'Schema'))
   // Set parameters to the default values from the schema so they are pre-filled in the form
-  const parameters = ref<Record<string, unknown>>(
-    Object.entries(props.workPool.baseJobTemplate.variables?.properties ?? {}).reduce((acc, [key, value]) => ({ ...acc, [key]: value?.default }), {}),
+  const parameters = ref(
+    getSchemaDefaults(props.workPool.baseJobTemplate.variables ?? {}),
   )
-
   const schemaHasProperties = computed(() => {
     const { properties } = schema.value
 
     return properties && Object.keys(properties).length > 0
   })
+  const showSchemaForm = computed(() => type.value && schemaHasProperties.value && can.access.workers)
 
   function cancel(): void {
     router.back()
   }
 
   const submit = async (): Promise<void> => {
-    const baseJobTemplateSchema = mapper.map('WorkerSchemaProperty', { values: parameters.value, schema: props.workPool.baseJobTemplate }, 'WorkerSchemaPropertyRequest')
-
     const valid = await validate()
     if (valid) {
       const values: WorkPoolEdit = {
         description: description.value,
         concurrencyLimit: concurrencyLimit.value,
-        baseJobTemplate: baseJobTemplateSchema,
+        baseJobTemplate: props.workPool.baseJobTemplate,
+        updatedDefaultVariableValues: parameters.value,
       }
       try {
         await api.workPools.updateWorkPool(props.workPool.name, values)
--- a/src/components/WorkPoolEditForm.vue
+++ b/src/components/WorkPoolEditForm.vue
@@ -20,7 +20,7 @@
       </p-label>
 
       <p-label v-if="can.access.workers" label="Type">
-        <WorkPoolTypeSelect v-model:selected="type" disabled />
+        <WorkPoolTypeSelect :selected="type" disabled />
       </p-label>
 
       <p-label v-if="type && schemaHasProperties && can.access.workers" label="Base Job Configuration">
--- a/src/components/WorkPoolEditForm.vue
+++ b/src/components/WorkPoolEditForm.vue
@@ -43,7 +43,7 @@
   import { useValidationObserver } from '@prefecthq/vue-compositions'
   import { ref, computed } from 'vue'
   import { useRouter } from 'vue-router'
-  import { SubmitButton, WorkPoolTypeSelect, SchemaFormFieldsWithValues, SchemaPropertiesKeyValues } from '@/components'
+  import { SubmitButton, WorkPoolTypeSelect, SchemaFormFieldsWithValues } from '@/components'
   import { useCan, useWorkspaceApi, useWorkspaceRoutes } from '@/compositions'
   import { localization } from '@/localization'
   import { WorkPool, WorkPoolEdit } from '@/models'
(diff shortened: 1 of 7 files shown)


src/components/WorkPoolEditForm.vue
```diff
@@ -18,8 +18,17 @@
           <p-number-input :id="id" v-model="concurrencyLimit" placeholder="Unlimited" :min="0" />
         </template>
       </p-label>
+
+      <p-label v-if="can.access.workers" label="Type">
+        <WorkPoolTypeSelect v-model:selected="type" disabled />
```
znicholasbrown, line 23:
```suggestion
        <WorkPoolTypeSelect :selected="type" disabled />
```
Since you don't intend for this value to be updatable by the component, you don't need to v-model this

src/components/WorkPoolEditForm.vue
```diff
@@ -18,8 +18,17 @@
...
+      </p-label>
+
+      <p-label v-if="type && schemaHasProperties && can.access.workers" label="Base Job Configuration">
```
znicholasbrown, line 26:
```suggestion
      <p-label v-if="type && schemaHasProperties && can.access.workers" label="Base Job Configuration">
```
Can we move the `label` to a localized field? We also generally use `computed` variables in the setup script for more complicated conditionals, so:
```ts
type && schemaHasProperties && can.access.workers
```

would become

```ts
const showSchemaForm = computed(() => type.value && schemaHasProperties.value && can.access.workers)
```

src/components/WorkPoolEditForm.vue
```diff
@@ -32,35 +41,53 @@
...
   const { validate, pending } = useValidationObserver()
 
   const description = ref<string | null | undefined>(props.workPool.description)
+  const type = ref<string>(props.workPool.type)
   const concurrencyLimit = ref<number | null | undefined>(props.workPool.concurrencyLimit)
+  const schema = computed<Schema>(() => mapper.map('SchemaResponse', props.workPool.baseJobTemplate.variables ?? {}, 'Schema'))
+  // Set parameters to the default values from the schema so they are pre-filled in the form
+  const parameters = ref<Record<string, unknown>>(
+    Object.entries(props.workPool.baseJobTemplate.variables?.properties ?? {}).reduce((acc, [key, value]) => ({ ...acc, [key]: value?.default }), {}),
+  )
```
znicholasbrown, line 70: This is pretty messy - could we use the existing parameters types and mappers for this?

src/components/WorkPoolEditForm.vue
```diff
@@ -32,35 +41,53 @@
...
+
+    return properties && Object.keys(properties).length > 0
+  })
 
   function cancel(): void {
     router.back()
   }
 
   const submit = async (): Promise<void> => {
+    const baseJobTemplateSchema = mapper.map('WorkerSchemaProperty', { values: parameters.value, schema: props.workPool.baseJobTemplate }, 'WorkerSchemaPropertyRequest')
```
znicholasbrown, line 83: I'm a little surprised this isn't happening as part of the `updateWorkPool` call - I don't frequently see us invoke the mapper before we make an API request in a component

src/components/WorkPoolTypeSelect.vue
```diff
@@ -13,9 +13,10 @@
   import { useWorkspaceApi } from '@/compositions'
   import { titleCase } from '@/utilities'
 
-  const props = defineProps<{
+  const props = withDefaults(defineProps<{
     selected: string | null | undefined,
-  }>()
+    disabled: boolean,
+  }>(), { disabled: false })
```
znicholasbrown, line 19: If you use a `boolean` prop, you don't need to [specify a default](https://vuejs.org/guide/components/props.html#boolean-casting)

znicholasbrown (commented):
A few comments @desertaxle - lmk if you want to talk through any of them

src/components/WorkPoolEditForm.vue
desertaxle: I used `getSchemaDefaults` from `utilities/parameters` which appears to do the thing I want.

src/components/WorkPoolTypeSelect.vue
desertaxle: I got a type error if I didn't specify a default, so I make `disabled` optional in the props typing. Is that standard when declaring `boolean` props?

src/components/WorkPoolEditForm.vue
desertaxle: I moved the mapper into the `updateWorkPool` call and updated `updateWorkPool` to accept the inputs from the form. Let me know if that pattern works better.

src/components/WorkPoolTypeSelect.vue
pleek91: An optional boolean defaults to `false` automatically. You probably got a type error when you went to use this component? Making it optional is preferred over setting a default value. 

More context from the [docs](https://vuejs.org/guide/components/props.html#prop-validation)
> The Boolean absent props will be cast to false. You can change this by setting a default for it — i.e.: default: undefined to behave as a non-Boolean prop.

src/components/WorkPoolEditForm.vue
```diff
@@ -32,24 +41,40 @@
...
   const { validate, pending } = useValidationObserver()
 
   const description = ref<string | null | undefined>(props.workPool.description)
+  const type = ref<string>(props.workPool.type)
   const concurrencyLimit = ref<number | null | undefined>(props.workPool.concurrencyLimit)
+  const schema = computed<Schema>(() => mapper.map('SchemaResponse', props.workPool.baseJobTemplate.variables ?? {}, 'Schema'))
+  // Set parameters to the default values from the schema so they are pre-filled in the form
+  const parameters = ref(
+    getSchemaDefaults(props.workPool.baseJobTemplate.variables ?? {}),
+  )
```
pleek91, line 71: This should already be happening in the `SchemaFormFieldsWithValues` component in its values getter. Is this necessary to do it again here?

src/components/WorkPoolTypeSelect.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <PSelect v-model="model" :options="options" class="block-type-select">
+  <PSelect v-model="model" :options="options" class="block-type-select" :disabled="disabled">
```
pleek91, line 2: nit:
```suggestion
  <PSelect v-model="model" class="block-type-select" v-bind="{ options, disabled }">
```

src/services/WorkspaceWorkPoolsApi.ts
```diff
@@ -39,6 +39,14 @@ export class WorkspaceWorkPoolsApi extends WorkspaceApi implements IWorkspaceWor
...
 
   public updateWorkPool(name: string, request: WorkPoolEdit): Promise<void> {
+    const baseJobTemplateSchema = mapper.map(
+      'WorkerSchemaProperty',
+      { values: request.updatedDefaultVariableValues ?? {}, schema: request.baseJobTemplate ?? {} },
+      'WorkerSchemaPropertyRequest',
+    )
+
+    request.baseJobTemplate = baseJobTemplateSchema
+
```
pleek91, line 49: This is a little out of place. There's already a `WorkPoolEdit` mapper on line 50. Rather than mapping and updating the `baseJobTemplate` property here in this service method, mapping the template should be part of the `WorkPoolEdit` mapper.

src/components/WorkPoolEditForm.vue
desertaxle: The default values were rendered as placeholders instead of populating the inputs with the default values. Is populating the inputs with the default values something that is already built into `SchemaFormFieldsWithValues`?

pleek91: It should be based on what I'm seeing https://github.com/PrefectHQ/prefect-ui-library/blob/edit-work-pool-base-job-template/src/components/SchemaFormFieldsWithValues.vue#L25

desertaxle: After some inspection, the function that's called there gets default values based on the type (empty string for a `string` type, `null` for a `boolean` type). I'm getting the default value defined in the schema and using that for the values in the inputs, which is why this `getSchemaDefaults` call is necessary.

pleek91: Oh I see, I confused the two utilities. Thanks for clarifying

src/maps/workerCollection.ts
```diff
@@ -4,23 +4,26 @@ import {
...
-  ).map((worker_data) => ({
-    defaultBaseJobConfiguration: worker_data.default_base_job_configuration,
-    description: worker_data.description,
-    documentationUrl: worker_data.documentation_url,
-    installCommand: worker_data.install_command,
-    logoUrl: worker_data.logo_url,
-    type: worker_data.type,
-  }))
+  return Object.values(source)
+    .reduce<WorkerCollectionItemResponse[]>((acc, package_data) => [...acc, ...Object.values(package_data)], [])
```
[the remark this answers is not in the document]
desertaxle: Unfortunately, `.flat()` doesn't work in this case because we're dealing with nested objects, but I was able to use `.flatMap()` to clean the logic up a bit.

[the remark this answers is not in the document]
pleek91: Ah that makes, sense. I was struggling to pick apart that reduce. Much cleaner ty

pleek91 (approved):

PR #1195, opened 2023-03-01
Enhancement: Add modal when execute quick run

bsignoret:
Currently when we want to run a deployment with the quick run button, if there are required parameters the deployment is still run.

The PR adds a modal where you have to fill in the required parameters. In case there are no required parameters, the deployment is launched with the default parameters


pleek91:
@bsignoret thanks for the PR! This is great solution to quick run deployments with required parameters. You've got all the pieces here but there are some changes you made that I'd like to undo. Primarily including the RunMenu component as part of the DeploymentMenu component and removing the `showAll` prop. 

I pulled down your branch and ran it locally and messed with it a bit and I think it would work great if the QuickRunParametersModal you made was used in the DeploymentQuickRunOverflowMenuItem. That way that component has all the logic it needs to do a quick run. Then the other components that use it won't need to worry about parameters. 

Let me know if that makes sense or if you have any questions. Also, as you've done most of the hard work already I don't mind helping carry this over the finish line. Let me know how I can help 👍

bsignoret:
Thanks @pleek91 for the review! My first attempt was to modify only DeploymentQuickRunOverflowMenuItem which should contain all logic, I agree with you.
However, I encountered rendering problems. DeploymentQuickRunOverflowMenuItem is used inside p-icon-button-menu or p-pop-over component. The menu is not displayed anymore when we add a modal inside it. We get something that should look like this: 
```html
<template>
    <page-heading class="page-heading-deployment">
        <template #actions>
            <p-icon-button-menu>
                <p-overflow-menu-item @click="openParametersModal">
                    <slot>Quick run</slot>
                </p-overflow-menu-item>
                <p-modal :showModal="showParametersModal" class="parameters-modal" title="Run Deployment"/>
            </p-icon-button-menu>
        </template>
    </page-heading>
      <br>
</template>

<script lang="ts" setup>
import {useShowModal} from "@/compositions";
import {PageHeading} from "@/components";

const { showModal: showParametersModal, open: openParametersModal, close: closeParametersModal } = useShowModal()
</script>
```

There seems to be a conflict between p-icon-button-menu and p-modal.

When we move p-modal out of p-icon-button-menu it works
```html
<template>
    <page-heading class="page-heading-deployment">
        <template #actions>
            <p-icon-button-menu>
                <p-overflow-menu-item @click="openParametersModal">
                    <slot>Quick run</slot>
                </p-overflow-menu-item>
            </p-icon-button-menu>
            <p-modal :showModal="showParametersModal" class="parameters-modal" title="Run Deployment"/>
        </template>
    </page-heading>
      <br>
</template>

<script lang="ts" setup>
import {useShowModal} from "@/compositions";
import {PageHeading} from "@/components";

const { showModal: showParametersModal, open: openParametersModal, close: closeParametersModal } = useShowModal()
</script>
```

To get around this problem I had two ideas. The first one is to move the modal and quick run logic out of p-icon-button-menu and therefore move it into DeploymentMenu (the one I proposed).
The second was to pass the modal run event to DeploymentQuickRunOverflowMenuItem. The component becomes much more difficult to use and we have a lot of repetition

Do you have any idea that I can investigate?

pleek91:
@bsignoret ah you know what we've run into that before. And looking around we always put the modal outside the menu. With that in mind I think what I would do is add the modal to both the RunMenu and the DeploymentMenu. Rather than combining the two components. That way all the existing layouts and logic (showAll for example) work the same. 

Thanks for the detailed explanation! Super helpful.

bsignoret:
Hi @pleek91,
As requested I have rollback the changes and added the modal for RunMenu and for DeploymentMenu.

Let me know if it looks good to you.

Thank you!

pleek91:
@bsignoret nice work! This is looking good to me. I think you accidentally removed the content of the DeploymentsTable and PageHeadingDeployment components. Maybe just checkout those files from main so they don't show up as part of your PR? 

@znicholasbrown can you take a look at this from a deployments perspective? The code looks good to me.

src/components/DeploymentQuickRunOverflowMenuItem.vue
```diff
@@ -13,29 +13,40 @@
...
 
   const run = async (): Promise<void> => {
-    try {
-      const flowRun = await api.deployments.createDeploymentFlowRun(props.deploymentId, {
-        state: {
-          type: 'scheduled',
-          message: 'Run from the Prefect UI with defaults',
-        },
-      })
+    if (props.deployment.parameterOpenApiSchema.required) {
```
znicholasbrown, line 28: @bsignoret this gave me a bit of pause just from a readability perspective. Suggestion:
```suggestion
    if (props.deployment.parameterOpenApiSchema.required?.length > 0) {
```

This will ensure 1) the key exists and 2) there are required properties in the list; your implementation works but imo requires too much thought into how truthiness works in JS

znicholasbrown (commented):
1 suggestion from me @bsignoret otherwise LGTM and really appreciate this contribution 🥳

bsignoret:
Thanks for your review! I have made the suggested changes

znicholasbrown (approved):
Thanks for the changes @bsignoret!

PR #1196, opened 2023-03-01
Maintenance: Updates to WorkPoolCreateForm

desertaxle:
- Updates to `WorkPoolCreate` -> `WorkPoolCreateRequest` mapping to reflect structure when editing work pools
- Populate default values for `WorkPoolCreateForm` with the default values in the `variables` schema for the default base job template for the corresponding worker type


src/components/WorkPoolCreateForm.vue
```diff
@@ -67,18 +69,23 @@
...
+  const schemaDefaultValues = computed(() => getSchemaDefaults(baseJobConfigs.value.variables ?? {}))
   const parameters = ref()
 
-  const testSchema = computed<Schema>(() => {
-    return { ...schema.value, type: 'object' }
+  // Update default values for form selected type changes
+  watch(type, (newType, oldType) => {
+    if (newType !== oldType) {
+      parameters.value = schemaDefaultValues.value
+    }
```
pleek91, line 80: This is so the correct values are used when the schema changes yeah? I did something similar for automations when picking a deployment (changing the deployment changes the schema). I added the ability to remember previous values if you switched back and forth. Might be worth implementing that pattern here. 

https://github.com/PrefectHQ/nebula-ui/blob/main/src/components/AutomationActionRunDeployment.vue#L32

pleek91 (approved):
One comment about tracking values when the schema changes but overall LGTM 👍

src/components/WorkPoolCreateForm.vue
```diff
@@ -23,9 +23,10 @@
         <WorkPoolTypeSelect v-model:selected="type" />
       </p-label>
 
-      <p-label v-if="type && schemaHasProperties && can.access.workers" label="Base Job Configuration">
-        <SchemaFormFieldsWithValues v-model:values="parameters" :schema="testSchema" />
-      </p-label>
+      <template v-if="showSchemaForm">
```
zhen0, line 26: Do we need to retain the can.access.workers feature flag check here?

src/components/WorkPoolCreateForm.vue
desertaxle: Oh yeah, that's a much cleaner pattern. Thanks for the tip!

PR #1197, opened 2023-03-01
Remove the ability to delete a flow from UI

cicdw:
Currently flows can be deleted from both the UI and the API; however, flows are emergent concepts that are implicitly created whenever runs that reference them are created.  Additionally, flows as an API concept maintain no state (as opposed to deployments, for example, which contain configuration).  Deleting a flow therefore does not _do_ anything, as the flow could re-appear on the next run.  This PR is a part of ongoing reliability work to ensure our object relationships and their mechanics are principled and don't offer ambiguous user experiences.


znicholasbrown (approved):
all red, my favorite

PR #1198, opened 2023-03-01
Chore: Artifact mocker and types update

znicholasbrown:
This PR:
- adds support for 4 artifact types: `Result`, `Table`, `Markdown`, and `Unknown` (a catch-all for forwards/backwards compatibility)
- adds markdown mockers
- adds artifact mockers (including basic support for the data fields of known artifact types listed above)
- updates the experimental artifact api endpoint
- fixes filter references in the artifact api

### DIFF
--- a/src/mocks/artifact.ts
+++ b/src/mocks/artifact.ts
@@ -1,18 +1,39 @@
-import { Artifact } from '@/models'
+import { Artifact, ArtifactType } from '@/models'
 import { MockFunction } from '@/services/Mocker'
 import { choice } from '@/utilities'
 
 export const randomArtifact: MockFunction<Artifact, [Partial<Artifact>?]> = function(overrides = {}) {
+  const type = choice<ArtifactType>([ArtifactType.Markdown, ArtifactType.Result, ArtifactType.Table])
+  let data = null
+
+  switch (type) {
+    case ArtifactType.Result:
+      data = {
+        type: choice(['literal']),
+        value: this.create('string'),
+      }
+      break
+    case ArtifactType.Markdown:
+      data = this.create('markdownString', [{ sections: 4 }])
+      break
+    case ArtifactType.Table:
+      // TODO: Figure out table data type
+      data = []
+      break
+    default:
+      break
+  }
+
   return new Artifact({
     id: this.create('string'),
     created: this.create('date'),
     updated: this.create('date'),
     key: choice([null, this.create('noun')]),
-    type: this.create('string'),
-    description: this.create('markdownString', [{ sections: 2 }]),
+    type,
+    description: this.create('markdownString', [{ sections: 1 }]),
     flowRunId: this.create('id'),
     taskRunId: this.create('id'),
-    data: { [this.create('noun')]: this.create('string') },
+    data,
     metadata: {},
     ...overrides,
   })
--- a/src/mocks/artifact.ts
+++ b/src/mocks/artifact.ts
@@ -1,0 +1,19 @@
+import { Artifact } from '@/models'
+import { MockFunction } from '@/services/Mocker'
+import { choice } from '@/utilities'
+
+export const randomArtifact: MockFunction<Artifact, [Partial<Artifact>?]> = function(overrides = {}) {
+  return new Artifact({
+    id: this.create('string'),
+    created: this.create('date'),
+    updated: this.create('date'),
+    key: choice([null, this.create('noun')]),
+    type: this.create('string'),
+    description: this.create('markdownString', [{ sections: 2 }]),
+    flowRunId: this.create('id'),
+    taskRunId: this.create('id'),
+    data: { [this.create('noun')]: this.create('string') },
+    metadata: {},
+    ...overrides,
+  })
+}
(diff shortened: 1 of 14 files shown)


src/models/Artifact.ts
```diff
@@ -1,34 +1,70 @@
-export type ArtifactData = Record<string, unknown>
+export enum ArtifactType {
+  Result = 'result',
+  Markdown = 'markdown',
+  Table = 'table',
+  Unknown = 'unknown'
+}
```
pleek91, line 6: Can we use a string union rather than an enum? We've avoided using enums because they have weird side effects.

src/mocks/artifact.ts
```diff
@@ -0,0 +1,40 @@
...
+import { choice } from '@/utilities'
+
+export const randomArtifact: MockFunction<Artifact, [Partial<Artifact>?]> = function(overrides = {}) {
+  const type = choice<ArtifactType>([ArtifactType.Markdown, ArtifactType.Result, ArtifactType.Table])
+  let data = null
+
+  switch (type) {
+    case ArtifactType.Result:
+      data = {
+        type: choice(['literal']),
```
pleek91, line 12: choice from a single value?

src/types/artifact.ts
```diff
@@ -0,0 +1,33 @@
...
+  return artifact.type === ArtifactType.Table
+}
+
+export function isKnownArtifactType(type: string): type is ArtifactType {
+  return Object.values(ArtifactType).includes(type as ArtifactType)
+}
+
+export function isUnknownArtifact(artifact: Artifact): artifact is UnknownArtifact {
+  return artifact.type === ArtifactType.Unknown
+}
```
pleek91, line 21: These are a little odd to me. ArtifactType includes `unknown` as a type. So `isKnownArtifactType` will return true if the type being checked is `unknown`. And `isUnknownArtifact` will return false if a valid type is passed. Both of which make sense on their own. But in context of each other they appear to be inverse checks but they're not.

src/types/artifact.ts
```diff
@@ -0,0 +1,33 @@
+import { ArtifactType, ResultArtifact, ResultArtifactData, MarkdownArtifact, MarkdownArtifactData, UnknownArtifact, Artifact, ArtifactData, TableArtifact, TableArtifactData } from '@/models/Artifact'
+
+export function isResultArtifact(artifact: Artifact): artifact is ResultArtifact {
```
pleek91, line 3: Just a thought.

```typescript
export function isArtifactType<T extends ArtifactType>(artifact: Artifact, type: T): artifact is Artifact & { type: T }
```

src/models/Artifact.ts
znicholasbrown: can you expand on the weird side effects piece? We have type guards so this isn't strictly necessary but it's certainly a cleaner expectation

src/mocks/artifact.ts
znicholasbrown: it's the illusion of choice, really

src/models/Artifact.ts
pleek91: Sadly a video is the best explanation I can give. https://www.youtube.com/watch?v=jjMbPt_H3RQ

We currently intentionally haven't used enums in any of our projects and unless we have a good reason to do so I would rather be consistent and use an array const with a type. 

I also personally prefer string unions because with an enum you have to import the enum everywhere you want to use one of the values. A string union gives you all the type safety without the additional import.

src/types/artifact.ts
pleek91: Ah that helps. I think then all I would do is change the `isKnownArtifactType` to `type !== ArtifactType.unknown` Since the mapper is intentionally narrowing the type from `string` to `ArtifactType`.

src/models/Artifact.ts
znicholasbrown: Nice that's helpful - I'll make that update

src/mocks/artifact.ts
znicholasbrown: for real though I've got some more updates to this and there will be an actual choice when I write real mocks for result and table artifacts data - can ignore for now

src/types/artifact.ts
```diff
@@ -0,0 +1,21 @@
+import { ArtifactType, artifactTypes, ResultArtifactData, MarkdownArtifactData, Artifact, ArtifactData, TableArtifactData } from '@/models/Artifact'
+
+export function isArtifactType<T extends ArtifactType>(artifact: Artifact, type: T): artifact is Artifact & { type: T } {
+  return artifact.type === type
+}
+
+export function isKnownArtifactType(type: unknown): type is ArtifactType & Exclude<ArtifactType, 'unknown'> {
+  return typeof type === 'string' && artifactTypes.includes(type as ArtifactType) && type !== 'unknown'
```
pleek91, line 8: I thought the mapper ensured it was always an ArtifactType? So I expected this to not need the string type check or the includes.

pleek91 (approved):
One last question but approving because it's not blocking.

src/types/artifact.ts
znicholasbrown: yea the mapper does but I'm not sure a typeguard should presume the existence of a mapper, especially when we'd be passing in a primitive either way (the type string). If we were passing in a fully-hydrated artifact object i don't think the full type check would make sense but in that case the narrowing has already been done

pleek91: Oh duh cause it accepts an unknown value. In that case makes perfect sense. 

I think the return type can remove the first `ArtifactType &`

PR #1200, opened 2023-03-01
Allow graceful handling of flow runs with deleted deployment IDs

cicdw:
At some point we plan to remove cascading behavior between deployment deletion -> flow run table.  This means that we could end up with flow runs that have populated deployment IDs for deployments that no longer exist.

This PR adjusts the UI to handle that in a more robust way - to be clear, the UI was still working but this makes the display a bit cleaner.

Whoever reviews: I want to make sure that I haven't accidentally increased the number of lookups; I don't believe I have because we were already looking up deployment names for each individual flow run, but I wanted to still check!

### DIFF
--- a/src/components/FlowRunDetails.vue
+++ b/src/components/FlowRunDetails.vue
@@ -24,7 +24,7 @@
       </template>
     </p-key-value>
 
-    <p-key-value v-if="can.read.deployment && flowRun.deploymentId" label="Deployment" :alternate="alternate">
+    <p-key-value v-if="can.read.deployment && flowRun.deploymentId && deployment?.name" label="Deployment" :alternate="alternate">
       <template #value>
         <DeploymentIconText :deployment-id="flowRun.deploymentId" />
       </template>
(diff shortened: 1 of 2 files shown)


zangell44 (approved):

src/components/FlowRunDetails.vue
```diff
@@ -24,7 +24,7 @@
       </template>
     </p-key-value>
 
-    <p-key-value v-if="can.read.deployment && flowRun.deploymentId" label="Deployment" :alternate="alternate">
+    <p-key-value v-if="can.read.deployment && flowRun.deploymentId && deployment?.name" label="Deployment" :alternate="alternate">
```
znicholasbrown, line 27: Can you move this to its own computed variable in the script? 
```ts
const showDeployment = computed(() => can.read.deployment && props.flowRun.deploymentId && deployment.value?.name)
```
ought to do it; it helps with template readability as these conditionals get longer and prevents syntax highlighting in web editors from breaking on the optional chain

znicholasbrown: Also you probably don't actually care about checking that the name is defined (here or in the other file), you only care about the existence of the deployment at all

znicholasbrown (commented):
Just a small comment @cicdw

src/components/FlowRunDetails.vue
cicdw: I wasn't seeing the right behavior if I just did `&& deployment` -- what could I do differently to avoid the arbitrary choice of checking on `name`?

znicholasbrown: Try:
```ts
import { isDefined } from '@/utilities'

const showDeployment = computed(() => can.read.deployment && isDefined(deployment.value))
```

pleek91:
> Whoever reviews: I want to make sure that I haven't accidentally increased the number of lookups; I don't believe I have because we were already looking up deployment names for each individual flow run, but I wanted to still check!

@cicdw you are correct. Since the deployment is already being looked up in the `DeploymentIconText` component that request will be shared and no additional requests will be made.

src/components/FlowRunDetails.vue
cicdw: Done

znicholasbrown (approved):

PR #1201, opened 2023-03-02
Feature: Artifact descriptions

znicholasbrown:
Adds an `ArtifactDescription` component that wraps `p-markdown-renderer` in a scrollable container and renders the contents.

Adds some demo boilerplate.


brandonreid (approved):
Looks like there is a lint warning in ArtifactDescription, otherwise lgtm

PR #1206, opened 2023-03-06
Enhancement: Allow WorkPoolQueueMenu to be injected

pleek91:
# Description
In cloud we need to inject and override the WorkPoolQueueMenu in order to add the "automate" menu item. Updating components that use this component to implement the `useComponent` composition rather than importing it directly.


brandonreid (approved):

znicholasbrown (approved):
glad we're getting some mileage out of this pattern

PR #1207, opened 2023-03-06
Feature: `ArtifactCard` component

znicholasbrown:
Adds a new `ArtifactCard` component used for both lists and summary displays on artifact (results) pages


src/utilities/dates.ts
```diff
@@ -6,7 +6,7 @@ import { dateFunctions, toDate, formatDateInTimezone } from '@/utilities/timezon
 // https://www.notion.so/prefect/Formatting-Basics-1d4d83bcb2ba471ead90910aeb5913b2?pvs=4#b79673ff9e9b47f58236c449e5fe764a
 const dateFormat = 'MMM do, yyyy'
 const timeFormat = 'hh:mm a'
-const dateTimeFormat = `${dateFormat} ''at'' ${timeFormat}`
+const dateTimeFormat = `${dateFormat} 'at' ${timeFormat}`
```
znicholasbrown, line 9: @pleek91 I think this format wasn't quite correct before but wanted to call it out just in case

pleek91: I'm not and this wasn't used anywhere when I added it. I misunderstood the date-fns docs example. Thanks for fixing.

src/components/ArtifactCard.vue
```diff
@@ -0,0 +1,232 @@
+<template>
+  <p-card class="artifact-card" :class="classes.root">
+    <div class="artifact-card__body" :class="classes.body">
+      <header class="artifact-card__header" :class="classes.header">
+        <h6 class="artifact-card__subheader" :class="classes.subheader">
+          {{ artifact.type }}
+        </h6>
+        <h3 v-if="hasKey">
+          {{ artifact.key }}
+        </h3>
```
pleek91, line 10: I don't know what this should be but an `h6` followed by an `h3` seems a little off to me? I thought generally headings were supposed to be in order semantically. Could be way off and this is fine.

src/components/ArtifactCard.vue
```diff
@@ -0,0 +1,232 @@
...
+
+        <template v-if="hasRun && crumbs.length">
+          <p-bread-crumbs :crumbs="crumbs" class="artifact-card__bread-crumbs" :class="classes.breadCrumbs" />
+        </template>
+      </header>
+
+      <div class="artifact-card__summary-container" :class="classes.summaryContainer">
+        <div
+          class="artifact-card__summary-item"
+          :title="artifact.updated.toLocaleString()"
```
pleek91, line 20: What's the thought behind these titles attrs? The same info is present on the screen. Won't a screen reader end up reading this date twice?

src/components/ArtifactCard.vue
```diff
@@ -0,0 +1,232 @@
...
+  const hasKey = computed(() => !!props.artifact.key)
+
+  const crumbs = computed<BreadCrumbs>(() => {
+    const internalCrumbs: BreadCrumbs = []
+
+    if (flowRunId.value && !routeIncludes.value.flowRun) {
+      internalCrumbs.push({
+        text: flowRun.value?.name ?? '',
+        to: routes.flowRun(flowRunId.value),
+      })
```
pleek91, line 84: Switching this around a bit will get rid of some checks. (untested, but pretty sure this is correct)
```suggestion
    if (flowRun.value && !routeIncludes.value.flowRun) {
      internalCrumbs.push({
        text: flowRun.value.name,
        to: routes.flowRun(flowRun.value.id),
      })
```

src/components/ArtifactCard.vue
```diff
@@ -0,0 +1,232 @@
...
+    }
+
+    if (taskRunId.value && !routeIncludes.value.taskRun) {
+      internalCrumbs.push({
+        text: taskRun.value?.name ?? '',
+        to: routes.taskRun(taskRunId.value),
+      })
+    }
```
pleek91, line 92: same here

src/components/ArtifactCard.vue
```diff
@@ -0,0 +1,232 @@
...
+  const flowRunId = computed(() => props.artifact.flowRunId)
+  const taskRunId = computed(() => props.artifact.taskRunId)
+
+  const flowRun = useFlowRun(flowRunId)
+  const taskRun = useTaskRun(taskRunId)
+
+  const hasRun = computed(() => !!props.artifact.flowRunId || !!props.artifact.taskRunId)
+  const hasKey = computed(() => !!props.artifact.key)
+
+  const crumbs = computed<BreadCrumbs>(() => {
```
pleek91, line 77: So with flow runs when we show crumbs we omit the flow if we're on the flow page. But here you're doing the opposite. 

I'm probably missing some context. But won't all the artifacts have the same crumbs?

pleek91:
I might stack these dates on the condensed view so that things line up a little better. 
image

src/utilities/dates.ts
znicholasbrown: In your defense the example isn't great 😆

src/components/ArtifactCard.vue
znicholasbrown: You know I can't find anything about that but I'm also not confident this is semantically helpful - I've dumped both

src/components/ArtifactCard.vue
znicholasbrown: I actually think we should be doing this more but I've removed this instance in favor of something holistic down the line. I don't think this is an issue for screen readers in general (I might be wrong here) but the idea is to provide a locale-compliant browser-supplied timestamp in addition to whatever our display format is. This is particularly helpful in places where we use a relative datetime format (this used to which is what spurred the idea)

pleek91: Ah yes for relative times that would totally make sense 👍

src/components/ArtifactCard.vue
znicholasbrown: Ah nice suggestion 👍🏻

src/components/ArtifactCard.vue
znicholasbrown: Ah maybe this isn't as readable as it should be but the idea is exactly the same: don't show crumbs for pages we're on. Starting with no crumbs, we check if the flow run exists and that its ID isn't in the current route:
```ts
    if (flowRun.value && !routeIncludes.value.flowRun) {
      internalCrumbs.push({
        text: flowRun.value.name ?? '',
        to: routes.flowRun(flowRun.value.id),
      })
    }
```

znicholasbrown: On the flow run page, for instance, all the flow run crumbs would be the same (and so they would be omitted) but the task run crumbs would be different

pleek91: Ah okay that makes sense. I was just reading this wrong. 

I might break up the `routeIncludes` computed into two checks that are similar to `const viewingFlowRun = computed(() => ...)`. Might make it a little more readable

znicholasbrown: Makes sense - modified that

pleek91 (approved):

PR #1208, opened 2023-03-06
Bugfix: Filter hotkey events on flow run timeline 

znicholasbrown:
This PR fixes an issue where html input keypresses were masked by flow run timeline shortcuts.


pleek91 (approved):
Works for me. These keyboard shortcuts are a little overreaching imo and could probably just be removed in favor of a full screen route.

PR #1213, opened 2023-03-08
Feature: Flow Run Results

znicholasbrown:
This PR adds a few new components in service of flow run results:
- `FlowRunResults` - responsible for fetching, filtering, and displaying `result` artifacts as they relate to a flow run
- `ArtifactDataView` - responsible for determining how to render the `data` field of an artifact
- `ArtifactDataMarkdown` - renders the string of a `markdown` artifact
- `ArtifactDataResult` - renders the key/value pairs of a `result` artifact
- `ArtifactDataUnknown` - renders the `data` field with no extra formatting/components; unimplemented artifact types will fall back to this component
- `ArtifactMenu` - overflow menu used in artifact page headings
- `ArtifactPageHeading` - page heading lol

This also implements a mock artifacts API and augments some of the existing mockers


demo/sections/flowRuns/FlowRunResults.vue
```diff
@@ -0,0 +1,34 @@
...
+
+  const flowRun = useFlowRunMock()
+  useTaskRunsMock(10, { flowRunId: flowRun.id })
+
+  const flowRunNoResult = useFlowRunMock()
+  const artifact = data.artifacts.getAll().find(artifact => artifact.flowRunId === flowRunNoResult.id)
+
+  if (artifact) {
+    data.artifacts.delete(artifact.id)
+  }
```
pleek91, line 33: Why seed, then find the artifact and then delete it?

demo/sections/flowRuns/FlowRunResults.vue
```diff
@@ -0,0 +1,34 @@
...
+    {
+      title: 'No Results',
+    },
+  ]
+
+  const flowRun = useFlowRunMock()
+  useTaskRunsMock(10, { flowRunId: flowRun.id })
+
+  const flowRunNoResult = useFlowRunMock()
+  const artifact = data.artifacts.getAll().find(artifact => artifact.flowRunId === flowRunNoResult.id)
```
pleek91, line 29: I think DataStore has a find method

src/components/ArtifactCard.vue
```diff
@@ -132,6 +132,8 @@
 .artifact-card__body { @apply
   flex
   flex-col
+  h-full
```
pleek91, line 135: I don't remember exactly what css this is but why set the height on this component? Usually a component like this would have height auto

src/components/ArtifactDataMarkdown.vue
```diff
@@ -0,0 +1,21 @@
...
+  }>()
+
+  const data = computed<string>(() => props.artifact.data)
+</script>
+
+<style>
+.artifact-data-markdown { @apply
+  max-w-full
+  overflow-auto
+}
```
pleek91, line 20: Should the markdown renderer behave like this already?

src/components/ArtifactDataUnknown.vue
```diff
@@ -0,0 +1,23 @@
+<template>
+  <div class="artifact-data-unknown">
+    {{ data }}
```
pleek91, line 3: This is effectively raw text?

src/components/ArtifactDataView.vue
```diff
@@ -0,0 +1,30 @@
...
+    artifact: Artifact,
+  }>()
+
+  const component = computed(() => {
+    switch (props.artifact.type) {
+      case 'result':
+        return ArtifactDataResult
+      case 'markdown':
+        return ArtifactDataMarkdown
+      case 'table':
```
pleek91, line 24: Probably already considered this but couldn't we use a p-table?

src/components/FlowRunResults.vue
```diff
@@ -0,0 +1,107 @@
+<template>
+  <div class="flow-run-results">
+    <div v-if="!noResults" class="flow-run-results__button-group-container">
```
pleek91, line 3: Super nit: making the computed `hasResults` would prevent the double negative of `!noResults` and be a bit more readable.

src/localization/locale/en.ts
```diff
@@ -86,6 +86,8 @@ export const en = {
     updateWorkQueue: 'Work queue updated',
   },
   info: {
+    artifact: 'Artifact',
+    noResults: 'No tracked results',
```
pleek91, line 90: Curious why we sometimes refer to artifacts as "results". "Tracked" is also a little confusing to me. Are there untracked results?

pleek91 (approved):
Overall really clean. Few comments but nothing blocking.

demo/sections/flowRuns/FlowRunResults.vue
znicholasbrown: Recording this for posterity: it's because the mocker for flow runs currently auto-generates a result artifact for every run but isn't overridable; this was an easy way of accomplishing a "no result" view without modifying the mocker

src/components/ArtifactDataMarkdown.vue
znicholasbrown: Perhaps but I didn't want to rely on that since it's technically not part of the component contract

src/components/ArtifactDataView.vue
znicholasbrown: Discussed synchronously but short answer yes but we don't have a data structure contract to enforce here yet so this will fallback to the unknown renderer until we do

src/components/ArtifactDataMarkdown.vue
pleek91: I'm just thinking generally block elements (and components) have height auto by default. No issue, just curious.

PR #1222, opened 2023-03-15
Bugfix: Sub schemas

znicholasbrown:
This PR removes `maxLevel` considerations for schema properties and ignores meta-defined rendering in favor of a property-first version that falls back to the same component.

Closes https://github.com/PrefectHQ/prefect/issues/7816

### DIFF
--- a/src/components/SchemaFormProperty.vue
+++ b/src/components/SchemaFormProperty.vue
@@ -18,10 +18,6 @@
   }>()
 
   const is = computed(() => {
-    if (schemaHas(props.property, 'meta')) {
-      return SchemaFormInput
-    }
-
     if (schemaHas(props.property, 'properties')) {
       return SchemaFormProperties
     }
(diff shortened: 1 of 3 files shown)


pleek91 (approved):
Seems like a logical fix. Makes me wonder if this is going to brake some other edge case though. But I cannot think of a specific case to test for.

zhen0:
Looks like it fixes the issue for me:

image

zhen0:
Could we link the relevant issue in the description so it gets closed: https://github.com/PrefectHQ/prefect/issues/7816

pleek91:
closes https://github.com/PrefectHQ/prefect/issues/7816

znicholasbrown:
Throwing a DNM label on this; @pleek91 and I chatted through this change and decided we want to lean into it further by removing max-level limitations as well. Will ask for re-approval once that's done (should be today)

pleek91 (approved):
Thanks @znicholasbrown !

PR #1223, opened 2023-03-16
Enhancement: Flow run results lists

znicholasbrown:
This PR:

- improves mock flow run and task run results artifact description generation (they now more closely resemble action descriptions)
- breaks flow run and task run results into their own lists
- improves some `ArtifactCard` designs
- adds result descriptions to the default slot of `ArtifactCard`s 
- adds an interval to the artifact subscription so results update in real time
- persists a user's list vs grid preference (just using local storage right now but that seemed light enough)
- adds some new demo sections

Screen Shot 2023-03-15 at 9 53 59 PM

Screen Shot 2023-03-15 at 9 54 08 PM

### DIFF
--- a/src/localization/locale/en.ts
+++ b/src/localization/locale/en.ts
@@ -88,6 +88,8 @@ export const en = {
   info: {
     artifact: 'Artifact',
     noResults: 'No tracked results',
+    flowRunResults: 'Flow run',
+    taskRunResults: 'Task runs',
     created: 'Created',
     lastUpdated: 'Last Updated',
     deprecatedWorkQueue: 'This work queue uses a deprecated tag-based approach to matching flow runs; it will continue to work but you can\'t modify it',
--- a/src/localization/locale/en.ts
+++ b/src/localization/locale/en.ts
@@ -88,8 +88,8 @@ export const en = {
   info: {
     artifact: 'Artifact',
     noResults: 'No tracked results',
-    flowRunResults: 'Flow run',
-    taskRunResults: 'Task runs',
+    flowRun: 'Flow run',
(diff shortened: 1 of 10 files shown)


src/localization/locale/en.ts
```diff
@@ -88,6 +88,8 @@ export const en = {
   info: {
     artifact: 'Artifact',
     noResults: 'No tracked results',
+    flowRunResults: 'Flow run',
+    taskRunResults: 'Task runs',
```
pleek91, line 92: Nit: As we start adding more of these we need to be careful about naming. If we ever add a second language the names will be important. I'd literally name these `flowRun` and `taskRuns` as these are pretty generic labels and aren't restricted to results.

pleek91:
Couple things I noticed while looking at the deploy preview on my phone. The grid view overlaps and doesn't scroll well. And in the list view the breadcrumbs overflow the card

image
image

znicholasbrown:
@pleek91 which browser are you seeing that in? I've got a pretty nice looking version in chrome desktop at 330px:

Screenshot 2023-03-16 at 11 51 41 AM

zhen0:
Small note here that the deploy preview looks good on my iphone now.  Looks like you're digging in further though? Let me know if I can help keep this one moving.

znicholasbrown:
Thanks @zhen0; the remember your selection feature doesn't seem to be working in the deploy preview but is working in a regular ole build. I've added some fallbacks so that we can feel comfortable releasing just in case that continues to fail downstream. 

@pleek91 I've added a cache-key wrapper

pleek91 (approved):

Issue #1229, opened 2023-03-17
Block create/edit forms don't allow for the selection of existing blocks


desertaxle:
When creating or editing blocks, the fields for the block are shown, but the ability to choose an existing one is not shown. This is problematic for block types with optional attributes that are block types, but the referenced block types have required fields — preventing the form's submission.

An example is the `KubernetesJob`/`KubernetesClusterConfig` relationship (screenshot below) since a user doesn't need to provide a cluster config if they run their agent within a cluster.
![Screenshot 2023-03-17 at 8 11 42 AM](https://user-images.githubusercontent.com/12350579/225914358-e671179e-6984-41c3-9cd4-626b157a38ad.png)

I think this behavior was introduced in https://github.com/PrefectHQ/prefect-ui-library/pull/1222.

PR #1230, opened 2023-03-17
BugFix: Nested blocks not using the block document input

pleek91:
# Description
This [PR](https://github.com/PrefectHQ/prefect-ui-library/pull/1222) introduced a bug where block references were being rendered as form inputs rather than a block document input. Fixing that by adding a check specifically for block properties. 

Closes [1229](https://github.com/PrefectHQ/prefect-ui-library/issues/1229)

Before
image

After
image

### DIFF
--- a/src/components/SchemaFormProperty.vue
+++ b/src/components/SchemaFormProperty.vue
@@ -18,6 +18,10 @@
   }>()
 
   const is = computed(() => {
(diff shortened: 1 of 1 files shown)


zhen0 (approved):
Tested locally and LGTM.  Thanks @pleek91

PR #1231, opened 2023-03-17
Enhancement: Allow the empty message for FlowRunFilteredList to be overridden at implementation

pleek91:
# Description
Adds a slot so oss and cloud and customize the message. Leaving the existing message as the default value of the slot until all the implementations have been updated.


pleek91:
First step for https://github.com/PrefectHQ/nebula-ui/issues/2438

brandonreid (approved):

zhen0 (approved):

PR #1232, opened 2023-03-20
Feature/sub flows on timeline

brandonreid:
Adds subflow runs to the Flow Run Timeline.

image

Note that the `TaskRunPanel` has been renamed to `FlowRunTimelineSelectionPanel` and updated to handle subflows as well as tasks.


src/components/FlowRunTimeline.vue
```diff
@@ -49,16 +49,20 @@
           :is-running="isRunning"
           :format-date-fns="formatDateFns"
           :theme="theme"
-          :selected-node-id="selectedNode"
-          @click="selectNode"
+          :sub-node-labels="subFlowRunLabels"
+          :selected-node-id="selectedNodeId"
+          :expanded-sub-nodes="expandedSubFlows"
+          @selection="selectNode"
+          @sub-flow-toggle="toggleSubFlow"
```
brandonreid, line 56: realizing this event should probably be called `sub-node-toggle`..

src/components/FlowRunTimelineSelectionPanel.vue
```diff
@@ -51,16 +53,42 @@
...
-      to: taskRunId.value ? routes.taskRun(taskRunId.value) : '',
-    },
-  ])
+  let dataSubscription = assignSubscription()
+
+  function assignSubscription(): ReturnType<typeof useSubscriptionWithDependencies> {
+    return useSubscriptionWithDependencies(
+      type.value === 'task'
+        ? api.taskRuns.getTaskRun
+        : api.flowRuns.getFlowRun,
```
znicholasbrown, line 64: Another example of what I mentioned above - this isn't a pattern I'd like to adopt.

src/components/FlowRunTimelineSelectionPanel.vue
```diff
@@ -71,8 +99,9 @@
 .task-run-panel { @apply
   border
   dark:border-background-600
-  bg-background-800
-  dark:bg-background
+  bg-background
+  bg-opacity-80
+  backdrop-blur
```
znicholasbrown, line 104: Why the blur and change in opacity? Isn't this panel still part of the normal layout?

src/components/FlowRunTimelineSelectionPanel.vue
```diff
@@ -51,16 +53,42 @@
 
   const api = useWorkspaceApi()
   const routes = useWorkspaceRoutes()
-  const taskRunIdArgs = computed<[string] | null>(() => taskRunId.value ? [taskRunId.value] : null)
-  const taskRunSubscription = useSubscriptionWithDependencies(api.taskRuns.getTaskRun, taskRunIdArgs)
-  const taskRun = computed(() => taskRunSubscription.response)
+  const subscriptionArgs = computed<[string] | null>(() => id.value ? [id.value] : null)
```
znicholasbrown, line 56: You shouldn't need this check or `useSubscriptionWithDependencies` where it's used later - your `id` is a require property on the component. If other places dictate that it actually isn't required, I don't think I understand what this component is meant to render, currently it won't render anything without an id.

src/components/FlowRunTimeline.vue
```diff
@@ -236,6 +242,119 @@
...
+      if (subFlowId) {
+        return subFlowId
+      }
+    }
+
+    return null
+  }
+
+  function removeNullStartsAndSort(nodes: TimelineNode[]): GraphTimelineNode[] {
+    return nodes.filter((node): node is GraphTimelineNode => isValidGraphTimelineNode(node))
```
znicholasbrown, line 301: If `isValidGraphTimelineNode` is a typeguard already, you shouldn't need to specify `node is GraphTimelineNode`, it should be inferred.

znicholasbrown (changes requested):
A couple of broader notes from my first pass, in addition to the inline comments:

- function and property names here need to be cleaned up - you're referencing "subFlow" and "subFlowRun" interchangeably, which is confusing (and incorrect)
- (not blocking) your types are really messy. a lot of it stems from the decision to diverge from the extant flow run graph types and then modify and extend the types in the `graphs` package. these could really use some attention across the board but i've added this as a non-blocking comment because it spans repos

src/components/FlowRunTimelineSelectionPanel.vue
brandonreid: I'm not sure which previous comment you're referring to here?

Are you apposed to the conditional call type?

src/components/FlowRunTimelineSelectionPanel.vue
brandonreid: In full screen it overlays the content:

image

src/components/FlowRunTimelineSelectionPanel.vue
brandonreid: So it appears the `id`, and the entire content really, is optional for the sake of animating the panel away. When a selection is cleared, the panel is animated out of view, during that time, the selection is null but the panel still needs to render. Thinking through other ways of handling this..

src/components/FlowRunTimelineSelectionPanel.vue
znicholasbrown: https://github.com/PrefectHQ/prefect-ui-library/pull/1232#discussion_r1142677140

This one, sorry, I wrote these comments side by side then added more; we should have a really good motivation to use a ternary subscription and I don't think this rises to the level

src/components/FlowRunTimeline.vue
```diff
@@ -236,6 +242,119 @@
...
+    const isValueVisible = expandedSubFlows.value.has(id)
+
+    if (isValueVisible) {
+      const subFlowRun = expandedSubFlows.value.get(id)
+      subFlowRun?.subscription.unsubscribe()
+      expandedSubFlows.value.delete(id)
+      return
+    }
+
+    const subFlowId = findSubFlowId(id)
```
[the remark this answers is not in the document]
brandonreid: Yeah the graph was referencing the node id, while here it needed the subFlowRunId.. Updated graph to use subFlowRunId.

src/components/FlowRunTimeline.vue
```diff
@@ -236,32 +243,128 @@
...
-    pending: getStateColor('--state-pending-600'),
-    failed: getStateColor('--state-failed-600'),
-    cancelled: getStateColor('--state-cancelled-600'),
-    crashed: getStateColor('--state-crashed-600'),
-    paused: getStateColor('--state-paused-600'),
+  const stateColors: Record<string, Record<'default' | 'hover', string>> = {
+    completed: {
+      default: getStateColor('--state-completed-600'),
+      hover: getStateColor('--state-completed-700'),
+    },
```
znicholasbrown, line 329: Seems like there's no need to write all of these out - can you map over one of our states arrays instead?

src/components/FlowRunTimelineSubFlowRunDetails.vue
```diff
@@ -0,0 +1,76 @@
+<template>
+  <template v-if="subFlowRun">
+    <page-heading :crumbs="crumbs" size="sm" class="mb-2">
+      <template #actions>
+        <slot />
+      </template>
+    </page-heading>
```
znicholasbrown, line 7: Is the page heading component what you're looking for here? I'd expect that to be used solely in the context of a page and not an aside; maybe this is ok, just not something I've seen done before and could have unexpected results if we do make any app-wide changes to headings

src/components/FlowRunTimelineSubFlowRunDetails.vue
```diff
@@ -0,0 +1,76 @@
...
+    <div class="timeline-sub-flow-details__content">
```
znicholasbrown, line 8: This class needs to be updated for the component (also inconsistent sub flow usage)

src/components/FlowRunTimelineSubFlowRunDetails.vue
```diff
@@ -0,0 +1,76 @@
...
+      <p-key-value label="State" :alternate="alternate">
+        <template #value>
+          <StateBadge :state="subFlowRun.state" class="timeline-task-details__state-badge" />
```
znicholasbrown, line 11: This class needs to be updated for the component

src/components/FlowRunTimelineSubFlowRunDetails.vue
```diff
@@ -0,0 +1,76 @@
...
+      <p-key-value label="Flow Run ID" :value="subFlowRun.id" :alternate="alternate" />
+      <p-key-value label="Duration" :alternate="alternate">
+        <template #value>
+          <DurationIconText :duration="subFlowRun.duration" />
+        </template>
+      </p-key-value>
+      <p-key-value label="Created" :value="formatDateTimeNumeric(subFlowRun.created)" :alternate="alternate" />
+      <p-key-value label="Tags" :alternate="alternate">
+        <template v-if="subFlowRun.tags?.length" #value>
+          <p-tags :tags="subFlowRun.tags!" class="timeline-sub-flow-details__tags" />
```
znicholasbrown, line 23: This class needs to be updated for the component (also inconsistent sub flow usage)

src/components/FlowRunTimelineSubFlowRunDetails.vue
```diff
@@ -0,0 +1,76 @@
...
+
+<script lang="ts" setup>
+  import { useSubscriptionWithDependencies } from '@prefecthq/vue-compositions'
+  import { computed, toRefs } from 'vue'
+  import { StateBadge, PageHeading, DurationIconText } from '@/components'
+  import { useWorkspaceApi, useWorkspaceRoutes } from '@/compositions'
+  import { formatDateTimeNumeric } from '@/utilities/dates'
+
+  const props = defineProps<{
+    id: string,
```
znicholasbrown, line 38:
```suggestion
    flowRunid: string,
```
It's usually better to be explicit with what id is being passed to this component and to avoid collision with the functional html id, which can be used as a selector. In this case it's also particularly important because of the way that we query for sub flow runs

src/components/FlowRunTimelineTaskDetails.vue
```diff
@@ -0,0 +1,76 @@
...
+
+<script lang="ts" setup>
+  import { useSubscriptionWithDependencies } from '@prefecthq/vue-compositions'
+  import { computed, toRefs } from 'vue'
+  import { StateBadge, PageHeading, DurationIconText } from '@/components'
+  import { useWorkspaceApi, useWorkspaceRoutes } from '@/compositions'
+  import { formatDateTimeNumeric } from '@/utilities/dates'
+
+  const props = defineProps<{
+    id: string,
```
znicholasbrown, line 38:
```suggestion
    taskRunId: string,
```
It's usually better to be explicit with what id is being passed to this component and to avoid collision with the functional html id, which can be used as a selector

src/components/FlowRunTimelineTaskDetails.vue
```diff
@@ -0,0 +1,76 @@
+<template>
+  <template v-if="task">
+    <page-heading :crumbs="crumbs" size="sm" class="mb-2">
+      <template #actions>
+        <slot />
+      </template>
+    </page-heading>
+    <div class="timeline-task-details__content">
```
znicholasbrown, line 8:
```suggestion
    <div class="flow-run-timeline-task-details__content">
```
BEM says this should contain the full name of the component

src/components/FlowRunTimelineTaskDetails.vue
```diff
@@ -0,0 +1,76 @@
...
+      <p-key-value label="State" :alternate="alternate">
+        <template #value>
+          <StateBadge :state="task.state" class="timeline-task-details__state-badge" />
```
znicholasbrown, line 11:
```suggestion
          <StateBadge :state="task.state" class="flow-run-timeline-task-details__state-badge" />
```
BEM says this should contain the full name of the component

znicholasbrown (changes requested):

src/components/FlowRunTimeline.vue
```diff
@@ -49,16 +49,20 @@
           :is-running="isRunning"
           :format-date-fns="formatDateFns"
           :theme="theme"
-          :selected-node-id="selectedNode"
-          @click="selectNode"
+          :sub-node-labels="subFlowRunLabels"
+          :selected-node-id="selectedNode?.id"
+          :expanded-sub-nodes="expandedSubFlowRuns"
+          @selection="selectNode"
```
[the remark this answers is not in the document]
brandonreid: I opted to make the selected state owned by the consumer. I could see it being appended to the URL so a link can be shared to a particular part, or maybe search functionality in the future results in a node selection etc. I might have an IOU for v-model here, I agree it might as well be, but unless I have another item sending me to update the graphs repo, I might do that in another PR so I can get this to staging.

src/components/FlowRunTimeline.vue
brandonreid: I haven't tested every color, I imagine something like a yellow might be tweeked at some point? Idk what the value of a map would really give here over a simple object?

src/components/FlowRunTimeline.vue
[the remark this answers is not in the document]
znicholasbrown: Yeah fair enough, no need to update the graphs repo, just commenting that's probably a pattern more in line with how you're using it

src/components/FlowRunTimelineSelectionPanel.vue
```diff
@@ -0,0 +1,66 @@
+<template>
+  <div :class="classes">
+    <FlowRunTimelineTaskDetails
+      v-if="selectedNode && selectedNode.type === 'task'"
+      :id="selectedNode.id"
+    >
+      <p-button size="xs" icon="XIcon" flat @click="closePanel" />
```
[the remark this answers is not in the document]
brandonreid: Yeah I could use some advice on that if you have another idea.. Essentially I want the close button to be inline with the header, but the header is being set in the child after the data is fetched. So I end up adding it to the `action` slot of the header component.. Will see if I can think of another way here..

image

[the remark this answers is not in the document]
znicholasbrown: I'd drop that requirement if possible or you risk misalignment/weird collisions as your title grows to fit that space or overflows it. Better to have those on separate lines probably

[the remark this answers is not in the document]
znicholasbrown: (it'll also make your code simpler, which is a big plus)

src/components/FlowRunTimeline.vue
brandonreid: ahh I see.. For some reason I couldn't find a state array before, but i see now: https://github.com/PrefectHQ/prefect-ui-library/blob/d055c46f25e52bf8a5fd1fffd874080675b9aacb/src/types/states.ts#L3

src/components/FlowRunTimelineSelectionPanel.vue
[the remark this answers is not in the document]
brandonreid: Yeah that makes sense. Will need to spend more time on these panels at a later date anyway.

src/components/FlowRunTimeline.vue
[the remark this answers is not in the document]
brandonreid: Will update this next graphs update for sure.

src/components/FlowRunTimelineSubFlowRunDetails.vue
brandonreid: Ok, swapped this for `p-heading`

PR #1233, opened 2023-03-20
Enhancement: No results messages link to docs

znicholasbrown:
Note for reviewers: decided to try something slightly new here, using the markdown renderer for localization content; seems to work nicely but also open to other suggestions.

### DIFF
--- a/src/localization/locale/en.ts
+++ b/src/localization/locale/en.ts
@@ -7,10 +7,11 @@ export const en = {
     workQueues: 'https://docs.prefect.io/ui/work-queues/',
     notifications: 'https://docs.prefect.io/ui/notifications/',
     deployments: 'https://docs.prefect.io/ui/deployments/',
-    concurrency: 'https://docs.prefect.io/concepts/tasks/?h=conc#task-run-concurrency-limits',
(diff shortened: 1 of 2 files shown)


src/localization/locale/en.ts
```diff
@@ -87,7 +88,7 @@ export const en = {
   },
   info: {
     artifact: 'Artifact',
-    noResults: 'No tracked results',
+    noResults: 'No tracked results, enable [result persistence](https://docs.prefect.io/concepts/results/#persisting-results) to track results.',
```
madkinsz, line 91: This only affects old client versions basically right?

madkinsz: Won't they need to upgrade to a version with result artifacts? Maybe we should just suggest being on the latest version.

znicholasbrown: Afaik, yeah. It'll show up for empty states for task runs or the flow run itself as well which at the very least is a helpful "what this feature is"

pleek91:
@znicholasbrown does the markdown rendered look the same as if we did this with html like we normally would? I'm thinking the link. Does that use `p-link` and get the external url icon? 

Otherwise I love this idea.

pleek91 (approved):

Issue #824, opened 2022-11-28
Card from scatterPlot is missing new styling


marichka-offen:
The card that appears when you hover over a dot on scatterPlot is missing new link styling with `>`
<img width="1624" alt="Screenshot 2022-11-28 at 11 56 15 AM" src="https://user-images.githubusercontent.com/40467112/204336727-625c81cf-56d9-4c92-8fde-6b6594e62007.png">

PR #1239, opened 2023-03-21
Enhancement: Make flow run bread crumbs consistent

pleek91:
# Description

Closes https://github.com/PrefectHQ/prefect-ui-library/issues/824

image

### DIFF
--- a/src/components/FlowRunBreadCrumbs.vue
+++ b/src/components/FlowRunBreadCrumbs.vue
@@ -1,0 +1,38 @@
+<template>
+  <div class="flow-run-bread-crumbs">
+    <FlowRouterLink :flow-id="flowRun.flowId" class="flow-run-bread-crumbs__flow-link">
+      <template #after>
+        <p-icon icon="ChevronRightIcon" size="small" />
+      </template>
+    </FlowRouterLink>
+    <p-link :to="routes.flowRun(flowRun.id)" class="flow-run-bread-crumbs__flow-run-link">
+      <span>{{ flowRun.name }}</span>
+    </p-link>
+  </div>
+</template>
+
+<script lang="ts" setup>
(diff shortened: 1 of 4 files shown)


src/components/FlowRunBreadCrumbs.vue
```diff
@@ -0,0 +1,38 @@
...
+  defineProps<{
+    flowRun: FlowRun,
+  }>()
+
+  const routes = useWorkspaceRoutes()
+</script>
+
+<style>
+.flow-run-bread-crumbs { @apply
+  text-slate-700
```
znicholasbrown, line 28:
```suggestion
  text-foreground-600
```
I think you're just looking for bolded text here and not this color specifically, right?

znicholasbrown (approved):

src/components/FlowRunBreadCrumbs.vue
pleek91: This is the color that the `StateListItem` uses which is where this was getting its color from. Adding it here so this bread crumb looks as it does currently in the list item in the popover. Not intending to change any styles. This applies to the task run list item as well. 

image

pleek91: foreground-600 is to bright IMO. 
image

200 looks good to me. I could change both this component and the StateRunListItem if you'd like
image

znicholasbrown: Oh it's wrong there too lol - we just haven't noticed because all the important text there is links that are highlighted in blue

znicholasbrown: Yes please, that looks great

pleek91: Updated 👍

znicholasbrown (approved):

PR #1240, opened 2023-03-21
Enhancement: Validate any schema property that is a JsonInput has valid json

pleek91:
# Description
Closes https://github.com/PrefectHQ/prefect/issues/7623

Makes sure any json input is validated as valid json before submitting
image

### DIFF
--- a/src/services/schemas/properties/SchemaPropertyService.ts
+++ b/src/services/schemas/properties/SchemaPropertyService.ts
@@ -1,4 +1,5 @@
 import { SelectOption } from '@prefecthq/prefect-design'
+import { JsonInput } from '@/index'
 import { InvalidSchemaValueError } from '@/models/InvalidSchemaValueError'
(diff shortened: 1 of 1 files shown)


src/services/schemas/properties/SchemaPropertyService.ts
```diff
@@ -99,6 +100,10 @@ export abstract class SchemaPropertyService {
     const defaults = getSchemaPropertyDefaultValidators(this.property, required)
     const validators = fieldRules(title, ...this.validators)
 
+    if (this.componentIs(JsonInput)) {
+      validators.push(isJson(title))
```
znicholasbrown, line 104: In case anyone else ends up wondering, `isJson` gracefully handles null values:
```
  if (isNullish(value) || isEmptyString(value)) {
    return true
  }
```

znicholasbrown (approved):

zhen0:
This looks nice to me but I read the issue it's closing as being about identifying the json input box. For example in some blocks, it's not clear that json is needed. image

pleek91:
@zhen0 agreed. But I think this solves the primary issue of being able to save the block with invalid data. This is now consistent with the rest of the app where we use json. We might want to make json inputs as a whole more intuitive.

PR #1246, opened 2023-03-22
Feature: `ArtifactCollectionTimeline` and `ArtifactCollectionTimelineItem` components

znicholasbrown:
Introduces 2 WIP components we'll be using for showing lineage-type data for artifacts. The `ArtifactCollectionTimelineItem` I expect to change/evolve quite a bit in the next week so I've intentionally not spent much time cleaning it up. I also took a crack at filters for the artifacts mock api that should work nicely for gets and counts

### DIFF
--- a/demo/utilities/data.ts
+++ b/demo/utilities/data.ts
@@ -22,7 +22,7 @@ function hydrateGraph({ id, graph }: FlowRunGraphMock): FlowRunGraphMock {
 // eslint-disable-next-line @typescript-eslint/explicit-function-return-type
 export function createDataStores(seeds: ApiMockSeeds = {}) {
   return {
-    artifacts: new KeyedDataStore({ seeds: seeds.artifacts, hydrate: artifact => new Artifact(artifact) }),
+    artifacts: new KeyedDataStore({ seeds: seeds.artifacts, hydrate: artifact => new Artifact({ ...artifact, created: new Date(artifact.created), updated: new Date(artifact.updated) }) }),
     flows: new KeyedDataStore({ seeds: seeds.flows, hydrate: flow => new Flow(flow) }),
     flowRuns: new KeyedDataStore({ seeds: seeds.flowRuns, hydrate: flowRun => new FlowRun(flowRun) }),
     flowRunGraphs: new KeyedDataStore({ seeds: seeds.flowRunGraphs, hydrate: hydrateGraph }),
--- a/src/components/ArtifactTimeline.vue
+++ b/src/components/ArtifactTimeline.vue
@@ -1,12 +1,18 @@
 <template>
-  <div class="artifact-timeline">
-    {{ artifacts }}
-  </div>
+  <p-virtual-scroller v-if="executed" :items="artifacts" :item-estimate-height="112" class="artifact-timeline" @bottom="fetchMore">
+    <template #default="{ item: artifact, index }">
+      <p-heading heading="3">
+        {{ index }} - {{ artifact.id }}
+      </p-heading>
+      <ArtifactTimelineItem v-bind="{ artifact }" />
+    </template>
+  </p-virtual-scroller>
 </template>
 
 <script lang="ts" setup>
   import { useSubscription } from '@prefecthq/vue-compositions'
-  import { computed } from 'vue'
+  import { computed, ref } from 'vue'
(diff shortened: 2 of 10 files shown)


zhen0:
Not necessary for this PR but I wonder if we could make it more obvious that the artifacts are clickable and expandable.  I didn't realize I could do that until I saw your preview video.

zhen0 (approved):
Thanks @znicholasbrown

demo/utilities/data.ts
```diff
@@ -22,7 +22,7 @@ function hydrateGraph({ id, graph }: FlowRunGraphMock): FlowRunGraphMock {
 // eslint-disable-next-line @typescript-eslint/explicit-function-return-type
 export function createDataStores(seeds: ApiMockSeeds = {}) {
   return {
-    artifacts: new KeyedDataStore({ seeds: seeds.artifacts, hydrate: artifact => new Artifact(artifact) }),
+    artifacts: new KeyedDataStore({ seeds: seeds.artifacts, hydrate: artifact => new Artifact({ ...artifact, created: new Date(artifact.created), updated: new Date(artifact.updated) }) }),
```
pleek91, line 25: This makes sense to me since dates in json don't get parsed automatically. But it makes me wonder how all the other dates in these mocks aren't breaking things 🤔

src/components/ArtifactTimeline.vue
```diff
@@ -0,0 +1,98 @@
...
+    <p-virtual-scroller
+      :items="artifacts"
+      :item-estimate-height="112"
+      class="artifact-timeline__virtual-scroller"
+      @bottom="fetchMore"
+    >
+      <template #default="{ item: artifact }">
+        <ArtifactTimelineItem
+          v-bind="{ artifact }"
+          v-model:expanded="expanded"
```
pleek91, line 12: Curious why the list needs to know which items are expanded. It doesn't appear to be doing anything based on this. Could this just be internal to the items themselves?

src/components/ArtifactTimeline.vue
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  watch(artifactsFilterOffset, getArtifacts)
+
+  const fetchMore = (): void => {
+    if (artifacts.value.length >= artifactsCount.value) {
+      return
+    }
+    artifactsFilterOffset.value += ARTIFACTS_DEFAULT_FILTER_LIMIT
+  }
```
pleek91, line 72: Rather than watching `artifactsFilterOffset` why not just call `getArtifacts` in the `fetchMore` function?

src/components/ArtifactTimelineItem.vue
```diff
@@ -0,0 +1,238 @@
...
+        expandedModel.value = expandedModel.value.filter((value) => value !== props.value)
+      } else {
+        expandedModel.value = [...expandedModel.value, props.value]
+      }
+
+      return
+    }
+
+    expandedModel.value = !expandedModel.value
+  }
```
pleek91, line 119: All this can go away if each item has its own boolean for expanded right?

src/components/ArtifactTimelineItem.vue
```diff
@@ -0,0 +1,238 @@
...
+  transition-all
+  text-sm
+}
+
+.artifact-timeline-item__card-heading--expanded,
+.artifact-timeline-item__card-body { @apply
+  p-4
+}
+
+.artifact-timeline-item__card:not(.artifact-timeline-item__card--expanded) { @apply
```
pleek91, line 228: Personally I'd rather see a `artifact-timeline-item__card--collapsed` modifier class than these more complex selectors.

demo/utilities/data.ts
znicholasbrown: Yeah this was surprising to me as well; it's possible we just aren't using any built in date methods anywhere and the strings we're passing around are perfectly fine for most of our date utilities

brandonreid:
Super slick! Really like the snappiness of the animations. I didn't actually discover the clickability until I saw Jenny's comment, agree that more prominent hover styles would be helpful, and/or some sort of semi-pseudo toggle button (`▼`).

brandonreid:
Super slick! Really like the snappiness of the animations.

PR #1257, opened 2023-03-23
Enhancement: Batch block document get requests

pleek91:
# Description
Enables request batching for individual block document requests. Something to note is the `isAnonymous: null` part of the filter. This ensures that both anonymous and non anonymous blocks are returned.

### DIFF
--- a/src/services/WorkspaceBlockDocumentsApi.ts
+++ b/src/services/WorkspaceBlockDocumentsApi.ts
@@ -3,8 +3,10 @@ import { BlockDocument } from '@/models/BlockDocument'
 import { BlockDocumentCreate } from '@/models/BlockDocumentCreate'
 import { BlockDocumentUpdate } from '@/models/BlockDocumentUpdate'
 import { BlockDocumentsFilter } from '@/models/Filters'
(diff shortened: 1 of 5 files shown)


zhen0:
In case you've not already seen, it looks like we need a type update for this.

pleek91:
> In case you've not already seen, it looks like we need a type update for this. 

I hadn't. Thank you!

src/services/WorkspaceBlockDocumentsApi.ts
```diff
@@ -18,10 +20,19 @@ export class WorkspaceBlockDocumentsApi extends WorkspaceApi implements IWorkspa
...
+    const blockDocuments = await this.getBlockDocuments({
+      blockDocuments: {
+        id: ids,
+        isAnonymous: null,
+      },
+    })
 
-    return mapper.map('BlockDocumentResponse', data, 'BlockDocument')
+    return toMap(blockDocuments, 'id')
+  }, { maxBatchSize: 200 })
```
znicholasbrown, line 32: 👁️ 👁️

znicholasbrown (approved):
Ignore the eyes, we'll grab the batch size issue later

PR #1259, opened 2023-03-24
Feature: Artifact key routes

znicholasbrown:


src/router/index.ts
```diff
@@ -258,6 +258,11 @@ export function createWorkspaceRouteRecords(components: Partial<WorkspaceRouteCo
           path: '',
           component: components.artifacts,
         },
+        {
+          name: 'workspace.artifacts.artifact.key',
+          path: 'key/:key',
```
pleek91, line 263:
```suggestion
          path: 'key/:artifactKey',
```

pleek91 (approved):
One suggestion that might not be correct. But we try and keep generic terms out of the url params. Suggesting `artifactKey`

znicholasbrown:
Nah you're absolutely correct @pleek91 - updated

PR #1261, opened 2023-03-24
Feature: Artifacts collections utilities

znicholasbrown:
This PR adds several artifacts and results-related components/utilities.

- (new) `ArtifactCollections` component - used to display a list of keyed artifacts (with filtering)
- (new) `ArtifactCollectionsEmptyState` component - displayed when a user has 0 artifacts
- (new) `ArtifactTypeSelect` component - selects all or individual artifact types (e.g. `markdown`, `result`, `unknown` etc)
- (new) `PageHeadingArtifacts` component
- (new) `PageHeadingArtifactKey` component - heads the page for an individual artifact collection (a keyed artifact)
- (new) `ViewModeButtonGroup` component - consolidates the row/grid view button group logic into a single component
- (new) `RowGridLayoutList` component - takes a list of items and applies row/grid styling based on the active view mode; also provides an empty state for convenience
- (breaking change) the `artifactsViewMode` utility (composition?) has been trimmed quite a bit and converted to a more generic `activeViewMode` utility; this allows remembered and shared layout functionality
- (updated) `FlowRunResults` component now uses the shared layout functionality
- (updated) artifacts filters have been updated to include pending functionality to filter by type, key existence, and latest


src/components/ArtifactCollections.vue
```diff
@@ -0,0 +1,113 @@
+<template>
+  <div class="artifact-collections">
+    <div class="artifact-collections__filters">
+      <ResultsCount :label="localization.info.artifact" :count="artifactsCount" class="artifact-collection-list__results" />
+      <SearchInput v-model="searchTerm" :placeholder="localization.info.artifactSearch" :label="localization.info.artifactSearch" class="artifact-collection-list__search" />
+      <ArtifactTypeSelect v-model:selected="selectedType" class="artifact-collection-list__type" />
+      <ViewModeButtonGroup class="artifact-collections__view-mode-button-group" />
+    </div>
+
+    <template v-if="artifactsCountSubscription.executed && artifactsCount == 0">
```
znicholasbrown, line 10: Right now this doesn't take into account filters - will add that functionality at a later date when the `/counts` endpoint is working

src/components/ArtifactCollections.vue
```diff
@@ -0,0 +1,114 @@
...
+
+  const searchTerm = ref('')
+  const selectedType = ref<ArtifactType | null>(null)
+
+  const api = useWorkspaceApi()
+
+  const artifactsFilter = computed<ArtifactsFilter>(() => {
+    return {
+      artifacts: {
+        keyLike: searchTerm.value ? `%${searchTerm.value}%` : undefined,
```
pleek91: Is the api expecting this percent formatted string? Any other `like` filters that I've used accept a plain string without and tokens.

src/components/ArtifactCollections.vue
```diff
@@ -0,0 +1,114 @@
...
+      <ResultsCount :label="localization.info.artifact" :count="artifactsCount" class="artifact-collection-list__results" />
+      <SearchInput v-model="searchTerm" :placeholder="localization.info.artifactSearch" :label="localization.info.artifactSearch" class="artifact-collection-list__search" />
+      <ArtifactTypeSelect v-model:selected="selectedType" class="artifact-collection-list__type" />
+      <ViewModeButtonGroup class="artifact-collections__view-mode-button-group" />
+    </div>
+
+    <template v-if="artifactsCountSubscription.executed && artifactsCount == 0">
+      <ArtifactCollectionsEmptyState />
+    </template>
+    <template v-else-if="artifacts && artifacts.length">
```
pleek91, line 13: If you default artifacts to an empty array you can at least remove the first part of this check. Possible the whole condition?

src/components/ArtifactCollectionsEmptyState.vue
```diff
@@ -0,0 +1,34 @@
...
+  import { PEmptyState, PIcon } from '@prefecthq/prefect-design'
+  import DocumentationButton from '@/components/DocumentationButton.vue'
+  import { localization } from '@/localization'
+</script>
+
+<style>
+.artifact-collections-empty-state__icon { @apply
+  w-12
+  h-12
+}
```
pleek91, line 33: p-empty-state already has sizing for icons and we should let it do so. Helps keep our empty state icons consistent.

src/components/ArtifactTypeSelect.vue
```diff
@@ -0,0 +1,32 @@
...
+    },
+    set(value: string | null) {
+      emit('update:selected', value)
+    },
+  })
+
+  const options: (SelectOption & { value: ArtifactType | null })[] = [
+    { label: 'All', value: null },
+    ...artifactTypes.map((type) => ({ label: capitalize(type), value: type })),
+  ].sort((alpha, beta) => alpha.label.localeCompare(beta.label))
```
pleek91, line 31: super nit: this is almost code golf lol. Its a super small component so its fine. But if you wanted to I would break this out a bit.

src/components/FlowRunResults.vue
```diff
@@ -3,46 +3,46 @@
...
-        <p-button-group v-model="activeViewMode" :options="viewOptions" />
+        <ViewModeButtonGroup class="flow-run-results__view-mode-button-group" />
       </div>
 
       <p-heading heading="6" class="flow-run-results__subheading">
         {{ localization.info.flowRun }}
       </p-heading>
-      <div class="flow-run-results__list" :class="classes.list">
-        <template v-if="flowRunResults.length">
+      <template v-if="flowRunResults.length">
```
pleek91, line 12: I noticed all your RowGridLayout's have a v-if checking the array length. Is that necessary? Does it not handle being empty gracefully?

src/components/PageHeadingArtifactKey.vue
```diff
@@ -0,0 +1,41 @@
...
+  const props = defineProps<{
+    artifact: Artifact,
+  }>()
+
+  const routes = useWorkspaceRoutes()
+
+  const crumbs = computed<BreadCrumbs>(() => {
+    const internalCrumbs: BreadCrumbs = [
+      {
+        text: pluralize(localization.info.artifact),
```
pleek91, line 28: nit: this should probably just be its own localization string.

src/components/RowGridLayout.vue
pleek91, line 1: Thought: Considering the intent of this component is to format items. I would accept those items as a prop and do the v-for internally. That way empty arrays can be handled gracefully. The default slot would have the item bound to its slot scope.

src/components/ArtifactCollections.vue
pleek91: If that's the case, do you need to cast to undefined if its falsey? Or can you just send an empty string. OR better yet handle that in the mapper.

src/components/FlowRunResults.vue
znicholasbrown: Oh it does, these are to make sure we show the empty state if there are no results of that specific lineage (flow run / task run)

src/components/ArtifactTypeSelect.vue
znicholasbrown: when all i really wanted is:
```ts
const options: (SelectOption & { value: ArtifactType | null })[] = [
    { label: 'All', value: null },
    { label: 'Markdown', value: 'markdown' },
    { label: 'Table', value: 'table' },
  ]
```

src/components/FlowRunResults.vue
pleek91: Then my nit would be let the RowGridLayout do its thing (without the v-if) and add an `empty` computed and v-if that to show the empty state. Keeps our checks truthy and gets rid of a v-else in the template.

src/components/PageHeadingArtifactKey.vue
znicholasbrown: Agreed - added

src/components/ArtifactTypeSelect.vue
pleek91: That's so much cleaner haha. But I'm not sure it'll guarantee that if a new artifact type is added it'll get added here. So that's worth looking into

znicholasbrown:
Nice suggestions @pleek91 - updated

src/components/ArtifactCollections.vue
```diff
@@ -0,0 +1,116 @@
...
+  import { useSubscription } from '@prefecthq/vue-compositions'
+  import { computed, ref } from 'vue'
+  import { ArtifactTypeSelect, ResultsCount, SearchInput } from '@/components'
+  import ArtifactCard from '@/components/ArtifactCard.vue'
+  import ArtifactCollectionsEmptyState from '@/components/ArtifactCollectionsEmptyState.vue'
+  import RowGridLayoutList from '@/components/RowGridLayoutList.vue'
+  import ViewModeButtonGroup from '@/components/ViewModeButtonGroup.vue'
+  import { useWorkspaceApi } from '@/compositions'
+  import { localization } from '@/localization'
+  import { ArtifactsFilter, ArtifactType, Artifact } from '@/models'
```
pleek91, line 32:
```suggestion
  import { ArtifactsFilter, ArtifactType } from '@/models'
```

src/components/ArtifactCollections.vue
```diff
@@ -0,0 +1,116 @@
...
+    const likeFilter = searchTerm.value ? searchTerm.value : undefined
+    const typeFilter = selectedType.value ? [selectedType.value] : undefined
+
+    return {
+      artifacts: {
+        keyLike: likeFilter,
+        keyExists: true,
+        type: typeFilter,
+      },
+    }
```
pleek91, line 49: Love a good opportunity to use object shorthand declaration
```suggestion
    const keyLike = searchTerm.value ? searchTerm.value : undefined
    const type = selectedType.value ? [selectedType.value] : undefined

    return {
      artifacts: {
        keyExists: true,
        keyLike,
        type,
      },
    }
```

src/components/FlowRunResults.vue
```diff
@@ -56,26 +56,21 @@
...
-  import { ButtonGroupOption } from '@prefecthq/prefect-design'
   import { useSubscription } from '@prefecthq/vue-compositions'
   import { computed } from 'vue'
   import ArtifactCard from '@/components/ArtifactCard.vue'
+  import RowGridLayoutList from '@/components/RowGridLayoutList.vue'
+  import ViewModeButtonGroup from '@/components/ViewModeButtonGroup.vue'
   import { useWorkspaceApi } from '@/compositions'
   import { localization } from '@/localization'
-  import { FlowRun } from '@/models'
+  import { FlowRun, ResultArtifact } from '@/models'
```
pleek91, line 66:
```suggestion
  import { FlowRun } from '@/models'
```

src/components/RowGridLayoutList.vue
```diff
@@ -0,0 +1,53 @@
...
+  </div>
+</template>
+
+<script lang="ts" setup>
+  import { computed } from 'vue'
+  import { activeViewMode } from '@/utilities/activeViewMode'
+
+  const props = defineProps<{
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    items: any[],
```
pleek91, line 19: `unknown` doesn't work here?

pleek91: I can never keep straight when it won't work...

pleek91 (approved):
Approving but made a couple small suggestions. Coupling are just lint fixes.

src/components/ArtifactCollections.vue
znicholasbrown: This is used in the template, eslint is being a bum :(

src/components/RowGridLayoutList.vue
znicholasbrown: It doesn't, it's bugging me cause any isn't correct. TS throws an error when trying to cast/infer the type in the slot caller if this is `unknown`

src/components/FlowRunResults.vue
znicholasbrown: love it - idk why i thought `type` was a reserved word lol

PR #1262, opened 2023-03-27
Enhancement: Artifacts filters and collections

znicholasbrown:
This PR:
- moves run-related logic out of the `ArtifactCard` component in favor of type-specific implementations
- adds a new `ArtifactResultCard` to encapsulate run logic changes^
- modifies the `FlowRunResults` component to use the new `ArtifactResultCard` component instead of the base card
- updates workspace-routes to use artifact ids instead of keys
- updates filters and filter mappers to use new artifact filter utilities


pleek91 (approved):
There's some linting warnings but other than that looks good

PR #1264, opened 2023-03-27
Enhancement: Expose subscription in useFlowRun and similar compositions

pleek91:
# Description
This would be a breaking change and would require updating everywhere we currently do `const flowRun = useFlowRun(flowRunId)` to be `const { flowRun } = useFlowRun(flowRunId)`

The motivation for this change is currently there is no way to know if the api request actually succeed or failed. Exposing the underlying subscription allows the most flexibility at implementation.


znicholasbrown (approved):
very clean; let's do a minor release of this immediately to signal the breaking change

znicholasbrown (approved):

PR #1270, opened 2023-03-28
Feature: Artifact table renderer

znicholasbrown:
This PR:
- adds a new renderer for artifacts
- fixes a few issues with the artifact mocker
- adds a new mock generator for tabular data

### DIFF
--- a/src/components/ArtifactDataTable.vue
+++ b/src/components/ArtifactDataTable.vue
@@ -1,6 +1,13 @@
 <template>
   <p-layout-table class="artifact-data-table">
     <template #header-end>
+      <span class="artifact-data-table__results-count">
+        <template v-if="filteredCount !== dataCount">
+          <span class="artifact-data-table__count">{{ filteredCount }}</span> of
+        </template>
+        <span class="artifact-data-table__count">{{ dataCount }}</span>
+        {{ toPluralString(localization.info.item, dataCount) }}
+      </span>
       <SearchInput v-model="search" />
     </template>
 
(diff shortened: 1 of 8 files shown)


src/components/ArtifactDataTable.vue
```diff
@@ -0,0 +1,64 @@
+<template>
+  <p-layout-table class="artifact-data-table">
+    <template #header-end>
+      <span class="artifact-data-table__results-count">
+        <template v-if="filteredCount !== dataCount">
+          <span class="artifact-data-table__count">{{ filteredCount }}</span> of
+        </template>
+        <span class="artifact-data-table__count">{{ dataCount }}</span>
+        {{ toPluralString(localization.info.item, dataCount) }}
+      </span>
```
pleek91, line 10: I feel like this is different than our other tables. Don't we usually just say x items and lower the number as users search?

src/mocks/table.ts
```diff
@@ -0,0 +1,34 @@
...
+export type MockTableDataTypes = 'noun' | 'number' | 'boolean' | 'dateString' | 'email' | 'runName'
+
+export type MockTableOverrides = {
+  keys?: string[],
+  types?: MockTableDataTypes[],
+  columnCount?: number,
+  rowCount?: number,
+}
+
+export const randomTable: MockFunction<TableData[], [Partial<MockTableOverrides>?]> = function(overrides = {}) {
```
pleek91, line 14: This is cool. I can see this being expanded to a randomRecord and randomList that can create a list of objects that all have the same type but who's properties could have values from any of the other mocks

pleek91 (approved):
One question but otherwise LGTM

src/components/ArtifactDataTable.vue
znicholasbrown: Yeah I was trying to make a pretty distinct break from other tables since this is for user-supplied content; the expectations are necessarily different for how this will operate

pleek91: This feels like a pattern break that is going to make our other tables feel off compared to this one. My vote would be to decide if this is the pattern we want to follow for counting results and update the other places as well. I think this is a good pattern. But I wanna keep the user experience consistent.

PR #1272, opened 2023-03-29
Feature: Artifact composition, details component, and raw data view

znicholasbrown:
As the title states, this PR introduces 2 new components, an `ArtifactDetails` (in line with existing `EntityDetails` components) that shows entity information about the artifact and any flow or task run that generated it, and an `ArtifactDataRaw` component for safely rendering whatever is in the `data` column, with syntax highlighting when appropriate.

It also updates the artifact filters with the latest API changes and introduces an artifact composition


znicholasbrown:
Supersedes #1271

src/components/PageHeadingArtifact.vue
```diff
@@ -32,7 +31,7 @@
   const crumbs = computed<BreadCrumbs>(() => {
```
pleek91, line 31: I was trying to understand the logic here and rewrote it this way. Feel free to do nothing with this

```typescript
  const crumbs = computed<BreadCrumbs>(() => {
    const artifactIdCrumb: BreadCrumb = {
      text: props.artifact.id
    }

    if(props.artifact.key) {
      return [
        { 
          text: props.artifact.key, 
          to: routes.artifactKey(props.artifact.id) 
        },
        artifactIdCrumb
      ]
    }

    const internalCrumbs: BreadCrumbs = []

    if(flowRun.value) {
      internalCrumbs.push({
        text: flowRun.value.name ?? '',
        to: routes.flowRun(flowRun.value.id),
      })
    }

    if (taskRun.value) {
      internalCrumbs.push({
        text: taskRun.value.name ?? '',
        to: routes.taskRun(taskRun.value.id),
      })
    }

    internalCrumbs.push(artifactIdCrumb)

    return internalCrumbs
  })
```

src/types/artifact.ts
```diff
@@ -19,3 +19,7 @@ export function isMarkdownArtifactData(data: ArtifactData): data is MarkdownArti
 export function isTableArtifactData(data: ArtifactData): data is TableArtifactData {
   return Array.isArray(data)
 }
+
+export function isSerializedArtifactData(type: string, data: ArtifactData): data is string {
+  return type !== 'markdown' && type !== 'result' && typeof data === 'string'
```
pleek91, line 24: I'm not totally sure what values are in ArtifactData but is there a positive check we could make here? Rather than checking if its not markdown or result can we check that it IS another type? I feel like this is fragile if we add more types where the data is a string but its not serialized.

zhen0:
Run locally and looks good apart from table rendering. 

 In the docs we are showing this but it creates an empty table in the UI.
def my_fn():
    highest_churn_possibility = {
        'customer_id': ['123456789', '987654321', '246810121', '135791113', '864208046'],
        'name': ['John Smith', 'Jane Doe', 'Bob Johnson', 'Sarah Jones', 'Tom Wilson'],
        'churn_probability': [0.85, 0.79, 0.67, 0.61, 0.57]
    }
    create_table_artifact(
        key="personalized-reachout",
        table=highest_churn_possibility,
        description= "# Marvin, please reach out to these customers today!"
    )
If I change it to this is works nicely:
def my_fn():
    highest_churn_possibility = [
       {'customer_id':'12345', 'name': 'John Smith', 'churn_probability': 0.85 }, 
       {'customer_id':'56789', 'name': 'Jane Jones', 'churn_probability': 0.65 } 
    ]

    create_table_artifact(
        key="personalized-reachout",
        table=highest_churn_possibility,
        description= "# Marvin, please reach out to these customers today!"
    )

src/components/PageHeadingArtifact.vue
```diff
@@ -32,7 +31,7 @@
...
     const internalCrumbs: BreadCrumbs = []
 
-    if (hasRun.value) {
+    if (hasRun.value && !props.artifact.key) {
```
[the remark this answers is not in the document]
znicholasbrown: It's a string used to associate disparate artifacts; it allows tracking of data over time as new artifacts are created with the same key

src/compositions/useArtifact.ts
```diff
@@ -0,0 +1,28 @@
...
+  const api = useWorkspaceApi()
+  const id = ref(artifactId)
+
+  const parameters = computed<[string] | null>(() => {
+    if (!id.value) {
+      return null
+    }
+
+    return [id.value]
+  })
```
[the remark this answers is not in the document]
znicholasbrown: Not as far as I know, though that will likely change in the future

src/models/Artifact.ts
```diff
@@ -58,6 +62,7 @@ export class Artifact implements IArtifact {
   public type: ArtifactType
   public description: string | null
   public data: ArtifactData
+  public readonly serializedData: string | null | undefined
```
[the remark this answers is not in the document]
znicholasbrown: serializedData is basically whatever "raw" data was returned from the API. Different artifacts have different types returned for the `data` field that we need parse to do anything meaningful with. Rather than require the rest of the application to care about that nuance, I'm storing a reference to the un-parsed data in case it's needed (currently only the raw data view uses this)

src/maps/artifact.ts
```diff
@@ -1,17 +1,29 @@
...
 export const mapArtifactResponseToArtifact: MapFunction<ArtifactResponse, Artifact> = function(source) {
+  let { data } = source
+  let serializedData
+
+  if (isSerializedArtifactData(source.type, source.data)) {
+    data = parseUnknownJson(source.data)
+    serializedData = source.data
+  } else {
+    serializedData = stringifyUnknownJson(source.data)
+  }
```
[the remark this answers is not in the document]
znicholasbrown: Correct

src/types/artifact.ts
znicholasbrown: You know I think I'm going to drop this type guard and just check if the data is already a string. If it's not I'll serialize it myself so we can use it in code highlighters

src/components/ArtifactDataRaw.vue
```diff
@@ -0,0 +1,38 @@
+<template>
+  <section class="artifact-data-raw">
```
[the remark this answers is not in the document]
pleek91: I see, so raw here means we're going to show it in a code view rather than trying to output it as content.

znicholasbrown:
@pleek91 raised some good questions so I'm going to walk back some of the new model logic in favor of renderer-level checks; that means I'll tackle the other data type at the same time @zhen0. Will ping when done, thanks both of you!

znicholasbrown:
Ok both issues should be addressed.

@zhen0 renderer should be working for the map of lists now:

Screenshot 2023-03-29 at 5 56 26 PM
Screenshot 2023-03-29 at 5 56 46 PM

znicholasbrown:
Note: I also updated the workspace artifacts API to hit the non-experimental route since that was cut over already

zhen0:
image

 Yup can confirm that's rendering correctly now.  Thanks @znicholasbrown

zhen0 (approved):
One tiny lint import warning on src/components/ArtifactCollections.vue and it would be good, if time permits, to have @pleek91 take another look at the code but functionality LGTM.

src/components/ArtifactDataTable.vue
```diff
@@ -22,30 +28,62 @@
...
-  const data = computed<TableData[]>(() => props.artifact.data)
+  const data = computed(() => parseUnknownJson(props.artifact.data))
+  const isValid = computed(() => isTableArtifactData(props.artifact.data))
+
+  const normalizedData = computed(() => {
+    if (isArrayOfMaps(data.value)) {
+      return data.value
+    }
+
+    if (isMapOfArrays(data.value)) {
```
pleek91, line 46: Can give me a small example of this? Just to visualize

src/types/artifact.ts
```diff
@@ -16,6 +16,14 @@ export function isMarkdownArtifactData(data: ArtifactData): data is MarkdownArti
   return typeof data === 'string'
 }
 
-export function isTableArtifactData(data: ArtifactData): data is TableArtifactData {
-  return Array.isArray(data)
+export function isArrayOfMaps(data: unknown): data is Record<string, unknown>[] {
+  return Array.isArray(data) && data.every(row => typeof row === 'object')
```
pleek91, line 20: Do we need to exclude arrays and nulls here? In the every

src/utilities/json.ts
```diff
@@ -21,5 +21,11 @@ export function parseUnknownJson(value: unknown): unknown {
 export function stringifyUnknownJson(value: unknown): string | null | undefined {
   const parsed = parseUnknownJson(value)
 
+  return stringify(parsed)
+}
+
+export function formatUnknownJson(value: unknown): string {
```
pleek91, line 27: These two methods are the same except the return type. Seems like a red flag? They cannot both be correct right?

src/localization/locale/en.ts
```diff
@@ -93,6 +93,8 @@ export const en = {
     artifactSearch: 'Search artifacts',
     latest: 'Latest',
     item: 'Item',
+    noData: 'No data',
+    invalidData: (docsUrl: string) => `Invalid data, see [documentation](${docsUrl}) for more information`,
```
pleek91, line 97: Nit: 
```suggestion
    invalidArtifactData: (docsUrl: string) => `Invalid data, see [documentation](${docsUrl}) for more information`,
```


I missed where this is used. But wouldn't the docs url come from localization as well? I'd almost expect this to accept the type or something about the artifact and internally determine the docs url.

pleek91 (commented):
I like the changes away from doing the parsing in the mapper. Couple of questions but this looks pretty good to me

src/components/ArtifactDataTable.vue
znicholasbrown: Yup!

This:
```json
{
  "id": [0, 1, 2, 3, 4, 5, 6, 7],
  "name": ["Dublin", "London", "New York", "Oslo", "Paris", "Rome", "Stockholm", "Tokyo"],
  "lat": [53.3498, 51.5074, 40.7128, 59.9139, 48.8566, 41.9028, 59.3293, 35.6762],
  "lon": [-6.2603, -0.1278, -74.0060, 10.7522, 2.3522, 12.4964, 18.0686, 139.6503],
}
```

vs:
```json
[
    {
        "id": 0, "name": "Dublin", "lat": 53.3498, "lon": -6.2603,
    },
    {
        "id": 1, "name": "London", "lat": 51.5074, "lon": -0.1278,
    },
    {
        "id": 2, "name": "New York", "lat": 40.7128, "lon": -74.0060,
    },
    {
        "id": 3, "name": "Oslo", "lat": 59.9139, "lon": 10.7522,
    },
    {
        "id": 4, "name": "Paris", "lat": 48.8566, "lon": 2.3522,
    },
    {
        "id": 5, "name": "Rome", "lat": 41.9028, "lon": 12.4964,
    },
    {
        "id": 6, "name": "Tokyo", "lat": 35.6895, "lon": 139.6917,
    },
    {
        "id": 7, "name": "Vancouver", "lat": 49.2827, "lon": -123.1207,
    }
]
```

pleek91: Thank you!

src/localization/locale/en.ts
znicholasbrown: So it's only used in one place right now (empty state) but once docs are more solid I'll add other links so we can link directly to intended types if someone's data doesn't look quite right. I don't feel strongly about a ref vs a string, string just seemed clearer for the moment since I only have one ref

src/types/artifact.ts
znicholasbrown: Ah yeah we do - good catch

pleek91: I think there's an isRecord utility for this

znicholasbrown: nice, good looks 👍🏻

src/utilities/json.ts
znicholasbrown: oof, one day i'll explain my thought process here but it's embarrassing; removed in favor of the existing one

PR #1273, opened 2023-03-30
Enhancement: Artifacts timeline improvements

znicholasbrown:
This PR adjusts the artifact timeline component to use the new prefect-design timeline and improves on the overall UX


src/components/ArtifactTimelineItemContent.vue
```diff
@@ -0,0 +1,180 @@
...
+  import { Artifact } from '@/models'
+  import { isNullish } from '@/utilities'
+
+  type Expanded = boolean | unknown[] | undefined | null
+
+  const props = defineProps<{
+    artifact: Artifact,
+    // TODO: I don't think this logic makes sense here
+    // but I want to wait to create a more generic expanded card component
+    expanded?: Expanded,
```
pleek91, line 73: Mentioned this in other PR. But it doesn't seem like expanded needs to be a prop? There a lot of logic around this because its a v-model when I think this could be handled as a simple boolean ref internal to the component.

zhen0:
I think the upgrade in prefect-design has somehow removed colors from the flow run list item

image

I don't think that's deliberate?

src/components/ArtifactTimeline.vue
```diff
@@ -69,12 +101,23 @@
...
 
+  const getOffsetArtifacts = async (): Promise<void> => {
+    const result = await api.artifacts.getArtifacts(artifactsFilterWithOffset.value)
+    artifacts.value = [...new Map([...artifacts.value, ...result].map(obj => [obj.id, obj])).values()].sort((objA, objB) => sortDates(objB.created, objA.created))
+  }
+
   const getArtifacts = async (): Promise<void> => {
     const result = await api.artifacts.getArtifacts(artifactsFilter.value)
-    artifacts.value = [...artifacts.value, ...result]
+    artifacts.value = [...new Map([...artifacts.value, ...result].map(obj => [obj.id, obj])).values()].sort((objA, objB) => sortDates(objB.created, objA.created))
```
pleek91, line 111: 1. You're repeating the same logic when setting artifacts.value
2. This feels very code golfy. 
3. Why create a map just to spread it into an array? Are you trying to remove duplicates?

znicholasbrown: I can expand this a bit if it's helpful but yes the point is deduping since these are disparate calls - I'm not sure I understand your first point though

znicholasbrown: Moved the sorting and mapping pieces out a little to make this a bit more readable

src/components/ArtifactTimeline.vue
```diff
@@ -1,31 +1,53 @@
...
+        </template>
       </template>
-    </p-virtual-scroller>
+    </p-timeline>
   </div>
 </template>
 
 <script lang="ts" setup>
+  import { Icon } from '@prefecthq/prefect-design'
+  import { TimelineItem } from '@prefecthq/prefect-design/dist/types/src/types/timeline'
```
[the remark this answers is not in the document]
znicholasbrown: Agreed, can take a look later today

src/models/Artifact.ts
```diff
@@ -7,6 +9,14 @@ export const artifactTypes = [
...
 export type ArtifactType = typeof artifactTypes[number]
 
+export const artifactTypeIconMap: Record<ArtifactType | 'default', Icon> = {
+  default: 'FingerPrintIcon',
+  markdown: 'DocumentTextIcon',
+  table: 'TableIcon',
+  result: 'CalculatorIcon',
+  unknown: 'FingerPrintIcon',
+}
+
```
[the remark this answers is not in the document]
znicholasbrown: oh that's pretty cool - updated

zhen0:
> @zhen0 it looks like the fix for those missing colors is pretty easy - the selector coming from `prefect-design` changed slightly from `list-item-input__checkbox` to `p-list-item-input__checkbox` - I can include the fix in this PR if you'd like so we feel comfortable with the bump

Thank you!  Yeah if it's a small change let's get it in.  Thank you!

znicholasbrown:
Added [here](https://github.com/PrefectHQ/prefect-ui-library/pull/1273/files#diff-4f1f67393f65c8788e77ede1b67180febbeeb42a1c49d68c98240d76807b8ca8) (hopefully that diff works properly)

src/components/ArtifactTimeline.vue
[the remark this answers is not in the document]
znicholasbrown: Actually it was pretty simple, fix is [here](https://github.com/PrefectHQ/prefect-design/pull/712) (I don't think that should block this PR though)

src/models/Artifact.ts
```diff
@@ -7,6 +9,14 @@ export const artifactTypes = [
 
 export type ArtifactType = typeof artifactTypes[number]
 
+export const artifactTypeIconMap: Record<ArtifactType | 'default', Icon> = {
+  default: 'FingerPrintIcon',
+  markdown: 'DocumentTextIcon',
+  table: 'TableIcon',
+  result: 'CalculatorIcon',
+  unknown: 'FingerPrintIcon',
+} satisfies Record<ArtifactType | 'default', Icon>
```
pleek91, line 18: So if you have satisfies you should remove the type from the const itself. Having both negates the satisfies I believe. I'd also still add the `as const`. 

```suggestion
export const artifactTypeIconMap = {
  default: 'FingerPrintIcon',
  markdown: 'DocumentTextIcon',
  table: 'TableIcon',
  result: 'CalculatorIcon',
  unknown: 'FingerPrintIcon',
} as const satisfies Record<ArtifactType | 'default', Icon>
```

pleek91 (approved):
One nit on the satisfies thing (you don't really need to use it here but if you're going to it needs a slight tweak). But otherwise LGTM

src/models/Artifact.ts
znicholasbrown: Oh TIL - i'll update that, thanks for catching that

PR #1280, opened 2023-03-30
Chore: Variable models

znicholasbrown:
Nothing special here, just models and mappers for the new variables concept; API to follow


znicholasbrown:
Resolves #1282
Resolves: #1280

src/maps/index.ts
```diff
@@ -41,6 +41,7 @@ import { mapStateUpdateToStateUpdateRequest } from '@/maps/stateUpdate'
 import { mapTaskInputToTaskInputResponse, mapTaskInputResponseToTaskInput } from '@/maps/taskInput'
 import { mapTaskRunToTaskRunResponse, mapTaskRunResponseToTaskRun } from '@/maps/taskRun'
 import { mapUiFlowRunHistoryResponseToUiFlowRunHistory } from '@/maps/uiFlowRunHistory'
+import { mapVariableResponseToVariable, mapVariableEditToVariableEditRequest, mapVariableCreateToVariableCreateRequest } from '@/maps/variable'
```
znicholasbrown, line 44: This looks like I did a lot but I just sorted these and ran the linter cause this file was getting messy

src/models/VariableCreate.ts
```diff
@@ -0,0 +1,6 @@
+export type VariableCreate = {
+  name: string,
+} & Partial<{
```
pleek91, line 3: I personally find this hard to read when using `Partial<{...}>` when declaring types. Usually partial is used for modifying an existing type. Could we just use the `?` for these types?

pleek91 (approved):

src/models/VariableCreate.ts
znicholasbrown: Ah yeah I forgot you're not fond of that - I'll update haha

znicholasbrown: For the record I like it because at a glance you're able to separate the required model from the optional model though point taken about partial usage, you're absolutely right

pleek91: I'd rather just sort required at the top and optional at the bottom. I think it is more parseable to see a property and see it's optional rather than find what section it is in. But I won't die on this hill lol.

znicholasbrown: That's an interesting idea, I think we already have (had) a class sorting rule for that - I got the model for this slightly wrong anyway but I updated this and the `VariableEdit` model so we're g2g 👍🏻

znicholasbrown: ... except I didn't push those changes before I merged 🤦🏻  I'll add them to my next PR shortly, sorry

PR #1284, opened 2023-03-30
Feature: VariablesWorkspaceAPI service

znicholasbrown:
Implements the variables API as it (mostly) exists right now; note that the `delete` and `edit` services are subject to change since those aren't done yet but I don't think they will do so significantly. I didn't include the bulk get/filter route since that's also WIP and I don't want to implement the filter models until I've seen the API (we've agreed on one but that can change during implementation)


znicholasbrown:
Resolves: #1284

pleek91 (approved):

PR #1285, opened 2023-03-31
Feature: Variables page heading and create modal

znicholasbrown:
This PR introduces v1 `PageHeadingVariables` and `VariableCreateModal` components. It also implements the type suggestions from #1280, adds variable permissions, adds the get by name route to the `WorkspaceVariablesApi` and fixes some missing exports in the mapper bucket

Closes: #1286
Closes: #1287


znicholasbrown:
These commits are whack but the diff is correct, apologies for that it looks like github isn't picking up that the original base branch commits are already in main after the rebase

znicholasbrown:
Oh it's because I squash committed

src/components/PageHeadingVariables.vue
```diff
@@ -0,0 +1,19 @@
+<template>
+  <page-heading class="page-heading-variables" :crumbs="crumbs">
+    <template v-if="can.create.variable" #after-crumbs>
+      <p-button v-if="can.create.work_pool" inset size="xs" icon="PlusIcon" @click="open" />
```
pleek91, line 4: I'm guessing this permission check isn't supposed to be here

src/components/VariableCreateModal.vue
```diff
@@ -0,0 +1,82 @@
...
+  })
+  const nameRules = fieldRules(localization.info.name, isRequired, [isUnique, localization.error.variableAlreadyExists])
+  const valueRules = fieldRules(localization.info.value, isRequired)
+
+  const { handleSubmit, isSubmitting, errors } = useForm<Variable>()
+  const { value: name, meta: nameState } = useField<string>('name', nameRules)
+  const { value: value, meta: valueState } = useField<string>('value', valueRules)
+  const { value: tags, meta: tagsState } = useField<string[]>('tags')
+
+  const submit = handleSubmit(async (values) => {
```
pleek91, line 72: Can we not introduce more vee validate? I'd like to use the useValidation and useValidationObserver compositions

znicholasbrown: You know it's been a minute since I was creating forms and I forgot which version I was using - yeah definitely, I'll update that

src/components/PageHeadingVariables.vue
znicholasbrown: Oops yup - fixed

src/components/VariableCreateModal.vue
```diff
@@ -0,0 +1,102 @@
...
+    },
+  })
+
+  const api = useWorkspaceApi()
+
+  const isUnique = isValidIf(async (value) => {
+    if (isNull(value) || !isString(value)) {
+      return false
+    }
+    const variable = await api.variables.getVariableByName(value)
```
pleek91, line 65: Since validation is triggered every time the value changes you might want to debounce/cancel this. 

There's an example in automations where it denounced depending on the source. See the isValidTemplate method. 

https://github.com/PrefectHQ/nebula-ui/blob/main/src/components/AutomationActionSendNotification.vue

pleek91 (approved):
Approving but check the comment about async validation.

src/components/VariableCreateModal.vue
```diff
@@ -0,0 +1,102 @@
...
+  const { error: valueErrorMessage, state: valueState } = useValidation(value, localization.info.value, rules.value)
+
+  const submit = async (): Promise<void> => {
+    const valid = await validate()
+
+    if (valid) {
+      try {
+        const values: VariableCreate = {
+          name: name.value!,
+          value: value.value!,
```
znicholasbrown, line 90: Asserting that these exist seems like it shouldn't be necessary but if the form is valid they _do_ and it feels like additional checks would be inappropriate here

src/components/VariableCreateModal.vue
znicholasbrown: Great minds - I pushed a version of that after I caught that on self-review lol - let me double check it matches roughly what you've already done

znicholasbrown: Ah ok I don't think just a debounced ref makes sense, I'll write one that uses the validation factory methods

pleek91: A denounced ref is a good idea. It would be simpler than what I have. But I think that would have the possibility of passing validation before the denounces ref gets updated. 

But you could validate the denounced ref during normal validation and manually validate the actual ref during submission. Not sure if that would work correctly. 

The method i wrote debounces validation unless an observer is doing the validation then it fires right away.

Issue #1289, opened 2023-03-31
UI - Variable OverflowMenu


znicholasbrown:
For use in the variables table, should contain buttons to "Edit", "Delete", "Copy Value", "Copy Name", "Copy ID"

PR #1294, opened 2023-03-31
Feature: `VariableMenu` component

znicholasbrown:
Closes: #1289


src/components/VariableMenu.vue
```diff
@@ -0,0 +1,63 @@
...
+  const deleteVariable = async (id: string): Promise<void> => {
+    closeDeleteModal()
+
+    try {
+      await api.variables.deleteVariable(id)
+      showToast(localization.success.delete(localization.info.variable), 'success')
+    } catch (error) {
+      console.error(error)
+      showToast(localization.error.delete(localization.info.variable.toLowerCase()), 'error')
+      emit('delete', id)
```
zhen0, line 60: Is the emit meant to be in the catch block?

src/components/VariableMenu.vue
```diff
@@ -0,0 +1,63 @@
...
+  const { open: openEditModal } = useShowModal()
+
+  const api = useWorkspaceApi()
+
+  const deleteVariable = async (id: string): Promise<void> => {
+    closeDeleteModal()
+
+    try {
+      await api.variables.deleteVariable(id)
+      showToast(localization.success.delete(localization.info.variable), 'success')
```
zhen0, line 56: Should the success toast also be lower case as the error one is?

src/components/VariableMenu.vue
znicholasbrown: Nope, great catch!

src/components/VariableMenu.vue
znicholasbrown: Nah - this one places the passed string at the start of the sentence; the error one places it at the end; good looks though, maybe the other method should be making things lowercase by default? 🤔  Won't tackle that in this PR though

pleek91 (approved):

PR #1299, opened 2023-03-31
Enhancement: Variables components

znicholasbrown:
This PR contains miscellaneous fixes/enhancements for the newly introduced variables components and services for issues found during live testing


src/components/VariableCreateModal.vue
```diff
@@ -106,8 +112,8 @@
         const variable = await api.variables.createVariable(values)
 
         showToast(localization.success.createVariable, 'success')
-        internalValue.value = false
         emit('create', variable)
+        internalValue.value = false
```
znicholasbrown, line 116: This makes sure the component isn't unmounted before we emit the `create` event

src/components/VariablesTable.vue
```diff
@@ -3,24 +3,24 @@
...
-          <p-tags-input v-model="filter.variables.tags.name" :empty-message="localization.info.tags" class="variables-table__tags" />
+          <p-tags-input v-model="filter.variables.tags.name" :empty-message="localization.info.filterByTags" class="variables-table__tags-input" />
         </div>
       </template>
 
       <p-table :data="variables" :columns="columns">
         <template #selection-heading>
-          <p-checkbox v-model="model" @update:model-value="selectAllVariables" />
+          <p-checkbox v-if="variables.length" v-model="model" @update:model-value="selectAllVariables" />
+          <div v-else />
```
znicholasbrown, line 23: This new check on `variables.length` and the empty div ensure the checkbox isn't shown on an empty table and that the `selection` heading isn't shown when the checkbox is hidden (an empty slot will render the column label by default)

src/components/VariablesTable.vue
```diff
@@ -148,18 +166,22 @@
...
-  function refresh(): void {
+  function refreshSubscriptions(): void {
     variablesSubscription.refresh()
     variablesCountSubscription.refresh()
   }
 
+  defineExpose({
+    refreshSubscriptions,
+  })
+
```
znicholasbrown, line 177: I don't think we're doing this in a ton of places so I wanted to call it out - this allows us to refresh the subscription when an external event renders it necessary - in this case it allows us to refresh when a new variable is added from elsewhere

znicholasbrown: Also I made the method name more verbose to reflect its use externally - `refresh` didn't seem clear enough when calling it from elsewhere

src/compositions/filters.ts
```diff
@@ -515,8 +515,32 @@ export function useDeploymentsFilter(defaultValue: MaybeReactive<DeploymentsFilt
...
+    valueLike: toRef(defaultValueReactive, 'valueLike'),
+    tags: tags.filter,
+  })
+
+  return withFilterFunctions(filter)
+}
+
 export function useVariablesFilter(defaultValue: MaybeReactive<VariablesFilter> = {}): UseFilter<VariablesFilter> {
-  return useUnionFilter<VariablesFilter>(defaultValue, defaultVariableSort)
+  const defaultValueReactive = getDefaultValueWithDefaultSort(defaultValue, defaultVariableSort)
```
znicholasbrown, line 534: This prior implementation was incorrect - I didn't quite understand how these compositions were meant to be put together. Works now though

src/services/WorkspaceVariablesApi.ts
```diff
@@ -21,9 +21,13 @@ export class WorkspaceVariablesApi extends WorkspaceApi implements IWorkspaceVar
...
-  public async getVariableByName(variableName: string): Promise<Variable> {
-    const { data } = await this.get<VariableResponse>(`/name/${variableName}`)
-    return mapper.map('VariableResponse', data, 'Variable')
+  public async getVariableByName(variableName: string): Promise<Variable | null> {
+    const { data } = await this.get<VariableResponse | null>(`/name/${variableName}`)
+    if (data) {
+      return mapper.map('VariableResponse', data, 'Variable')
+    }
+
+    return null
```
znicholasbrown, line 30: Technically this will throw an error if the API can't find that variable but allowing this to return null allows us to create much more readable checks/error handling wherever this service is used

src/utilities/validation.ts
```diff
@@ -255,4 +255,28 @@ export const isHandle: ValidationMethodFactory = property => value => {
...
+  if (isNullish(value) || isEmptyString(value)) {
+    return true
+  }
+
+  if (typeof value === 'string' && value === value.toUpperCase()) {
+    return true
+  }
+
+  return `${property} must be uppercase`
+}
```
znicholasbrown, line 282: Technically these are out of scope since I ended up using the `isHandle` validator instead but they felt like nice to haves (`isHandle` is doing lowercase checks under the hood). I don't have an immediate use for them though so happy to remove from this PR if anyone feels strongly.

src/components/VariablesTable.vue
```diff
@@ -87,6 +91,15 @@
...
   const variableLikeDebounced = useDebouncedRef(variableLike, 1000)
   const offset = ref(0)
+  const page = computed({
+    get() {
+      return offset.value + 1
+    },
+    set(value: number) {
+      offset.value = value - 1
+    },
+  })
```
pleek91, line 101: this is confusing. Offset is used for the filter so I'd expect the offset to be `page * limit`. What's the reason for the get/set? The + 1 and - 1 is mysterious as well.

src/components/VariablesTable.vue
pleek91: This is something that the subscription pattern doesn't handle well. Don't think I have a better idea. Usually we just recreate the subscription externally but the filters here are so specific you wouldn't be able to.

src/utilities/validation.ts
pleek91, line 282: There's a new type for validation rules for the compositions. `ValidationRule` is the type. I'm actually presently surprised that we can use the existing factory methods with the new compositions. For some reason I had in my mind that they wouldn't work.

pleek91: Using the ValidationRule type enabled you to do this
```
useValidation(value, 'name', [isRequired, isLowerCase])
```

Rather than
```
useValidateion(value, [isRequired('name'), isLowerCase('name'])
```

znicholasbrown: Ah I can update these - I didn't realize I was using the old ones. These work perfectly though, but they're certainly verbose

src/components/VariablesTable.vue
znicholasbrown: Oh you're right, I forgot to take the limit into account; the reason is that the pager component doesn't work well with offsets and pages aren't something that exist API-side so we need a bridge to map the two.

pleek91: I think offset is computed. At least this is how I imagined it working when I built the pager

```
const offset = computed(() => page.value * limit)
```

src/utilities/validation.ts
pleek91: lol yeah they are really verbose. Another side effect of trying to fit things into vee-validate. But the compositions tried to make that better. You can also create type safe validation rules like `ValidationRule<string>`.

src/components/VariablesTable.vue
znicholasbrown: Agreed - honestly this doesn't feel that bad component-side; it's a little odd consumer-side cause you're creating a `ref`, but it's well-typed so it's easy to inspect

src/components/VariablesTable.vue
znicholasbrown: That makes sense - though we'll always need to subtract 1 from that offset if we're using human-readable page numbers (1 indexed instead of 0)

znicholasbrown: Updated to this method btw

pleek91: correct, my example didn't account for that.

src/utilities/validation.ts
znicholasbrown: I ended up removing these for now - can add them again later but felt this was getting too out of scope

pleek91 (approved):

Issue #1301, opened 2023-03-31
UI - Truncate variable values when they're in the table


znicholasbrown:
Values have a 5000 char limit so we'll want to show significantly less than that

PR #1303, opened 2023-04-03
Enhancement: Variables correctifs

znicholasbrown:
This PR:

- fixes an issue where create and edit modals had incorrect save/create/close ordering
- adds a check for value length (max 255 chars)
- ensures `update` is emitted across nested components (so any consumers can subscribe and refresh their data when a change has taken place)
- adds `key` to the `VariableMenu` instantiation in the table to work around an [issue in prefect-design](https://github.com/PrefectHQ/prefect-design/issues/719) (I don't want to tackle that issue as part of this since that could have broader impact)
- truncates `name` and `value` columns

table before:

Screenshot 2023-04-03 at 12 23 12 PM

table after:

Screenshot 2023-04-03 at 12 22 53 PM


znicholasbrown:
Closes: #1300 
Closes: #1301

src/components/VariableEditModal.vue
```diff
@@ -60,7 +57,15 @@
 
   const api = useWorkspaceApi()
 
-  const isUnique: ValidationRule<string | undefined> = async (value, label, { signal, source, previousValue }) => {
+  const valueIsLessThanOrEqual: ValidationRule<string | undefined> = (value) => {
+    if (isNull(value) || !isString(value)) {
+      return false
+    }
```
pleek91, line 63: So since you've defined your rule as `ValidationRule<string | undefined>` I'd change this to 
```
if(!isDefined(value)) {
  return false
}
```

src/components/VariableEditModal.vue
```diff
@@ -60,7 +57,15 @@
...
+
+    return value.length <= VALUE_MAX_CHARS ? true : localization.error.stringValueTooLong(localization.info.value, VALUE_MAX_CHARS)
```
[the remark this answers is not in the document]
pleek91: Also, You've defined this twice now. I'd extract this into a rule factory like
```typescript
export function isLessThanOrEqual(max: number): ValidationRule<string | number | undefined> {
  return (value, name) => {
    if(!isDefined(value)) {
      return true
    }

    const count = isString(value) ? value.length : value

    if(valueCount <= max) {
      return true
    }

    return localization.error.stringValueTooLong(name, max)
  }
}
```

src/components/VariableCreateModal.vue
```diff
@@ -114,6 +119,9 @@
         showToast(localization.success.createVariable, 'success')
         emit('create', variable)
         internalValue.value = false
+        name.value = undefined
+        value.value = undefined
+        tags.value = []
```
pleek91, line 124: this is resetting the form state right? I'd extract this into a separate `reset` method. This is something I wish prefect-design could handle for us.

pleek91 (approved):

PR #1304, opened 2023-04-03
add flow name to sub flows on flow run graph

brandonreid:
Add the flow name along side the flow-run name for sub-flow-runs.

image

Related to PrefectHQ/prefect#8969

### DIFF
--- a/src/components/FlowRunTimeline.vue
+++ b/src/components/FlowRunTimeline.vue
@@ -86,7 +86,7 @@
   import { UseSubscription, useSubscription } from '@prefecthq/vue-compositions'
   import { computed, onMounted, onUnmounted, ref, watch } from 'vue'
   import { FlowRunTimelineSelectionPanel, FlowRunTimelineOptions } from '@/components'
-  import { useFlowRuns, useWorkspaceApi } from '@/compositions'
+  import { useFlowRuns, useFlows, useWorkspaceApi } from '@/compositions'
   import { FlowRun, hasSubFlowRunId, isRunningStateType, isTerminalStateType, isValidGraphTimelineNode, TimelineNode } from '@/models'
   import { WorkspaceFlowRunsApi } from '@/services'
   import { prefectStateNames } from '@/types'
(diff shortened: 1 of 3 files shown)


src/components/FlowRunTimeline.vue
```diff
@@ -309,11 +309,45 @@
...
+
   const subFlowRunLabels = computed(() => {
     return (subFlowRunsSubscription.response ?? [])
       .reduce((acc, curr) => {
         if (curr.name) {
-          acc.set(curr.id, curr.name)
+          let subFlowRunName = ''
+
+          if (subFlowSubscription.response) {
+            const subFlow = subFlowSubscription.response.find((flow) => flow.id === curr.flowId)
```
pleek91, line 342: Do you have an idea of how many subflows might exist? If there are a lot we probably should convert the response to a map keyed by the id so we can do lookups like this without looping.

src/components/FlowRunTimelineSubFlowRunDetails.vue
```diff
@@ -43,18 +40,50 @@
...
-
   const api = useWorkspaceApi()
+
   const flowRunSubscriptionArgs = computed<[string]>(() => [flowRunId.value])
   const flowRunSubscription = useSubscriptionWithDependencies(
     api.flowRuns.getFlowRun,
     flowRunSubscriptionArgs,
   )
 
   const subFlowRun = computed(() => flowRunSubscription.response)
```
pleek91, line 51: This subscription is more complex than it needs to be. You're just getting a single flow run right? 
```typescript
const { flowRunId } = toRefs(props)
const { flowRun: subFlowRun } = useFlowRun(flowRunId)
```

src/components/FlowRunTimelineSubFlowRunDetails.vue
```diff
@@ -43,18 +40,50 @@
...
+
+  const flowSubscriptionArgs = computed<[string]>(() => [flowRunSubscription.response?.flowId || ''])
+  const flowSubscription = useSubscriptionWithDependencies(
+    api.flows.getFlow,
+    flowSubscriptionArgs,
+  )
+
+  const flow = computed(() => flowSubscription.response)
```
pleek91, line 59: Same here
```
const flowId = computed(() => subFlowRun.value?.flowId)
const { flow } = useFlow(flowId)
```

src/components/FlowRunTimeline.vue
```diff
@@ -309,11 +309,45 @@
...
+          id: subFlowIds,
+        },
+      },
+    ]
+  })
+
+  const subFlowSubscription = useSubscriptionWithDependencies(
+    api.flows.getFlows,
+    subFlowsSubscriptionFilter,
+  )
```
pleek91, line 333: We should make a `useFlows` composition for this.

src/components/FlowRunTimelineSubFlowRunDetails.vue
brandonreid: Maybe we can look at this together when we pair. When another node is selected, the component doesn't remount, it get's a new ID. So here it'll actually update the data, but there's probably another way.

pleek91: The id is reactive. So if flowId changes it'll fetch the new flow.

src/components/FlowRunTimelineSubFlowRunDetails.vue
brandonreid: ahhh, didn't know this existed

brandonreid: Updated the task run details panel to use the composition as well.

src/components/FlowRunTimeline.vue
brandonreid: You [resolved](https://github.com/PrefectHQ/prefect-ui-library/pull/1307) this one for me!

src/components/FlowRunTimeline.vue
```diff
@@ -292,6 +292,7 @@
     })
 
     expandedSubFlowRuns.value.set(id, {
+      // @ts-expect-error data actually is a valid property
```
pleek91, line 296: Did this error not go away when you were not linked? There shouldn't be a typescript error.

brandonreid: I forgot to `npm ci` after resetting `package.json`. Yeah it's not complaining now.

pleek91 (approved):

PR #1306, opened 2023-04-03
Chore: Move max chars to the variable model, separate value and name limits

znicholasbrown:
Related to https://github.com/PrefectHQ/nebula/pull/4295

Closes: #1301

### DIFF
--- a/src/models/Variable.ts
+++ b/src/models/Variable.ts
@@ -8,6 +8,9 @@ export interface IVariable {
   tags: string[],
 }
 
+export const MAX_VARIABLE_NAME_LENGTH = 255 as const
+export const MAX_VARIABLE_VALUE_LENGTH = 5000 as const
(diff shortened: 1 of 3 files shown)


src/models/Variable.ts
```diff
@@ -8,6 +8,9 @@ export interface IVariable {
   tags: string[],
 }
 
+export const MAX_VARIABLE_NAME_LENGTH = 255 as const
+export const MAX_VARIABLE_VALUE_LENGTH = 5000 as const
```
znicholasbrown, line 12: I'm not 100% sold on these living here but I'm not sure of a better place and this seemed central enough.

pleek91: I like this. Much better than component level

pleek91 (approved):

PR #1307, opened 2023-04-03
Enhancement: Flow run timeline interval and subscriptions

pleek91:
# Description
- Fixes timeline using a subscription interval when the flow is not running
- Adds useFlows and useFlowRuns compositions for better devx


src/compositions/useFlows.ts
brandonreid, line 1: I was just starting to do this, haha, ty!

src/components/FlowRunTimeline.vue
```diff
@@ -215,11 +215,12 @@
   })
 
   const api = useWorkspaceApi()
+  const interval = isRunningStateType(props.flowRun.stateType) ? { interval: 5000 } : undefined
```
znicholasbrown, line 218: This will maintain a 5 second interval for any runs that transition from non-terminal states during the lifetime of the component, is that ok?

brandonreid (approved):

src/components/FlowRunTimeline.vue
pleek91: @znicholasbrown take a look at the state type watcher. My thought is if the flowRun enters a terminal state
- Refresh immediately (get the final data)
- Unsubscribe (data cannot change anymore)
That will stop the interval

brandonreid:
Works in the UI

src/components/FlowRunTimeline.vue
znicholasbrown: ty!!

znicholasbrown (approved):

PR #1313, opened 2023-04-04
Feature: Array `toggle` utility

pleek91:
# Description
Utility for easily toggling a value in an array. Works similar to `classList.toggle`

### DIFF
--- a/src/utilities/arrays.ts
+++ b/src/utilities/arrays.ts
@@ -168,3 +168,14 @@ export function separate<T>(source: T[], filter: (value: T) => boolean): [found:
     return [found, notFound]
   }, [[], []])
 }
+
+export function toggle<T>(source: T[], value: T, force?: boolean): T[] {
+  const index = source.indexOf(value)
+  const add = force === true || index === -1
+
+  if (add) {
+    return [...source, value]
+  }
+
+  return source.filter(item => item !== value)
+}


src/utilities/arrays.ts
```diff
@@ -168,3 +168,14 @@ export function separate<T>(source: T[], filter: (value: T) => boolean): [found:
     return [found, notFound]
   }, [[], []])
 }
+
+export function toggle<T>(source: T[], value: T, force?: boolean): T[] {
+  const index = source.indexOf(value)
+  const add = force === true || index === -1
```
znicholasbrown, line 174: It seems like any scenario where you might want to pass `force` to this method would be as easily and more clearly written with `source = [...source, value]` instead of `source = toggle(source, value, true)`

znicholasbrown: I might be misunderstanding the use case though

pleek91: Agreed if you knew you were adding it. Its not uncommon for some dynamic variable to determine the the functionality. So having the logic for adding/removing handled for you and having the ability to just do `const values = toggle(source, 'value-to-toggle', dynamicBoolean)` is quite useful. Like half the time I'm using `classList.toggle` I'm using the `force` argument

znicholasbrown (approved):

src/utilities/arrays.ts
znicholasbrown: Got it - thanks for the explanation!

PR #1315, opened 2023-04-05
bump graphs and remove timeline data transforms

brandonreid:
The flow run graph has been updated to no longer require data be sorted and stripped of nodes with no start date. This PR removes those transforms from the data provided.


src/components/FlowRunTimeline.vue
```diff
@@ -287,33 +285,27 @@
...
-        return nodeA.start.getTime() - nodeB.start.getTime()
-      })
+  const getSubFlowRunIds = (data: GraphTimelineNode[]): string[] => {
+    return data
+      .map((node) => node.subFlowRunId)
+      .filter((subFlowRunId): subFlowRunId is string => subFlowRunId !== undefined)
   }
-
   const rootSubFlowRunIds = computed<string[]>(() => {
-    return graphData.value
```
brandonreid, line 308: For some reason typescript wasn't convinced this had not been returning possibly `undefined`. Went ahead and changed the check so it typecasts `string[]`

pleek91 (approved):

PR #1316, opened 2023-04-05
Feature: `FlowRunArtifacts` component

znicholasbrown:
Pretty straightforward component that takes a flow run and renders a list of all the artifacts associated with it, split by flow run and task run

Related: https://github.com/PrefectHQ/nebula-ui/issues/2623
Related: https://github.com/PrefectHQ/prefect/issues/9087


znicholasbrown:
Small demo: 

https://user-images.githubusercontent.com/27291717/230131414-9198e243-8737-42c6-888b-da5580bdc076.mov

brandonreid:
I recall some cards that were clickable to view expanded details within the same route? At some point we may want to consider what expectations we are setting for card clicks generally. Maybe a sort of `external link` icon or something would be helpful for these, especially since the resulting crumbs don't trail back to the flow run page.

znicholasbrown:
Agreed @brandonreid - I've elected to keep the behavior consistent with the artifacts collection components but a revisit wouldn't be bad

src/components/FlowRunArtifacts.vue
```diff
@@ -0,0 +1,131 @@
...
+
+  const artifactsFilter = computed<ArtifactsFilter>(() => {
+    return {
+      artifacts: {
+        flowRunId: [props.flowRun.id],
+        notType: ['result'],
+      },
+    }
+  })
+  const artifactsSubscription = useSubscription(api.artifacts.getArtifacts, [artifactsFilter], { interval: 5000 })
```
brandonreid, line 87: Is it worth using `isTerminalStateType` to conditionally poll here?

brandonreid (approved):

src/components/FlowRunArtifacts.vue
pleek91: Or only poll when `isRunningStateType` which is what we did for the timeline.

pleek91 (approved):

src/components/FlowRunArtifacts.vue
znicholasbrown: Yeah it definitely is - [ticket here](https://github.com/PrefectHQ/prefect-ui-library/issues/1320) so I can handle that in a follow up (will want to do that for results as well)

PR #1321, opened 2023-04-05
Chore: Artifact collections

znicholasbrown:
This PR adds more explicit support for artifact collections (containing objects for the `latest` artifact of a given key) and drops support for `is_latest` in the artifacts filter route. Updates all relevant components to use the new route

### DIFF
--- a/src/components/ArtifactTimeline.vue
+++ b/src/components/ArtifactTimeline.vue
@@ -74,18 +74,10 @@
     }
   })
 
-  const artifactsLatestFilter = computed<ArtifactsFilter>(() => {
-    return {
-      artifacts: {
-        key: [props.artifactKey],
-        isLatest: true,
-      },
-    }
-  })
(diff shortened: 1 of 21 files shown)


znicholasbrown:
Ah I just realized this'll break the mock api - let me update that quick

src/components/ArtifactTimeline.vue
```diff
@@ -119,7 +112,8 @@
       expanded.value = [val]
     }
     getArtifacts()
-  })
+  }, { immediate: true })
```
pleek91, line 115: You're calling getArtifacts in the onBeforeMount and then again here in this watcher. Adding immediate means it's going to get artifacts twice. Is that intentional?

pleek91 (approved):
Everything looks like I'd expect. One thing about the watcher that doesn't seem quite right but approving in case that's intentional or you want to fix in a future PR

src/components/ArtifactTimeline.vue
znicholasbrown: Ah actually that's an oversight; we should only refetch artifacts if this _isn't_ the first iteration. Will update, nice catch

PR #1322, opened 2023-04-05
Enhancement: Artifact card default slot

znicholasbrown:
This PR moves the default slot from the `ArtifactCard` component to instead wrap the existing summary item and adds additional slot for taking advantage of the summary row formatting. The default slot as previously implemented was a vestige of an earlier design and isn't used anywhere.


znicholasbrown:
Closes: #1323

brandonreid (approved):

PR #1324, opened 2023-04-05
Fixes display of block documents in work pool details

### LINKED ISSUE #1302: Display links for block documents in `WorkPoolDetails` component
When a block document is used as a default for a field in a base job template, the JSON representation of the block document  is displayed in the `WorkPoolDetails` component.

Image

This is issue is to update the `WorkPoolDetails` component to display a nice looking link like in the block document display components.

desertaxle:
Adds the ability to display block document values that come in the shape:
```json
{
    "$ref": {
        "block_document_id": "block_document_id"
     }
}
```

Blocks used in work pool configuration will now display as a link to the block matching the behavior in the `BlockDocumentCard` component.

Closes #1302

### DIFF
--- a/src/components/WorkPoolDetails.vue
+++ b/src/components/WorkPoolDetails.vue
@@ -24,12 +24,11 @@
 
 <script lang="ts" setup>
   import { useSubscription } from '@prefecthq/vue-compositions'
-  import { cloneDeep } from 'lodash'
   import { computed } from 'vue'
   import { SchemaPropertiesKeyValues } from '@/components'
   import { useCan, useWorkspaceApi } from '@/compositions'
   import { WorkPool } from '@/models'
-  import { schemaValuesBlockReferencesWithSchemaResolver } from '@/services'
+  import { mapper } from '@/services'
   import { getSchemaDefaults } from '@/utilities'
   import { formatDateTimeNumeric } from '@/utilities/dates'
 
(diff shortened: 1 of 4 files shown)


pleek91 (commented):
@desertaxle doesn't seem like this change should be needed if the schema is being mapped correctly. If the schema is mapped it should convert those `$ref`'s to match the existing `BlockDocumentValue` type.

pleek91:
Yeah looking at the `WorkPool` class the `baseJobTemplate` is `WorkPoolRequest`. And the `variables` property has the type `SchemasResponse`. Anything `SchemaResponse` shouldn't make it past the api service layer. If it was `Schema` then the components should just work? 

There's probable context I'm missing here. So please let me know if I'm misunderstanding something.

desertaxle:
After a discussion with @pleek91, we decided that it would be better to map the values given in the `WorkPoolDetails` component to match the format expected by the schema components. I've added a resolver to resolve block document references to block document values using a provided schema.

src/components/WorkPoolDetails.vue
```diff
@@ -52,7 +54,10 @@
   })
   const schema = computed(() => props.workPool.baseJobTemplate.variables ?? {})
   const showBaseJobTemplateDetails = computed(() => props.workPool.type && schemaHasProperties.value && can.access.workers)
-  const baseJobTemplateVariablesDefaults = getSchemaDefaults(props.workPool.baseJobTemplate.variables ?? {})
+  const baseJobTemplateVariablesDefaults = computed(() => {
+    const defaults = getSchemaDefaults(props.workPool.baseJobTemplate.variables ?? {})
+    return schemaValuesBlockReferencesWithSchemaResolver(defaults, cloneDeep(schema.value))
```
desertaxle, line 59: I don't want to have to use this `cloneDeep`, but I get an infinite update loop if I don't use it, so `schema` must be getting updated in this function call. Is there a better way to do this?

Issue #1314, opened 2023-04-04
UI - Update name form validation


znicholasbrown:
Name should no longer allow dashes (underscores are still fine)

PR #1325, opened 2023-04-05
Chore: variable names should be snake_cased

znicholasbrown:
Closes: #1314


src/utilities/validation.ts
```diff
@@ -268,4 +269,10 @@ export const isHandle: ValidationMethodFactory = property => value => {
   }
 
   return `${property} must only contain lowercase letters, numbers, and dashes`
-}
\ No newline at end of file
+}
+
+const SNAKE_CASE_REGEX = /^[a-z0-9]+(?:_[a-z0-9]+)*$/
+
+export const isSnakeCase = (message?: string): ValidationRule<unknown> => (value) => {
```
pleek91, line 276: So part of the reason for validation methods is they can return a string error message. Do we really need to to be customizable? I'd expect this to return the proper error message.

src/localization/locale/en.ts
```diff
@@ -101,6 +101,7 @@ export const en = {
     artifacts: 'Artifacts',
     artifactSearch: 'Search artifacts',
     variablesSearch: 'Search variables',
+    variableMustBeSnakeCase: 'Variable names may only contain letters, numbers, and underscores and may not begin or end with an underscore',
```
pleek91, line 104: If this was a generic snake case message then the validation rule could pretend the name field.

src/utilities/validation.ts
pleek91: The second argument that's passed from the validation is the name of the field.

znicholasbrown:
Fixed that @pleek91 👍🏻

pleek91 (approved):

PR #1331, opened 2023-04-06
Enhancement: Account for block document references with default values when resolving schemas

pleek91:
# Description
Adds a new resolver for schemas that resolved default values properties that have block document references. As well as accounts for properties that are type `block` that might have a default value. 

Supersedes: https://github.com/PrefectHQ/prefect-ui-library/pull/1324

### DIFF
--- a/src/models/WorkPool.ts
+++ b/src/models/WorkPool.ts
@@ -11,7 +11,7 @@ export interface IWorkPool {
   isPaused: boolean,
   defaultQueueId: string,
   concurrencyLimit: number | null,
-  baseJobTemplate: Record<string, unknown>,
+  baseJobTemplate: BaseJobTemplateRequest,
 }
 
 export class WorkPool implements IWorkPool {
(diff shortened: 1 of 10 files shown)


src/models/WorkPool.ts
```diff
@@ -11,7 +11,7 @@ export interface IWorkPool {
   isPaused: boolean,
   defaultQueueId: string,
   concurrencyLimit: number | null,
-  baseJobTemplate: Record<string, unknown>,
+  baseJobTemplate: BaseJobTemplateRequest,
```
znicholasbrown, line 14: I'm surprised to see a request model here, shouldn't this be a mapped response model?

znicholasbrown (approved):
Just one comment, otherwise LGTM

src/models/WorkPool.ts
pleek91: Agreed, I saw this and noticed the interface didn't match the implementation. The implementation uses the request model. I went a few minutes down the road of mapping it to a new `BaseJobTemplate` model but realized I'd have to touch quite a few files to update that and ultimately decided to leave it as is for now.

znicholasbrown: Ok since this is how the work pool model is typed already (I'm surprised that union works tbh) I see this as at least a strict improvement, if not a perfect one

pleek91:
@desertaxle I'd like to get your approval on this as well. Just to verify this fixes your issue.

desertaxle (approved):
Can confirm that this fixes the issue!

PR #1332, opened 2023-04-07
BugFix: Schema form incorrectly marking reference objects as optional

pleek91:
# Description
Fixes two issues with schema resolving
- Object properties were only considered required if they had required child properties. This is incorrect and required should always be determined by the parent property
- `allOf` and `anyOf` references should inherit the required value of the property they are associated with. Was previously hard coded to `false`

Need to do some more regression testing on this. 

Closes https://github.com/PrefectHQ/prefect/issues/9055

### DIFF
--- a/src/components/SchemaFormProperties.vue
+++ b/src/components/SchemaFormProperties.vue
@@ -18,7 +18,7 @@
     property: SchemaProperty,
   }>()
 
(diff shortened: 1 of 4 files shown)


pleek91:
Change looks good in staging

PR #1335, opened 2023-04-10
Enhancement: Poll based on flow run and task run state names

pleek91:
# Description
Updates the the interval for fetching flow and task run related assets to be based on the state so polling only happens when the flow run or task run is "Pending" or "Running"


zhen0 (commented):
It looks like the isRunningStateType check only includes the running state. We might want to add a shouldPollStateType (or something else with a better name) that also includes pending.

src/components/FlowRunLogs.vue
```diff
@@ -61,7 +62,10 @@
   }))
 
   const api = useWorkspaceApi()
-  const logsSubscription = usePaginatedSubscription(api.logs.getLogs, [logsFilter], { interval: 5000 })
+  const options = {
+    interval: isRunningStateType(props.flowRun.stateType) ? 5000 : Infinity,
```
brandonreid, line 66: I'd expect `undefined` rather than `Infinity` to be better in that it would never kick off a timer that goes indefinitely. Didn't look at the `interval` implementation though.

brandonreid (approved):

src/components/FlowRunLogs.vue
pleek91: `Infinity` is the default value for `interval`. But I think `undefined` is probably easier to understand here. Will update 👍

pleek91:
@zhen0 good idea. We're using `isRunningStateType` in a few places. But including `pending` would be a nice touch. Will update 👍

src/compositions/useStatePolling.ts
```diff
@@ -0,0 +1,33 @@
...
+
+type StateResource = {
+  stateName: string | null,
+}
+
+const stateNamesThatShouldPoll: (string | null)[] = ['Pending', 'Running'] satisfies PrefectStateNames[]
+
+export function useStatePolling(stateName: MaybeRef<string | null>, interval?: number): ComputedRef<SubscriptionOptions>
+export function useStatePolling(resource: MaybeRef<StateResource>, interval?: number): ComputedRef<SubscriptionOptions>
+export function useStatePolling(resourceOrStateName: MaybeRef<StateResource | string | null>, interval: number = 5000): ComputedRef<SubscriptionOptions> {
```
znicholasbrown, line 13: Just so I understand, the `StateResource` type is so we can directly pass `flowRun` and `taskRun` instances to the composition, right?

src/components/FlowRunResults.vue
```diff
@@ -53,11 +53,12 @@
...
   import { useSubscription } from '@prefecthq/vue-compositions'
-  import { computed } from 'vue'
+  import { computed, toRefs } from 'vue'
   import ArtifactResultCard from '@/components/ArtifactResultCard.vue'
   import RowGridLayoutList from '@/components/RowGridLayoutList.vue'
   import ViewModeButtonGroup from '@/components/ViewModeButtonGroup.vue'
   import { useWorkspaceApi } from '@/compositions'
+  import { useStatePolling } from '@/compositions/useStatePolling'
   import { localization } from '@/localization'
   import { FlowRun, ResultArtifact } from '@/models'
```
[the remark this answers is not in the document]
znicholasbrown: This is used in the template and eslint doesn't allow ignoring the import, unfortunately. I think we'll be getting some better utilities around this in the next couple of months (working through the vue-tsc release cycle)

znicholasbrown (approved):
I think we could make this composition a lot more flexible/clear by allowing the caller to pass a list of state names, so we could do something like import the list of terminal or non-terminal states and pass that to the composition with the state. My question when reading components where this was used was "use which states to poll?" 

That said, I won't hold up the PR for that as it feels like a nice to have

pleek91:
@znicholasbrown the intent is that we should be polling consistently based on globally set rules for which states should poll. That way at implementation we're not making that decision case by case. Helps keep things consistent as well as reduces mental load.

src/compositions/useStatePolling.ts
pleek91: Correct. That way we don't have to make a computed just to grab the state name from the flow run

znicholasbrown:
Hm ok I buy it, in which case the name is a little off - what we're saying is "poll on state" but what we should be saying is "stop polling if you're passed a terminal state" or alternatively "only poll in a non-terminal, non-scheduled state". Maybe `useStatePolling` => `useNonTerminalStatePolling`? Not sure I like that much but it's at least clearer

PR #1344, opened 2023-04-14
forward FlowRunTimeline visibleDateRange

brandonreid:


znicholasbrown (approved):

src/components/FlowRunTimeline.vue
```diff
@@ -52,6 +53,7 @@
           :sub-node-labels="subFlowRunLabels"
           :selected-node-id="selectedNode?.id"
           :expanded-sub-nodes="expandedSubFlowRuns"
+          @update:visible-date-range="updateVisibleDateRange"
```
pleek91, line 56: Rather than using the update event to emit the pattern we prefer to use is a get/set like this. As is you're mutating the prop and emitting the event. At the very least you need to change `v-model:visibleDateRange="..."` to `:visibleDateRange="..."`

```typescript
const visibleDateRange = computed({
  get() {
    return props.visibleDateRange
  },
  set(value) {
    emit('update:visibleDateRange', value)
  }
})
```

pleek91 (changes requested):

pleek91 (approved):
Exactly 👍

src/components/FlowRunTimeline.vue
brandonreid: Ah gotchya, ty!

PR #1346, opened 2023-04-18
Feature: `FlowList` and `DeploymentList` components

znicholasbrown:
This PR:

New:
- `FlowList` component; functionally equivalent to the `FlowsTable`
- `FlowListItem` component; functionally equivalent to a row in the `FlowTable` except contains a list of its deployments, used in `FlowList`
- `DeploymentList` component; functionally equivalent to the `DeploymentsTable`, used in `FlowListItem`
- `DeploymentListItem` component, functionally equivalent to a row in the `DeploymentsTable`, used in `FlowListItemDeployments`
- `lastFlowRun` composition; creates a subscription for the most recent flow run (by expected start time) matching the passed filter and returns a `ComputedRef` for the run and the subscription itself
- `nextFlowRun` composition; creates a subscription for the next flow run (by expected start time) matching the passed filter and returns a `ComputedRef` for the run and the subscription itself

Enhanced:
- deployment, flow, and flow run mock APIs all support basic filtering and sorting
- `StateListItem` component now supports an `action` slot, along with several other enhancements (non-breaking, should all be purely additive)

Fixed:
- mock data store utilities are now properly hydrating date objects for flows, deployments, and flow runs and schedule objects from deployments


znicholasbrown:
I haven't had a chance to self-review this yet so there might be a few rough edges - will do so tmr morning.

demo/compositions/useFlowMock.ts
```diff
@@ -1,29 +1,62 @@
...
-import { repeat } from '@/utilities'
+import { choice, repeat } from '@/utilities'
+
+export type FlowMockOptions = {
+  deploymentsCount: number,
+}
+
+export function useFlowMock(override?: Partial<Flow>, options?: Partial<FlowMockOptions>): Flow {
+  const {
+    deploymentsCount = mocker.create('number', [0, 8]),
```
znicholasbrown, line 12: These changes give us a little more control of how downstream objects are created when we're mocking flows

src/components/FlowsDeleteButton.vue
```diff
@@ -1,7 +1,5 @@
 <template>
-  <Transition name="flows-delete-button-transition">
-    <p-button v-if="selected.length > 0" danger icon="TrashIcon" @click="open" />
-  </Transition>
+  <p-button v-if="selected.length > 0" v-bind="attrs" danger icon="TrashIcon" @click="open" />
```
znicholasbrown, line 2: This transition was a bit of an anomaly compared to similar functioning buttons - not against transitions overall but it felt superfluous

src/components/PageHeadingFlows.vue
```diff
@@ -1,19 +1,8 @@
 <template>
-  <PageHeading class="page-heading-flows" :crumbs="crumbs">
-    <template v-if="can.access.collections" #after-crumbs>
-      <p-button :to="routes.flowCollections()" secondary>
-        Add from Collection
-      </p-button>
-    </template>
-  </PageHeading>
```
znicholasbrown, line 8: This link isn't being curated atm so we've decided to remove navigation to it for the time being; easy enough to add later should things change with that project.

src/components/ResultsCount.vue
```diff
@@ -1,6 +1,6 @@
 <template>
   <span v-if="isNumber(count)" class="results-count">
-    {{ count }} {{ toPluralString(label, count) }}
+    {{ count.toLocaleString() }} {{ toPluralString(label, count) }}
```
znicholasbrown, line 3: This fixes a widespread issue where results counts > 999 wouldn't display nicely

src/components/StateListItem.vue
```diff
@@ -56,6 +67,10 @@
 </script>
 
 <style>
+.state-list-item .p-list-item-input__checkbox { @apply
+  bg-transparent
```
znicholasbrown, line 71: As far as I've tested this is the only real change in function/behavior for this component and overall should be a strict improvement; previously if a state was passed that wasn't one we could map (as could be the case if states are deleted or new ones are introduced ahead of the UI mapping utilities), the styling for this component would fall back to the base prefect-design version, which has a background not dissimilar to "cancelled" and "crashed" states in most CVD palettes. This causes the stateful background to instead show as transparent, allowing the component to feel "stateless" instead

src/components/StateListItem.vue
```diff
@@ -93,15 +108,30 @@
...
-  gap-2
-  md:flex-row
-  md:gap-4
+  gap-2;
+
+  grid-template-columns: 1fr auto;
+  grid-template-areas:
+    "name action"
+    "tags tags";
+}
```
znicholasbrown, line 127: This looks like a more meaningful change than it is - it should prevent a couple blow out issues and allow us to scale the top section for small screens keeping the action slot in the same place but overflowing tags to the next line

src/components/index.ts
```diff
@@ -1,12 +1,12 @@
 export { default as ActivityChart } from './ActivityChart.vue'
```
znicholasbrown, line 1: Changes here are for new components and I ran a sort on the list which changed the order of a few items

brandonreid:
Just testing in the UI a bit.

It looks like when expanding Deployments, the FlowRunListItem shadow and bottom border radiuses change somewhat unexpectedly:

https://user-images.githubusercontent.com/6776415/232790818-4c60e369-cfee-41ff-ade0-66da3245f550.mov

I kinda wonder if icon usage is worth here? Sort of helps distinguish things as this pattern is being used more, but maybe in a list it would look too noisy. I'm sure you considered it:
image

demo/sections/flows/FlowList.vue
```diff
@@ -0,0 +1,15 @@
+<template>
+  <ComponentPage title="FlowListItem">
```
zhen0, line 2:
```suggestion
  <ComponentPage title="FlowList">
```

znicholasbrown:
@brandonreid fixed the border radius issue; I'm not against an icon per-se but I also don't want to shoe-horn them into this design when we're not using them for top-level items anywhere else.

znicholasbrown:
Also this PR is missing commits atm cause of the ongoing github issues - some of the fixes won't yet be applied

znicholasbrown:
Ok looks like all commits are showing up and should be in the demo now

src/components/FlowsDeleteButton.vue
pleek91: Totally agree

src/components/ListItemMeta.vue
pleek91, line 1: Love the idea of abstracting this pattern. Lead me to think we could probably build this in directly to `p-key-value`.

src/components/PageHeadingFlows.vue
pleek91: In favor of this. It was way to prominent and didn't seem to provide much value. But should we add a menu and add it there? Just so we retain the functionality.

src/components/ResultsCount.vue
pleek91: nice, good catch

src/compositions/useLastFlowRun.ts
pleek91, line 1: Love this

src/models/StateType.ts
```diff
@@ -16,6 +16,8 @@ export function isStateType(value: unknown): value is StateType {
   return typeof value === 'string' && stateType.includes(value as StateType)
 }
 
+export const scheduledStateType = ['scheduled', 'pending']
```
pleek91, line 19: should there be a `isScheduledStateType` utility to go with this?

src/components/FlowListItem.vue
```diff
@@ -0,0 +1,198 @@
...
+  const emit = defineEmits<{
+    (event: 'delete', value: string): void,
+  }>()
+
+  const { FlowMenu } = useComponent()
+  const { attrs, classes, styles } = useAttrsStylesAndClasses()
+  const api = useWorkspaceApi()
+  const routes = useWorkspaceRoutes()
+
+  const { value: expanded } = useLocalStorage(`flow-list-item-${props.flow.id}--expanded`, false)
```
pleek91, line 95: I think this should be session storage if anything at all. Reason being local storage and session storage have a limited amount of space. I've worked on apps in the past that saved a lot of state in local storage and ended up in a situation where some power users would get storage errors because they ran out of space. Session storage at least ensures this gets cleared out automatically when closing the app.

pleek91: Generally I try and only use local storage for things that are intended to exist forever. Basically "unless something changes in the app, this will always be used". But since this includes a user value in it (the flow id) this is likely to be removed and the local storage key won't be cleaned up.

src/components/FlowListItem.vue
```diff
@@ -0,0 +1,198 @@
...
+    flow: Flow,
+    filter?: FlowsFilter,
+    disabled?: boolean,
+  }>()
+
+  const emit = defineEmits<{
+    (event: 'delete', value: string): void,
+  }>()
+
+  const { FlowMenu } = useComponent()
```
pleek91, line 90: thank you for retaining this and not breaking the "automate" menu item ❤️

src/components/FlowListItem.vue
```diff
@@ -0,0 +1,198 @@
...
+      },
+    }
+  })
+
+  const deploymentsCountSubscriptionArgs = computed<[DeploymentsFilter]>(() => [filter.value])
+  const deploymentsCountSubscription = useSubscriptionWithDependencies(
+    api.deployments.getDeploymentsCount,
+    deploymentsCountSubscriptionArgs,
+  )
+  const deploymentsCount = computed(() => deploymentsCountSubscription.response ?? 0)
```
pleek91, line 112: a `useDeploymentsCount` composition would be nice to reduce the boilerplate here and in other places in the app.

src/components/FlowList.vue
```diff
@@ -0,0 +1,158 @@
...
+      nameLike: searchDebounced,
+    },
+    offset,
+    limit: DEFAULT_LIMIT,
+  })
+
+  const selected = ref<string[]>([])
+
+  const flowsSubscription = useSubscription(api.flows.getFlows, [filter])
+  const flows = computed(() => flowsSubscription.response ?? [])
```
pleek91, line 105: There's a `useFlows` composition for this. It currently doesn't do the `?? []` piece but I think it probably should.

src/components/FlowList.vue
```diff
@@ -0,0 +1,158 @@
...
+
+  const flowsCountSubscription = useSubscription(api.flows.getFlowsCount, [filter])
+  const flowsCount = computed(() => flowsCountSubscription.response)
```
pleek91, line 108: Same as deployments. A `useFlowsCount` composition would be sweet for this.

src/components/FlowList.vue
```diff
@@ -0,0 +1,158 @@
+<template>
+  <div class="flow-list">
+    <p-layout-table sticky>
+      <template #header-start>
+        <slot name="header-start">
+          <div class="flow-list__header-start">
+            <ResultsCount v-if="selected.length == 0" :label="localization.info.flow" :count="flowsCount" />
+            <SelectedCount v-else :count="selected.length" />
```
pleek91, line 8: I really want to combine these components into something like this where the `selected` prop is optional. 

```
<ResultsCount :count="10" :selected="4" :label="localization.info.flow" />
```

src/components/DeploymentListItem.vue
```diff
@@ -0,0 +1,141 @@
...
+
+    <div class="deployment-list-item__schedule">
+      <DeploymentToggle :deployment="deployment" :disabled="!schedule" @update="handleUpdate" />
+
+      <div class="deployment-list-item__schedule-icon-text">
+        <p-icon icon="ClockIcon" />
+
+        <template v-if="schedule">
+          <div class="deployment-list-item__schedule-text">
+            <span :title="schedule?.toString({ verbose: true })">{{ scheduleText }}</span>
```
pleek91, line 44: is the `?` necessary since there's a `v-if="schedule"` on line 42? Is that just volar being dumb?

pleek91: Also, we have a `p-tooltip` component now. Which would be a better ux than the `title` attribute IMO.

pleek91 (commented):
Sorry for so many comments. There's a lot here that's really good. And I love the direction. Just wanting to really iron this out since its a significant rewrite.

src/components/PageHeadingFlows.vue
znicholasbrown: Chatted with bill about this - for now, no, but we'll retain the components and routes so we don't break any perm links

demo/compositions/useFlowMock.ts
```diff
@@ -1,29 +1,62 @@
...
-  const flowRuns = mocker.createMany('flowRun', mocker.create('number'), [
-    {
-      flowId: flow.id,
-      deploymentId: deployment.id,
-      workQueueName: workQueue.name,
-    },
-  ])
+
+  const workPools = mocker.createMany('workPool', mocker.create('number', [0, 2]))
+  const workPoolNames = [...workPools.map((workPool) => workPool.name), null]
```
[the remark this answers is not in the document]
znicholasbrown: yeah - our model has this as a nullable field so i figured the mock should match

src/components/FlowListItem.vue
znicholasbrown: Makes sense to me - if we want to retain this behavior cross-session we can come up with a more robust solution at that point 👍🏻

src/models/StateType.ts
znicholasbrown: Nope, this shouldn't have been added at all! Removed

src/components/FlowList.vue
znicholasbrown: Nice idea, [tracked here](https://github.com/PrefectHQ/prefect-ui-library/issues/1350)

src/components/DeploymentListItem.vue
znicholasbrown: Ah! That was directly from the previous times we've used it, it looks like volar has fixed that!

Agreed, I forgot we had that, will update

znicholasbrown:
So I went ahead and wrote `useDeployments`, `useDeploymentsCount`, `useFlowsCount` compositions and modified the existing `useFlows` composition to add a new overload for passing flow ids or a flows filter; lmk what you think of that implementation. The one caveat is that I retained permissions checks in the filter the composition uses, which means that technically every component will still need to do `objectList ?? []` because there's a chance a user is missing permissions and the subscription never runs, according to the `UseEntitySubscription` overload. We could relax that type a bit but I think it makes sense as a generic and the `undefined` return might be a useful indicator.

All other feedback has also been addressed 😄

pleek91 (approved):
LGTM. Thanks for working through all the details with me!

PR #1347, opened 2023-04-18
Flat FlowRunListItem, updated FlowRunDetails

brandonreid:
1. Create a `flat` variant of the FlowRunListItem
2. Remove information from the FlowRunDetails that is duplicate of what's found on the FlowRunListItem
3. Add a isPendingState function for checking if a flow run has not started yet

This is needed to allow the Flow Run page to remove the fixed well and be more full screen. Here's what it will look like (will need some minor updates in Nebula/OSS of course):

Before a flow has ran:
image

After a flow run has started:
image

This is helpful because next we'll be adding an Events visualization to the page, and both graphs will utilize pop outs to the right. This is also a step toward incorporating more into the visualizations and clarifying the page.

### DIFF
--- a/demo/sections/flowRuns/FlowRunListItem.vue
+++ b/demo/sections/flowRuns/FlowRunListItem.vue
@@ -9,6 +9,10 @@
     <template #no-relations>
       <FlowRunListItem :flow-run="flowRunWithNoRelations" :selected="[]" class="color-mode-default" disabled />
     </template>
+
+    <template #flat>
+      <FlowRunListItem :flow-run="flowRun" :selected="[]" class="color-mode-default" flat />
+    </template>
   </ComponentPage>
 </template>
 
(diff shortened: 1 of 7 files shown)


pleek91:
Overall I like this change. But I'm not sure about reusing the FlowRunListItem for the heading for a couple reasons
1. Every page in the app has the same page heading style/ux. But the FLowRunListItem's crumbs are smaller and use a different separator. 
2. Quite a few conditions had to be added to both the FlowRunListItem and the StateListItem components. Increasing their complexity

Could we leave the existing components as is and just abstract any pieces from them that are needed to accommodate reuse on the flow run page? That way we could keep using the `page-heading` component for consistency but still show the additional flow run information. That component already has slots for the necessary layout I believe.

brandonreid:
@pleek91 I'm down for that. Made a new PR for that approach: https://github.com/PrefectHQ/prefect-ui-library/pull/1348

Closing this one.

PR #1348, opened 2023-04-18
update FlowRunPageHeader and FlowRunDetails for full page layout

brandonreid:
Updates the FlowRunPageHeader to include top level meta details consistent with the FlowRunListItem. FlowRunDetails have also been updated to remove duplicate information.

Once this is deployed and updates are made in OSS and Nebula, the result will look like these examples:

A flow run has completed and has no worker/deployment/etc:
image

A flow run has not ran yet:
image

### DIFF
--- a/src/components/PageHeadingFlowRun.vue
+++ b/src/components/PageHeadingFlowRun.vue
@@ -1,26 +1,77 @@
 <template>
-  <page-heading v-if="flowRun" class="page-heading-flow-run" :crumbs="crumbs">
-    <template #after-crumbs>
-      <StateBadge :state="flowRun.state" />
-    </template>
-    <template #actions>
-      <template v-if="media.sm">
-        <FlowRunPauseButton :flow-run="flowRun" />
-        <FlowRunResumeButton :flow-run="flowRun" />
-        <FlowRunRetryButton :flow-run="flowRun" />
-        <FlowRunCancelButton :flow-run="flowRun" />
+  <div v-if="flowRun" class="page-heading-flow-run">
+    <page-heading class="page-heading-flow-run__heading" :crumbs="crumbs">
+      <template #after-crumbs>
+        <div v-if="flowRun.tags && flowRun.tags.length > 0" class="state-list-item__tags">
+          <p-tag-wrapper v-bind="{ tags: flowRun.tags }" />
+        </div>
       </template>
-      <FlowRunMenu :flow-run-id="flowRun.id" :show-all="!media.sm" @delete="emit('delete')" />
-    </template>
-  </page-heading>
+      <template #actions>
+        <template v-if="media.sm">
+          <FlowRunPauseButton :flow-run="flowRun" />
+          <FlowRunResumeButton :flow-run="flowRun" />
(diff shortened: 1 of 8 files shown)


src/components/PageHeadingFlowRun.vue
```diff
@@ -1,26 +1,77 @@
...
+        <template v-if="media.sm">
+          <FlowRunPauseButton :flow-run="flowRun" />
+          <FlowRunResumeButton :flow-run="flowRun" />
+          <FlowRunRetryButton :flow-run="flowRun" />
+          <FlowRunCancelButton :flow-run="flowRun" />
+        </template>
+        <FlowRunMenu :flow-run-id="flowRun.id" :show-all="!media.sm" @delete="emit('delete')" />
+      </template>
+    </page-heading>
+    <div class="page-heading-flow-run__meta">
```
pleek91, line 19: I believe this meta can go in the default slot of the `page-heading` component. That would simplify this a bit and remove the need for the root div. 

Maybe you already tried this and it didn't work right?

src/components/PageHeadingFlowRun.vue
```diff
@@ -1,26 +1,77 @@
...
-        <FlowRunRetryButton :flow-run="flowRun" />
-        <FlowRunCancelButton :flow-run="flowRun" />
+      <FlowRunStartTime :flow-run="flowRun" />
+      <DurationIconText :duration="flowRun.duration" />
+      <template v-if="!isPending">
+        <FlowRunTaskCount :tasks-count="taskRunsCount">
+          <template #default="{ count }">
+            {{ count }} task {{ toPluralString('run', count) }}
+          </template>
+        </FlowRunTaskCount>
```
pleek91, line 28: We don't use this `FlowRunTaskCount` component anywhere without using this custom slot that I can find. Can you just update the component to do this by default?

src/components/PageHeadingFlowRun.vue
```diff
@@ -1,26 +1,77 @@
...
+        <div class="flow-run-list-item__relation">
+          <span>Work Pool</span> <WorkPoolIconText :work-pool-name="flowRun.workPoolName" />
+        </div>
+      </template>
+
+      <template v-if="flowRun.workQueueName">
+        <div class="flow-run-list-item__relation">
+          <span>Work Queue</span> <WorkQueueIconText :work-queue-name="flowRun.workQueueName" :work-pool-name="flowRun.workPoolName" />
+        </div>
       </template>
```
pleek91, line 48: Might be a good time to abstract these into separate `FlowRunDeployment`, `FlowRunWorkPool`, and `FlowRunWorkQueue` components. And share between this and the list item component.

pleek91 (commented):
@brandonreid yeah this is great. Couple suggestions for moving things around but otherwise LGTM.

src/components/PageHeadingFlowRun.vue
brandonreid: Good looking out on this!

pleek91 (approved):
LGTM! Thanks for making those changes

PR #1351, opened 2023-04-19
Enhancement: Prefer concepts over ui for docs links where possible

zhen0:
The UI historically linked to the ui section of the docs instead of the main concepts.  That feels a bit duplicative and I believe we've made those UI pages less prominent in the docs so I've added a PR to update the UI links to go to the main concepts.  E.g. for workers that means we link to the more complete (and more fully titled)
https://docs.prefect.io/latest/concepts/work-pools/ instead of https://docs.prefect.io/latest/ui/work-pools/

For notifications and automations, there are no main concepts pages so I've left as is. 
For work queues - I think the link will soon be deprecated and it could be confusing to link to work pools so I've left as is.


pleek91 (approved):

billpalombi (approved):
This is much better! Thank you!

PR #1353, opened 2023-04-20
Enhancement: Add edit infra overrides on deployment form

zhen0:
Uses the JSON input component to add infrastructure overrides to the deployment edit form. This may be needed for new worker types.  

image

https://user-images.githubusercontent.com/40272060/233420691-e4375c82-3369-4811-a5b5-30ccc2e028e8.mov

https://user-images.githubusercontent.com/40272060/233421126-5d28076b-feae-463e-b519-4c50d0638fca.mov


src/components/DeploymentForm.vue
```diff
@@ -119,14 +134,21 @@
...
 
   const emit = defineEmits<{
     (event: 'submit', value: DeploymentUpdate): void,
     (event: 'cancel'): void,
   }>()
 
   const submit = handleSubmit((values) => {
-    emit('submit', values)
+    try {
+      JSON.parse(infrastructureOverrides.value)
```
pleek91, line 146: Is this parse attempting to validate the json? We have a validation rule for that which we should use for this sort of thing.

src/components/DeploymentForm.vue
```diff
@@ -61,6 +61,17 @@
...
     </p-content>
 
+    <p-divider />
+
+    <p-content>
+      <h3 class="deployment-form__section-header">
+        Infrastucture overrides
+      </h3>
+
+      <JsonInput v-model:model-value="infrastructureOverrides" show-format-button />
```
pleek91, line 71:
```suggestion
      <JsonInput v-model="infrastructureOverrides" show-format-button />
```

src/components/DeploymentForm.vue
```diff
@@ -61,6 +61,17 @@
       </p-content>
     </p-content>
 
+    <p-divider />
+
+    <p-content>
+      <h3 class="deployment-form__section-header">
+        Infrastucture overrides
```
znicholasbrown, line 68:
```suggestion
        Infrastructure overrides
```
would be good to move this to the localization module

src/components/DeploymentForm.vue
```diff
@@ -119,14 +134,21 @@
...
 
   const submit = handleSubmit((values) => {
-    emit('submit', values)
+    try {
+      JSON.parse(infrastructureOverrides.value)
+      emit('submit', values)
+    } catch (error) {
+      showToast(localization.error.invalidJSON, 'error')
+      throw new Error('Invalid JSON')
+    }
```
znicholasbrown, line 151: Instead of this bespoke validation, we have an `isValidJson` validation utility you can pass to the field which will prevent form submission

src/models/Deployment.ts
```diff
@@ -51,7 +51,7 @@ export class Deployment implements IDeployment {
   public entrypoint: string | null
   public storageDocumentId: string | null
   public infrastructureDocumentId: string | null
-  public infrastructureOverrides: Record<string, unknown> | null
+  public infrastructureOverrides: unknown | string | null
```
pleek91, line 54: I think we probably don't want to change this type. But in the component we can convert it to json and back.

src/models/DeploymentUpdate.ts
```diff
@@ -8,6 +8,7 @@ type Base = {
   tags?: string[] | null,
   workQueueName?: string | null,
   workPoolName?: string | null,
+  infrastructureOverrides?: string | null,
```
pleek91, line 11: I'd also expect this to be a record.

src/models/api/DeploymentUpdateRequest.ts
```diff
@@ -13,4 +13,5 @@ export type DeploymentUpdateRequest = Partial<{
   infrastructure_document_id: string | null,
   work_queue_name: string | null,
   work_pool_name: string | null,
+  infra_overrides: unknown | null,
```
pleek91, line 16: same here

znicholasbrown, line 16: For the request there's definitely a type the API is expecting, I don't think `unknown` is appropriate here

src/components/DeploymentForm.vue
pleek91: https://github.com/PrefectHQ/prefect-ui-library/blob/main/src/utilities/validation.ts#L246

src/models/DeploymentUpdate.ts
znicholasbrown, line 11: Is this correct? The API has this listed as an `object` in the payload.

znicholasbrown (commented):
I'm not able to delete my comments for some reason, apologies for any duplicates with @pleek91's review

src/components/DeploymentForm.vue
```diff
@@ -119,14 +134,25 @@
...
 
   const emit = defineEmits<{
     (event: 'submit', value: DeploymentUpdate): void,
     (event: 'cancel'): void,
   }>()
 
   const submit = handleSubmit((values) => {
-    emit('submit', values)
+    try {
+      isJson(infrastructureOverrides.value)
```
zhen0, line 146: Not sure this is necessary here anymore?

src/components/DeploymentForm.vue
```diff
@@ -119,14 +134,23 @@
...
-    emit('submit', values)
+    try {
+      const deploymentUpdate: DeploymentUpdate = {
+        ...values,
+        infrastructureOverrides: JSON.parse(infrastructureOverrides.value),
+      }
+      emit('submit', deploymentUpdate)
+    } catch (error) {
+      showToast(localization.error.invalidJSON, 'error')
+    }
```
pleek91, line 153: I think we're still missing the `isJson` validation rule for the overrides. Once that's added then I think this try catch can be removed. You won't be able to get here without the json being valid.

src/models/DeploymentUpdate.ts
```diff
@@ -8,16 +8,38 @@ type Base = {
...
+}
+
+type BaseEdit = {
+  description?: string | null,
+  schedule?: Schedule | null,
+  isScheduleActive?: boolean,
+  tags?: string[] | null,
+  workQueueName?: string | null,
+  workPoolName?: string | null,
+  infrastructureOverrides?: string,
```
pleek91, line 21: Why does this need to be a string? I thought it was being parsed back into an object when submitting the form.

zhen0: That's the model for the deployment within the form only - so for when we are editing the deployment.  It does get parsed back and we use the deploymentUpdate model everywhere else.

pleek91: ah in the component. Got it. To cut down on the duplication I'd suggest

```typescript
type BaseEdit = Omit<Base, 'infrastructureOverrides'> & {
  infrastructureOverrides?: string
}
```

zhen0: Nice!

src/components/DeploymentForm.vue
```diff
@@ -119,14 +134,20 @@
...
+  const { value: infrastructureOverrides } = useField<string>('infrastructureOverrides')
 
   const emit = defineEmits<{
     (event: 'submit', value: DeploymentUpdate): void,
     (event: 'cancel'): void,
   }>()
 
   const submit = handleSubmit((values) => {
-    emit('submit', values)
+    isJson(infrastructureOverrides.value)
```
zhen0, line 145: I've updated this @pleek91 but I don't like it as much from a UX perspective as there's no alert to tell the user the json is invalid.  Any thoughts on the best way to do that?

zhen0: Ohh.... sorry I think I know.  Ignore

zhen0: 😞 Sorry @pleek91 - should be updated now

pleek91 (approved):

PR #1358, opened 2023-04-21
Chore: Remove vee-validate from schema forms [WIP]

znicholasbrown:
This PR seeks to remove vee-validate from schema logic by moving to a `modelValue` model instead. 

This is a work in progress and is a potentially breaking change for several of our schema-generated forms.

### DIFF
--- a/src/components/DeploymentForm.vue
+++ b/src/components/DeploymentForm.vue
@@ -76,7 +76,7 @@
       <p-button inset @click="cancel">
         Cancel
       </p-button>
-      <p-button type="submit">
+      <p-button :loading="pending" type="submit">
         Save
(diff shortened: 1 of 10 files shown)


znicholasbrown:
Technically this already works for the deployments form (I was using that as a POC) but I seem to have reverted the behavior fixed in #1356 😞

pleek91:
Oh man here we go! This is exciting.

brandonreid (approved):
Not sure how thorough I should review here, but on the surface I don't see any red flags or anything. Let er rip, I say!

znicholasbrown:
We'll still do this at some point but this PR is too stale

PR #1359, opened 2023-04-21
Feature: `useModelValue` composition POC

znicholasbrown:
A boilerplate reducer, used like this:

```vue
<template>
  <elem v-model="value" />
</template>

<script lang="ts" setup>
import { useModelValue } from '@/compositions'

const { value } = useModelValue<MyType>()
</script>
```


znicholasbrown:
nevermind 😠

PR #1361, opened 2023-04-21
Bugfix: Fix missing flow link in flow run page heading component

znicholasbrown:


zhen0 (approved):
🙏

brandonreid (approved):
Thank you!!

znicholasbrown:
Note: this also contains a fix for the other relation components since they were implicitly relying on a relation class inherited from the flow run list item component, which is not guaranteed to be present and so would cause the components to lose flex styling

pleek91:
Nice. Thanks @znicholasbrown

PR #1362, opened 2023-04-21
Enhancement: Add options to the deployment tags select

zhen0:
Before:
image

After:

image

Closes https://github.com/PrefectHQ/prefect/issues/9272 
Closes https://github.com/PrefectHQ/prefect/issues/8791

### DIFF
--- a/src/components/DeploymentsTable.vue
+++ b/src/components/DeploymentsTable.vue
@@ -20,7 +20,7 @@
 
           <p-select v-model="filter.sort" :options="deploymentSortOptions" />
 
-          <p-tags-input v-model="filter.deployments.tags.name" empty-message="All tags" class="deployments-table__tags" multiple />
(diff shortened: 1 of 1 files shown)


pleek91:
@zhen0 I like this. Thoughts on moving all the logic into a DeploymentTags component? Also, @znicholasbrown thoughts on this with the work your doing for flows/deployments?

znicholasbrown:
We can try to support this in the new page as well but this work won't be used in the new page as-is

zhen0:
Good point about @znicholasbrown's work.  Hopefully we can still keep the functionality even if we don't keep the code. 

Since any component will be obsolete after Nicholas's updates, are you ok to keep without creating a new component @pleek91?

MattDelac:
@taylor-curran that's an old page "Deployment". Now all the deployments are under the "Flow" page
image

PR #1365, opened 2023-04-22
Enhancement: Use p-textarea instead of p-text-input for strings in schemas 

zhen0:
Closes https://github.com/PrefectHQ/prefect/issues/8447

Gives a bigger input box to better handle larger strings

After:

image

Before:

image


pleek91:
@zhen0 im not sure about this change. This would apply to all block schemas. A lot of which have properties that are strings which I'm not sure benefit from this change. The forms will get really long.

zhen0:
That's a good call out but I'm not sure it makes that much difference to form length:

image

image

Checking that made me realize that the textarea does not handle passwords so well so I've added a check for that and added back text-input there. 

Happy to discuss further though if you feel strongly.

brandonreid (approved):

PR #1369, opened 2023-04-24
Enhancement: Reactive flow/compositions, improved flow list header

znicholasbrown:
This PR addresses an issue where the flow list filters weren't working with the new flow compositions. This was because the `use*FiltersFromRoute` compositions return a reactive object that (correctly) is never reassigned and the current implementations of many of our filterable compositions use a computed property that acts on the entire object and doesn't watch sub properties. 

I suspect we'll want to do something similar in a few other compositions but for now I've kept the changes limited only to the `useFlows`, `useFlowsCount`, and `useDeploymentsCount` compositions

The PR also fixes some scaling issues with the flow list header


pleek91:
Interesting, so a reactive doesn't trigger a computed unless you access a property. I guess that makes sense. Vue has no way of knowing it should track it.

pleek91 (approved):

Issue #1330, opened 2023-04-06
UI - Task run results well


znicholasbrown:
Add the latest task run result (artifact of type `result`) to the task run details component

PR #1372, opened 2023-04-24
Enhancement: Task run result details

znicholasbrown:
Adds task run results to the task run details component:

Screenshot 2023-04-24 at 3 38 38 PM


znicholasbrown:
Resolves: #1330

src/components/TaskRunDetails.vue
```diff
@@ -12,6 +12,14 @@
       </template>
     </p-key-value>
 
+    <template v-if="result">
+      <p-key-value :label="localization.info.result" :alternate="alternate">
+        <template v-if="result.description" #value>
+          <p-markdown-renderer v-if="result.description" :text="result.description" class="task-run-details__markdown-renderer" />
```
brandonreid, line 18: Should the `v-if="result.description"` just be combined with the top level <template> check? I'd imagine you wouldn't care to show the label without the description.

Also looks like there's multiple checks for the description, maybe Typescript isn't clever enough to know it's already been checked for in the containing `<template #value>`?

src/components/TaskRunDetails.vue
```diff
@@ -57,28 +65,41 @@
...
+}
+
+.task-run-details__markdown-renderer {
+  font-size: inherit;
+}
+
+/*
+  Note: The core client writes the description as a paragraph token, so we need to override the default margin.
+*/
+.task-run-details__markdown-renderer .markdown-renderer__text--paragraph { @apply
```
brandonreid, line 102: No way to target a native `<p>` tag or something instead of the mutable nested class selector eh?

src/components/TaskRunDetails.vue
```diff
@@ -57,28 +65,41 @@
...
+.task-run-details { @apply
+  flex
+  flex-col
+  gap-3
+  items-start
+}
+
+.task-run-details__tags { @apply
+  mb-1
+  mr-1
```
brandonreid, line 92: Could grid styling be used instead to avoid hanging margins? I realize you didn't actually change this, so nbd.

src/components/TaskRunDetails.vue
pleek91: Usually a v-if on the `#value` ref of a `p-key-value` means that the `None` should show up.

src/compositions/useTaskRunResult.ts
```diff
@@ -0,0 +1,45 @@
...
+import { Artifact, ArtifactsFilter } from '@/models'
+import { WorkspaceArtifactsApi } from '@/services/WorkspaceArtifactsApi'
+import { MaybeRef } from '@/types'
+
+export type UseTaskRunResult = {
+  subscription: UseSubscription<WorkspaceArtifactsApi['getArtifacts']>,
+  result: ComputedRef<Artifact | undefined>,
+}
+
+export function useTaskRunResult(taskRunId: MaybeRef<string | null | undefined>): UseTaskRunResult {
```
pleek91, line 14: should this use the watch pattern you used to account for reactive filters?

src/components/TaskRunDetails.vue
znicholasbrown: Yup @pleek91 is correct here - that's potentially not super clear from reading the markup but it's how the `PKeyValue` component works (and how we use it elsewhere)

src/compositions/useTaskRunResult.ts
znicholasbrown: I don't think we need to unless there's a performance consideration there; this is a primitive ref so changes should be adequately captured by the `computed`

src/components/TaskRunDetails.vue
znicholasbrown: I didn't test this without since these styles existed before the PR (they had inconsistent indentation) - @brandonreid feel free to PR what you had in mind there if you feel strongly but I don't want to experiment with that in this PR

src/compositions/useTaskRunResult.ts
pleek91: oh right, this doesn't take a filter. Ignore me.

src/components/TaskRunDetails.vue
pleek91: With that PR I think this can just go away then right?

znicholasbrown: Yup - removed

pleek91 (approved):

PR #1384, opened 2023-04-26
Enhancement: Schema / deployment form

znicholasbrown:
This PR makes some minor improvements to deployment forms and fixes a few long standing issues with schemas. One bigish change is the removal of `PSelect` in list properties; this is to work around some incompatibilities between the component and the intent of list properties (in particular duplicate values and ordering)

### DIFF
--- a/src/components/JsonInput.vue
+++ b/src/components/JsonInput.vue
@@ -1,5 +1,5 @@
 <template>
-  <p-code-input v-model="internalValue" lang="json" class="json-input">
+  <p-code-input v-model="internalValue" lang="json" class="json-input" v-bind="{ showLineNumbers, minLines }">
     <template v-if="showFormatButton" #append>
       <p-button class="json-input__prettify-button" size="xs" @click="format">
         Format
@@ -17,8 +17,10 @@
   import { stringify } from '@/utilities/json'
 
   const props = defineProps<{
+    minLines?: number,
     modelValue: string | undefined,
     showFormatButton?: boolean,
+    showLineNumbers?: boolean,
   }>()
 
   const emit = defineEmits<{
--- a/src/components/SchemaFormInput.vue
+++ b/src/components/SchemaFormInput.vue
@@ -13,7 +13,7 @@
     </template>
 
     <template v-if="meta">
-      <component :is="meta.component" v-model="propValue" v-bind="{ ...meta.props, ...meta.attrs }" />
+      <component :is="meta.component" v-model="propValue" class="schema-form-input__component" v-bind="{ ...meta.props, ...meta.attrs }" />
     </template>
   </p-label>
 </template>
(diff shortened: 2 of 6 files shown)


src/components/JsonInput.vue
pleek91, line 1: This component is just for the format button basically?

src/components/SchemaFormInput.vue
```diff
@@ -50,6 +50,13 @@
   grid
   gap-1
 }
+
+.schema-form-input__component { @apply
+  max-w-full
+  min-h-[2.5rem]
+  resize-y
+}
```
pleek91, line 58: This seems a little too specific to apply to all schema components. The max width makes sense but the min-h and resize?

src/components/JsonInput.vue
```diff
@@ -17,8 +17,10 @@
   import { stringify } from '@/utilities/json'
 
   const props = defineProps<{
+    minLines?: number,
```
znicholasbrown, line 20: I added these purely for the schema form `withPropsWithoutExcludedFactory` type, which only pulls props that are exposed from the component, not any that it wraps; there might be a better way to handle this

src/components/SchemaFormInput.vue
znicholasbrown: I can target the code inputs instead, wdyt?

src/components/JsonInput.vue
znicholasbrown: Yeah, I don't think the `PCodeInput` should support a format button since it only applies to 1 language right now

znicholasbrown:
Form comparison:

Before:

https://user-images.githubusercontent.com/27291717/234671820-7173104c-0f0f-406c-abd0-a0fba4668af9.mov


After:

https://user-images.githubusercontent.com/27291717/234671847-2fbe1023-c93c-49a7-9e82-2eb7c7d00b48.mov

src/components/SchemaFormInput.vue
pleek91: I'm not sure, this just seems overly generic for this. Could be wrong 🤷

src/components/JsonInput.vue
pleek91: this makes sense to me. This is a general problem that soon we'll be able to fix (in vue 3.3) by making props their own types and exporting them from the component itself (cannot export multiple from a sfc today). Then in a component like this you could import the type and add it to your props (cannot use an imported type for defineProps today). Both fixed in 3.3

src/components/SchemaFormInput.vue
znicholasbrown: I added a more targeted selector for now cause I agree that was too generic. That'll let us close some tickets around resizing these inputs too

pleek91 (approved):

PR #1392, opened 2023-04-28
BugFix: Add check for color scheme to ensure correct favicon is updated

zhen0:
Our new favicons are color scheme/appearance (light/dark mode) aware so we now have more than one 'favicon-32' and 'favicon-16'. This PR adds a check for color mode and selects and updates the correct favicon. 

Partner to https://github.com/PrefectHQ/prefect/pull/9370 and https://github.com/PrefectHQ/nebula-ui/pull/2706 which set the correct id based on color scheme selection.


src/compositions/useFavicon.ts
```diff
@@ -1,9 +1,33 @@
...
+      favicon32 = document.getElementById('favicon-32-dark')
+      break
+    case 'light':
+      favicon16 = document.getElementById('favicon-16-light')
+      favicon32 = document.getElementById('favicon-32-light')
+      break
+    default:
+      favicon16 = document.getElementById('favicon-16-no-preference')
+      favicon32 = document.getElementById('favicon-32-no-preference')
+      break
```
pleek91, line 29: We have a specific favicon for no preference?

src/compositions/useFavicon.ts
```diff
@@ -1,9 +1,33 @@
...
 import { StateType } from '@/models/StateType'
 
+function getPreferredColorScheme(): 'dark' | 'light' | 'no-preference' {
+  if (window.matchMedia('(prefers-color-scheme: dark)').matches) {
+    return 'dark'
+  } if (window.matchMedia('(prefers-color-scheme: light)').matches) {
+    return 'light'
+  }
+  return 'no-preference'
+}
```
pleek91, line 11: Should this be based on the prefect-design mode? That way if a user switches their color settings this reflects that change.

znicholasbrown: Nope, this is a system setting - the background of the tab/window is what's important, not the application

src/compositions/useFavicon.ts
znicholasbrown: We do not - we default to the dark (official) one

src/compositions/useFavicon.ts
```diff
@@ -1,9 +1,33 @@
...
+  let favicon32: HTMLElement | null
+
+  switch (getPreferredColorScheme()) {
+    case 'dark':
+      favicon16 = document.getElementById('favicon-16-dark')
+      favicon32 = document.getElementById('favicon-32-dark')
+      break
+    case 'light':
+      favicon16 = document.getElementById('favicon-16-light')
+      favicon32 = document.getElementById('favicon-32-light')
```
znicholasbrown, line 24:
```suggestion
  default:
      favicon16 = document.getElementById('favicon-16')
      favicon32 = document.getElementById('favicon-32')
```

src/compositions/useFavicon.ts
znicholasbrown, line 29:
```suggestion
```
Can remove these in favor of the case above

zhen0: Any objection to removing the `prefers-color-scheme:no-preference` icon in the prefect and nebula-ui repos to go alongside this?  It looks like it's deprecated anyways and not supported anywhere but safari (which seems to be doing its own thing with favicons and ignoring the useFavicon change anyways).

https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme

Safari in light and dark when all other browsers are updating the favicon:
image

image

znicholasbrown: No objection from me - Safari might handle favicons a little differently, we've got a `mask-icon` link in the head which might be what Safari is pulling instead but I'm not sure

znicholasbrown (approved):

Issue #1269, opened 2023-03-28
When navigating to a Sub Flow Run from a Flow Run Details page, the graph doesn't remount.


brandonreid:
Steps to reproduce:

- View a flow run that contains sub flow runs.
- On the graph, select the sub flow run to view it's details in the side panel.
- Click the flow run graph's name to navigate to it.

Result:
The graph state is stuck rendering the parent flow run.

If you refresh on this page, you'll see the graph render with just the sub flow run nodes.

Expected:
The graph should rerender to show the same state shown on a fresh load.

pleek91:
Fixed the reactivity issue in prefect-ui-library's FlowRunTimeline component in #1393 but that revealed an issue within the timeline itself where its unable to remove nodes. Closing that PR for now until we can address this more holistically.

PR #1393, opened 2023-04-28
BugFix: Flow run timeline doesn't react to route changes

pleek91:
# Description
There was a reactivity issue in the FlowRunTimeline component which caused it to not update when `props.flowRun` was changed. Passing a computed flow run id into the subscription rather than `props.flowRun.id` fixes that. 

Also implemented the `useStatePolling` composition to prevent this from polling longer than it should

@brandonreid I tested this but it would be great if you could also test this. You would probably catch if something unexpected is happening. 

Closes https://github.com/PrefectHQ/prefect-ui-library/issues/1269


znicholasbrown (commented):
This looks ok to me but would love if @brandonreid could give the final sign off

brandonreid:
I've got this tab open and will totally scope it out when demo stuff is done. Really appreciate you picking this up!

brandonreid:
So testing this locally, it doesn't appear to clear the graph and redraw with just the sub-flow data. I may not have it setup properly, but even still, I know that the graph is not quite setup to manage new, completely different data. This is in part for performance, so it can cache already drawn things and add new ones when a flow run is running. I had operated under the assumption that new data would trigger a remount. Without pretty significant changes in the graphs repo, some efforts to check the tree and cleanup no-longer existing nodes, etc. the fastest and most simple route is likely to remount the graph when the route changes.

If this actually works for you and I'm somehow mistaken there, then heck yeah.

pleek91:
ah I see 
image

Probably something internal to the chart we can update to fix that. Caching makes sense but it seems like the ability to remove nodes from the graph isn't taken into consideration. 

This reactivity fix is valid and necessary. But until we fix that this cannot be merged. So I'm going to close this and make a note on the original issue.

PR #1398, opened 2023-04-28
Bugfix: `anyOf` and `array` schema mapping

znicholasbrown:
This PR makes two important changes to the way array values are mapped in schemas:

1. The property array response mapper first checks if a value is already a JSON string before attempting to stringify it for use in code input components
2. Schemas attempt to parse a value when deciding which union property to display

There's still a bug here where validators for fieldsets in the union will apply to the incorrect member, but I'm working on that

### DIFF
--- a/src/services/schemas/properties/SchemaPropertyArray.ts
+++ b/src/services/schemas/properties/SchemaPropertyArray.ts
@@ -35,6 +35,10 @@ export class SchemaPropertyArray extends SchemaPropertyService {
 
   protected response(value: SchemaValue): unknown {
     if (this.componentIs(JsonInput)) {
+      if (typeof value === 'string') {
+        return value
+      }
+
       return stringifyUnknownJson(value)
     }
 
--- a/src/services/schemas/utilities.ts
+++ b/src/services/schemas/utilities.ts
@@ -232,14 +232,10 @@ export function getSchemaValueDefinitionIndex(definitions: Schema[], value: Sche
     return definitions.findIndex(definition => definition.type === 'block')
   }
 
-  try {
-    const parsedValue = parseUnknownJson(value)
+  const parsedValue = parseUnknownJson(value)
 
-    if (isRecord(parsedValue) || Array.isArray(parsedValue)) {
-      return findObjectDefinitionIndex(definitions, parsedValue)
-    }
-  } catch {
-    // do nothing, use the raw value
+  if (isRecord(parsedValue) || Array.isArray(parsedValue)) {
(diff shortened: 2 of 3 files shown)


src/services/schemas/utilities.ts
```diff
@@ -231,15 +231,23 @@ export function getSchemaValueDefinitionIndex(definitions: Schema[], value: Sche
...
   }
 
-  switch (typeof value) {
+  let parsedValue = value
+
+  try {
+    parsedValue = parseUnknownJson(value)
+  } catch {
+    // do nothing, use the raw value
+  }
```
pleek91, line 240: Isn't this going to mess up some types that could be parsed as json but are actually just supposed to be strings?

src/services/schemas/properties/SchemaPropertyArray.ts
```diff
@@ -35,6 +35,10 @@ export class SchemaPropertyArray extends SchemaPropertyService {
 
   protected response(value: SchemaValue): unknown {
     if (this.componentIs(JsonInput)) {
+      if (typeof value === 'string') {
+        return value
+      }
```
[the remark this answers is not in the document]
znicholasbrown: In the response? I don't _think_ we need to. So long as it's a string, the validators at the component level should catch that and the json input components don't really care if it's valid

src/services/schemas/utilities.ts
znicholasbrown: For posterity: after chatting we've decided to add a check on the `parsedValue`'s type as well and only continue with objects so as not to unintentionally parse strings that contain numbers, booleans, etc

src/services/schemas/utilities.ts
```diff
@@ -231,6 +232,16 @@ export function getSchemaValueDefinitionIndex(definitions: Schema[], value: Sche
...
 
+  try {
+    const parsedValue = parseUnknownJson(value)
+
+    if (isRecord(parsedValue) || Array.isArray(parsedValue)) {
+      return findObjectDefinitionIndex(definitions, parsedValue)
+    }
+  } catch {
+    // do nothing, use the raw value
+  }
```
pleek91, line 243: We don't need a try catch here do we? `parseUnknownJson()` already silences the error.

znicholasbrown: Ah good point - updated

znicholasbrown:
Also found the source of the non-reactive field-level validators @pleek91 so those are fixed in this as well

pleek91 (approved):

PR #1405, opened 2023-05-02
Enhancement: Parameters JSON input

znicholasbrown:
This PR augments the flow run creation and deployment forms to include a JSON code input for bulk parameter input.

Note: for now, this input is only validated as JSON and not against the schema. This is because validating against the schema with our current system would be overly complicated and probably should be happening in the API anyway. In addition the inputs do not share state -- toggling between form and code input will only save your state for the form that's selected, and changes made to 1 will not impact the other.


pleek91 (approved):
This looks fine to me.

Issue #1399, opened 2023-05-01
Enhancement: Advanced flow/deployment list filters


znicholasbrown:
Allow filtering by:
- schedule active/inactive
- work pool
- deployment name
- last run state

znicholasbrown:
Closed by https://github.com/PrefectHQ/prefect-ui-library/pull/1406

PR #1406, opened 2023-05-02
Enhancement: flow list filters

znicholasbrown:
This PR adds some more granular filters to the flow list component and moves most of the filter + template logic out of the flow list and into a new component.


src/components/StateNameSelect.vue
```diff
@@ -1,26 +1,24 @@
...
-      <template #default="{ value }">
-        <StateBadge
-          class="state-name-select__option"
-          :state="getStateFromTagValue(value)"
-          flat
-        />
-      </template>
-    </p-select>
-  </div>
+  <p-select v-model="internalValue" class="state-name-select" v-bind="{ options, multiple, emptyMessage }">
```
znicholasbrown, line 2: This looks like a lot of changes but I just removed the wrapper div to allow callers to pass attributes directly to the underlying `PSelect` component - they weren't being passed at all before

src/components/FlowList.vue
```diff
@@ -81,43 +107,46 @@
...
+  const nameLike = computed({
+    get() {
+      return searchDebounced.value === '' ? undefined : searchDebounced.value
+    },
+    set(value: string | undefined) {
+      searchDebounced.value = value
     },
   })
+
+  const page = useRouteQueryParam<number>('page', NumberRouteParam, 1)
```
znicholasbrown, line 124: Adding a "page" query param instead of including the `offset` in the synced filter query params solved a lot of complexity. In a future PR I'd like to abstract this into a wrapper component and composition that can solve for this more broadly.

src/components/FlowList.vue
```diff
@@ -81,43 +107,46 @@
...
-      page.value = Math.ceil(value / DEFAULT_LIMIT) + 1
+  const search = ref<string>()
+  const searchDebounced = useDebouncedRef(search, 800)
+  const nameLike = computed({
+    get() {
+      return searchDebounced.value === '' ? undefined : searchDebounced.value
+    },
+    set(value: string | undefined) {
+      searchDebounced.value = value
     },
```
znicholasbrown, line 121: This prevents the `nameLike` filter from being applied to the query params when it's empty - follow up from https://github.com/PrefectHQ/vue-compositions/pull/243

znicholasbrown: It also ensures that an empty `nameLike` param doesn't incorrectly cause `isCustomFilter` to be `true`

znicholasbrown:
Resolves: #1399

src/components/FlowList.vue
```diff
@@ -1,35 +1,59 @@
...
-          <div class="flow-list__header-end">
-            <SearchInput v-model="search" :placeholder="localization.info.searchByFlowName" :label="localization.info.searchByFlowName" />
-            <p-select v-model="filter.sort" :options="flowSortOptions" />
-            <p-tags-input v-model="filter.flowRuns.tags.name" :placeholder="localization.info.addTagPlaceholder" class="flow-list__flow-run-tags">
-              <template #empty-message>
-                <span class="flow-list__flow-run-tags--empty">{{ localization.info.filterByFlowRunTags }}</span>
+              <template v-else-if="selected.length">
+                <SelectedCount :count="selected.length" />
+                <FlowsDeleteButton size="xs" :selected="selected" @delete="deleteFlows" />
               </template>
```
pleek91, line 19: I'd like to see this piece abstracted into its own component. Decent bit of logic/template that could be separated out.

src/components/FlowsFilterGroup.vue
pleek91, line 1: This component looks nice 👍

pleek91 (approved):

src/components/FlowList.vue
znicholasbrown: Agreed - I don't have any (good) ideas for a more generic component but I could certainly split out the logic for this instance

PR #1407, opened 2023-05-03
BugFix: Invalid Schema Values mapping on create flow run forms

zhen0:
https://github.com/PrefectHQ/prefect-ui-library/pull/1356 resolved https://github.com/PrefectHQ/prefect/issues/9248 for the deployment form but the issue remained when creating a flow run from a deployment.  This PR moves the logic added in https://github.com/PrefectHQ/prefect-ui-library/pull/1356 to a new DeploymentParameters component and uses that component wherever we offer the option to edit parameters. 

Resolves https://github.com/PrefectHQ/prefect/issues/9248
Also resolves https://github.com/PrefectHQ/prefect/issues/9410

PR #1412, opened 2023-05-08
Bugfix: Fix use of model value in deployment parameters component

zhen0:
Uses the prop parameters instead of deployment.parameters to set the values in the deployment form and flow run create form.  Fixes a bug introduced by #1407 which ignored values passed by the "copy to new run" button.

### DIFF
--- a/src/components/DeploymentParameters.vue
+++ b/src/components/DeploymentParameters.vue
@@ -12,6 +12,7 @@
 
   const props = defineProps<{
     deployment: Deployment,
+    modelValue: SchemaValues,
   }>()
 
   const emits = defineEmits<{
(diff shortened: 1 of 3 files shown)


src/components/DeploymentParameters.vue
```diff
@@ -30,11 +31,13 @@
 
   const parameters = computed({
     get() {
-      const source = { values: props.deployment.parameters, schema: parameterOpenApiSchema.value }
+      const source = { values: props.modelValue, schema: parameterOpenApiSchema.value }
```
znicholasbrown, line 34: Should this be taking both deployment and model value into account?

zhen0: No. We're extracting parameters from the deployment at the form level so they're passed as part of the model value. 

The only place where there's a difference between the deployment parameters and the model value is in the custom run form where we need to consider the parameter value passed as props because of the "copy to new run" option. We do the handling of those in the form and then pass the updated parameters to this component.

znicholasbrown: Got it - thanks for the explanation!

PR #1416, opened 2023-05-10
BugFix:  Add override to logs empty text color for light mode 

zhen0:
In light mode the flow run logs empty text is not showing because elsewhere we use `p-empty-text` with a light background and here we have a dark code background.  This PR adds an override to make the empty text white like the main logs container text.  It also updates the handling of the empty text for pending states. 

Current:

image

With update:
image


src/components/FlowRunLogs.vue
```diff
@@ -11,14 +11,14 @@
...
               This run is scheduled and hasn't generated logs
             </div>
-            <div v-else-if="flowRun.stateType == 'running'">
+            <div v-else-if="!isTerminalStateType(flowRun.stateType)" class="flow-run-logs__empty-text">
               Waiting for logs...
             </div>
-            <div v-else>
-              This run didn't generate Logs
+            <div v-else class="flow-run-logs__empty-text">
+              This run didn't generate logs
```
pleek91, line 21: Can we just add the class directly to `p-empty-results` rather than on each individual message? I'm guessing the "No logs match your filter criteria" also need the white test.

zhen0: image

zhen0: Sorry adding photo so don't lose it. 

The p-empty-results has an inner div that seems to conflict with whatever I set on the p-empty-results in the log container. Adding to the internal divs gives better targeting.  But good call on "No logs match your.."

pleek91: Hmm we should fix that in prefect design. But until that can we wrap all the slot text in a single div and add the class to that? So we don't keep repeating the class

pleek91 (approved):

PR #1417, opened 2023-05-11
Feature: Storage layer for managing flow run state in the ui

pleek91:
# Description
This PR introduces a new concept/pattern into the ui called "storage". Currently our components interact directly with the api. Which means other than subscriptions caching requests there is very little control over the data itself. 

Adding a storage layer does two things

## Increased sharing of flow runs across components
By having all components get flow runs out of storage rather than returning the api response directly means that both filter and individual get calls can benefit from "newer data". Which means if a flow run is requested individually by a component but another component already has that flow run in a list. They can share the same flow run reference. 

This isn't currently implemented but we could also use this to reduce api requests by not even making the single get request if the flow run exists in the store already. 

## Provides a single source of truth
Having a single place where all flow runs are stored means at any point in time we can inspect that storage and make decisions based on what is there. 

## Why?
Having a single source of truth means that in OSS we can have a worker consistently checking for any flow runs that should be refreshed. And in cloud we can have a worker that automatically refreshes flow runs in storage based on events received from the system itself (observability).


src/services/storage/Storage.ts
```diff
@@ -0,0 +1,184 @@
...
+        return this.store.get(idRef.value)
+      }
+
+      return undefined
+    })
+
+
+    let unsubscribe: Unsubscribe | undefined = undefined
+
+    watch(idRef, (id) => {
```
brandonreid, line 90: Can't totally tell if this watcher get's destroyed when the `unsubscribe` is called?

src/compositions/useFlowRuns.ts
```diff
@@ -33,7 +34,13 @@ export function useFlowRuns(flowRunIds: string[] | Ref<string[] | null | undefin
   })
 
   const subscription = useSubscriptionWithDependencies(api.flowRuns.getFlowRuns, parameters)
-  const flowRuns = computed(() => subscription.response)
+  const flowRunsResponse = computed(() => subscription.response)
```
brandonreid, line 37: Just making sure I understand correctly.

So every component that calls for flow runs will still create a request, but these requests, if there are more than one at a time, get bundled together and duplicates get stripped at that point. Each request updates the storage, and if a new component causes a storage update then that would update all subscribers to the shared ref.

src/services/storage/Storage.ts
pleek91: I ended up removing this

src/compositions/useFlowRuns.ts
pleek91: This has been reworked a bit. But I'll still answer the question to make sure we're all on the same page. A couple of things are working together here. Subscriptions are still being used. So if multiple components ask for the same individual flow run a single request will be made. 

Here's what happens
1. subscription is created
2. response from subscription is added to storage
3. flow run from storage is returned to component

The response from those subscriptions will be added into storage. A watch is used to pick up the transition from `undefined` (api request hasn't completed yet) to `FlowRun` (api request completed). TECHNICALLY that same flow run can trigger `storage.add` to be called once per component. But they all have the same version of the flow run and there's not much of a cost of re-adding things to the storage. 

The storage class and the `useStorage` composition work together to automatically track which flow runs are being used by components and and removes flow runs from storage when they are no longer being used.

src/compositions/useFlowRuns.ts
```diff
@@ -1,39 +1,50 @@
...
+
+  const flowRuns = computed(() => {
+    if (subscription.response) {
+      const ids = subscription.response.map(flowRun => flowRun.id)
+
+      return storage.getAll(ids)
+    }
+
+    return undefined
+  })
```
pleek91, line 47: @znicholasbrown these are important implementation pieces. The flow run ids from the subscription are being used to get flow runs from storage. And `getAll` automatically calls a cleanup (unsubscribe) method each time you call it that cleans out previous values. So whenever the subscription changes, the storage will get cleaned up.

src/services/storage/StorageItem.ts
```diff
@@ -0,0 +1,5 @@
+export abstract class StorageItem {
+  public abstract readonly id: string
+
+  public readonly storageTimestamp = new Date()
```
pleek91, line 4: This is important to allow the storage class to determine if a value is newer. Basically, since all flow runs are initialized with this timestamp we can safely check the timestamp when adding a flow run to the storage that already exists to see if its "newer". If its not, we don't overwrite the already newer flow run in the store.

src/services/storage/useStorage.ts
```diff
@@ -0,0 +1,53 @@
...
+
+    const getAll = (ids: string[]): T[] => {
+      const unsubscribe = getAllUnsubscribe
+      const unsubscribes = ids.map<Unsubscribe>(id => storage.subscribe(id))
+
+      getAllUnsubscribe = () => unsubscribes.forEach(unsubscribe => unsubscribe())
+      unsubscribe?.()
+
+      return storage.getAll(ids)
+    }
```
pleek91, line 42: devx 🪄 

each time you call get it will automatically unsubscribe from the previous value. This is done in a safe way so if you call this multiple times with the same value its fine but if you call it with different values it will remove the previous value from storage if necessary.

src/services/storage/Storage.ts
```diff
@@ -0,0 +1,83 @@
+import { isDefined } from '@prefecthq/prefect-design'
+import { reactive } from 'vue'
+import { StorageItem } from '@/services/storage/StorageItem'
+
+type Store<T extends StorageItem> = Map<string, T | undefined>
```
stackoverfloweth, line 5: Totally expected this to be the return of pinia. Guess its still dead lol

pleek91: lol I was just thinking last week about how long it had been since I'd built a store based app. This is 100% ui business logic. The `reactive` map is the only thing that makes this vue 😄

znicholasbrown (commented):
Just a couple of comments @pleek91, otherwise I really like this implementation

src/components/FlowRunTimeline.vue
```diff
@@ -305,8 +305,21 @@
...
+    }
+
+    return {
+      flowRuns: {
+        id: ids,
+      },
+    }
+  })
+
+  const { flowRuns: subFlowRuns } = useFlowRuns(subFlowRunsFilter)
```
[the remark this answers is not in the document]
pleek91: The contract for useFlowRuns changed. It used to just take an array of ids. But this was the only place it was used. So in preparation for using compositions exclusively I updated that composition to accept a FlowRunsFilter. So needed to update this implementation.

src/compositions/useFlowRuns.ts
```diff
@@ -1,39 +1,50 @@
...
   })
 
   const subscription = useSubscriptionWithDependencies(api.flowRuns.getFlowRuns, parameters)
-  const flowRuns = computed(() => subscription.response)
+
+  const storage = useFlowRunStorage()
+
+  watch(() => subscription.response, response => {
+    if (response) {
+      flowRunStorage.addAll(response)
```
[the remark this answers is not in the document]
pleek91: Ah yes good catch. I think I missed this in both compositions. Will update

znicholasbrown (approved):

PR #1428, opened 2023-05-17
Enhancement: The return of the work queue health icon

zhen0:
image


billpalombi:
THANK YOU! This is absolutely better. I have some longer term ideas for improving this, but they are not inconsistent with this change and I don't want to let perfect be the enemy of the good.

src/components/FlowRunWorkQueue.vue
```diff
@@ -1,16 +1,24 @@
...
+  import { isTerminalStateType } from '@/models'
 
-  defineProps<{
+  const props = defineProps<{
     workQueueName: string,
     workPoolName?: string | null,
+    flowRunState?: string | null,
   }>()
+
+  const isNotTerminal = computed(() => props.flowRunState && !isTerminalStateType(props.flowRunState))
```
zhen0, line 21: I've added a check for state here because the health of the work queue is only really relevant on upcoming and in progress runs.  Showing an unhealthy work queue icon for a run that has successfully completed feels confusing to me.

src/components/WorkPoolQueueDetails.vue
```diff
@@ -6,6 +6,16 @@
...
     </p-key-value>
 
+    <p-key-value label="Status" :alternate="alternate">
+      <template #value>
+        <WorkQueueStatusBadge v-if="workPool" :work-queue="workPoolQueue" :work-pool="workPool" />
+      </template>
+    </p-key-value>
+
+    <template v-if="workQueueStatus">
+      <p-key-value label="Last Polled" :value="workQueueStatus.lastPolled ? formatDateTimeNumeric(workQueueStatus.lastPolled) : null" :alternate="alternate" />
```
znicholasbrown, line 16: Can you move this value logic out of the template? It'll be easier to understand there

src/components/WorkQueueDetails.vue
```diff
@@ -67,7 +68,22 @@
     return num > 0 ? `${num} Deployments` : 'Deployments'
   })
 
+  const api = useWorkspaceApi()
+
   const { workQueueStatus } = useWorkQueueStatus(props.workQueue.id)
+  const workPoolArgs = computed<Parameters<typeof api.workPools.getWorkPools> | null>(() => {
```
znicholasbrown, line 74:
```suggestion
  const workPoolArgs = computed<[WorkPoolsFilter] | null>(() => {
```
We should use the filter type here - it's more explicit and easier to read

WillRaphaelson (approved):

src/components/WorkQueueStatusIcon.vue
```diff
@@ -1,31 +1,33 @@
 <template>
```
znicholasbrown, line 1: Before we merge this I'd double check that this component isn't used elsewhere - as written this will no longer work for deprecated work queues (not sure that matters but worth a callout)

znicholasbrown: If this is being used elsewhere I'd revert these changes and create a new component specifically for work pool queues - if not, please rename this component to avoid confusion

znicholasbrown (commented):
Several components are changing responsibilities from displaying information about work queues => work pool queues. We had retained these components for deprecated work queues - if these are still referenced in an application the information will be misleading. If not, we should rename the components to reflect their new responsibilities to avoid confusion (I only commented this on one of the components but it applies to all)

src/components/WorkQueueStatusIcon.vue
zhen0: Ahhh... Good call out thank you.  Let me check!

src/components/WorkQueueStatusIcon.vue
```diff
@@ -1,31 +1,33 @@
...
   import { computed } from 'vue'
   import { useWorkQueueStatus, useWorkspaceApi } from '@/compositions'
 
   const props = defineProps<{
     workQueueName: string,
+    workPoolName: string,
   }>()
 
   const api = useWorkspaceApi()
+  const workPoolQueueArgs = computed<Parameters<typeof api.workPoolQueues.getWorkPoolQueueByName> | null>(() => [props.workPoolName, props.workQueueName])
```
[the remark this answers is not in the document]
pleek91: +1 to these. At one point I think I recommended using the Parameters utility type for these but I've also transitioned to using the actual types. Much easier to read and its a lot shorter.

src/components/WorkQueueStatusIcon.vue
zhen0: I'm not sure those components are actively being used but I've copied the updates to a new WorkPoolQueueStatusBadge and WorkPoolQueueStatusIcon components and reverted the WorkQueueStatusBadge, WorkQueueStatusIcon, WorkQueueDetails and WorkQueuesTable components.  This should give us cover if there are any old work queues out there and make it easier to remove those deprecated components when we think we can.

znicholasbrown: Thank you @zhen0!!

znicholasbrown (approved):

PR #1430, opened 2023-05-17
Feature: Deployment tags input

znicholasbrown:
Adds a new component for selecting deployment tags that simulates a tags lookup - adds it to the flows filter group used in flow lists


brandonreid (approved):

src/components/DeploymentTagsInput.vue
```diff
@@ -0,0 +1,43 @@
...
+
+  const filter = computed<DeploymentsFilter>(() => {
+    return {
+      ...props.filter,
+    }
+  })
+
+  const { deployments } = useDeployments(filter)
+  const tagList = computed(() => deployments.value?.flatMap(deployment => deployment.tags ?? []))
+  const options = computed(() => [...new Set(tagList.value)])
```
pleek91, line 42: there's a `unique` array utility for this.

src/components/DeploymentTagsInput.vue
```diff
@@ -0,0 +1,43 @@
...
+    set(value) {
+      emits('update:selected', value)
+    },
+  })
+
+  const filter = computed<DeploymentsFilter>(() => {
+    return {
+      ...props.filter,
+    }
+  })
```
pleek91, line 38: `useDeployments` accepts an undefined ref so rather than a computed for this I'd just do `const { filter } = toRefs(props)`. Little cleaner and more performant.

src/components/DeploymentTagsInput.vue
pleek91, line 42: I'd combine these into a single computed as well.

src/components/FlowsFilterGroup.vue
```diff
@@ -6,7 +6,12 @@
       </p-label>
 
       <p-label :label="localization.info.deploymentTags">
-        <p-tags-input v-model="filter.flowRuns.tags.name" :empty-message="localization.info.all" :placeholder="localization.info.addTagPlaceholder" class="flows-filter-group__deployment-tags" />
+        <DeploymentTagsInput
+          v-model:selected="filter.flowRuns.tags.name"
+          v-model="filter.flowRuns.tags.name"
```
pleek91, line 11: Why is there two v-models on this?

znicholasbrown: Ah bad copy-paste - let me update

src/components/DeploymentTagsInput.vue
znicholasbrown: Hm the overload for `useDeployments` isn't quite correct - but ` const { filter = {} } = toRefs<{ filter?: DeploymentsFilter }>(props)` works. I'm not gonna dig into that more tbh

znicholasbrown:
Addressed your comments @pleek91 - once you sign off on those I'll bring the changes to #1432 as well

pleek91 (approved):
LGTM!

PR #1432, opened 2023-05-17
Feature: Variable tags input

znicholasbrown:
Adds a new component to auto-suggest tags for variables; also adds composition for single and plural variable look ups


src/compositions/useVariables.ts
```diff
@@ -0,0 +1,54 @@
...
+  const variablesFilter = computed<[VariablesFilter] | null>(() => {
+    if (!can.read.variable) {
+      return null
+    }
+
+    if (isNullish(filterRef.value)) {
+      return [{}]
+    }
+
+    if (Array.isArray(filterRef.value)) {
```
pleek91, line 30: Nice. I like this. Going to have to use this for the `useFlowRuns` compositions.

pleek91 (approved):
LGTM

Issue #1443, opened 2023-05-23
Testing: Vitest config


znicholasbrown:
We've decided we want to use [Vitest](https://vitest.dev/) to handle unit tests in this project; this ticket is to set up vitest and add a simple set of CI jobs to make sure it runs on each PR. 

(Stretch) Explore exposing coverage for certain directories (i.e. schemas) alone and make coverage above a certain level a requirement for PRs that include that directory.

PR #1444, opened 2023-05-23
Chore: Vitetest setup

znicholasbrown:
This PR introduces vitestest to the project and updates our actions workflows to run the test suite on every PR. 

Resolves: #1443

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -37,3 +37,21 @@ jobs:
 
       - name: Build package
         run: npm run build
+
+  test:
+    name: Test
+    runs-on: ubuntu-latest
+
+    steps:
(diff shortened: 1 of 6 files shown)


.github/workflows/tests.yaml
```diff
@@ -32,8 +32,20 @@ jobs:
       - name: Setup
         uses: ./.github/actions/setup
 
-      - name: Install dependencies
```
znicholasbrown, line 35: Removed this step because dep installation happens in `Setup` as well so this was just burning compute

znicholasbrown:
Yeah I'm still working on this, just wanted a PR open so I could modify actions

znicholasbrown:
Ok this should be g2g @pleek91

PR #1445, opened 2023-05-23
Chore: update prefect-design@1.8.3

pleek91:
# Description
Update package and fix type errors


src/components/WorkQueuesTable.vue
```diff
@@ -115,19 +115,17 @@
   const workPoolsSubscription = useSubscription(api.workPools.getWorkPools, [], subscriptionOptions)
   const workPools = computed(() => workPoolsSubscription.response ?? [])
 
-  const workQueuesData = computed<TableData[]>(() => workQueues.value.map(queue => {
-    return {
-      ...queue,
-      disabled: workPools.value.some(pool => pool.defaultQueueId == queue.id),
-    }
-  }))
+  const workQueuesData = computed(() => workQueues.value.map(queue => new WorkQueueTableData({
```
znicholasbrown, line 118: I'm not thrilled with this solution but I don't want to hold up the PR - would you mind opening a ticket describing why this is necessary in prefect-design?

znicholasbrown (approved):

src/components/WorkQueuesTable.vue
pleek91: I'm honestly not sure what the ticket would even say tbh. And I'm honestly not unhappy with this implementation. It makes sense at the p-table level for each row to have a disabled property. The only thing I can think of is possibly changing to definition of the data prop in p-table to something like this. What do you think of this? 
```typescript
type TableRow<T extends TableData> = {
  disabled?: boolean,
  value: T,
}

// type for p-table's data prop. 
T | TableRow<T>
```

znicholasbrown: I'd be more on board with that - having to create a dedicated class to extend each of our models for use in a component feels like a poor devx, _especially_ if it's for a single property that the underlying component should be providing

pleek91: 100% agree. I'll open an issue for that. That'll also fix the type safety issue.

PR #1447, opened 2023-05-23
BugFix: Deployment parameter overrides not showing up when creating a custom run

pleek91:
# Description
Part of the fix for https://github.com/PrefectHQ/prefect/issues/9534

Adds deployment parameters to the url when creating a custom run from the deployment menu. Fixes the `DeploymentParameters` component implementation where certain values (like dates) would revert to default values because of being mapped multiple times.

### DIFF
--- a/src/components/DeploymentCustomRunOverflowMenuItem.vue
+++ b/src/components/DeploymentCustomRunOverflowMenuItem.vue
@@ -1,5 +1,5 @@
 <template>
-  <router-link :to="routes.deploymentFlowRunCreate(deploymentId)">
+  <router-link v-if="deployment" :to="routes.deploymentFlowRunCreate(deploymentId, deployment.rawParameters)">
     <p-overflow-menu-item>
       <slot>
         Custom run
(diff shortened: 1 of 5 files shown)


zhen0:
Before we merge let's QA this really carefully. I was seeing an issue with datetime params simply from editing the form (not a custom run) so want to make sure we've captured everything here.

pleek91:
> Before we merge let's QA this really carefully. I was seeing an issue with datetime params simply from editing the form (not a custom run) so want to make sure we've captured everything here. 

@zhen0 agreed. I was seeing the same issue with editing and creating runs.

znicholasbrown (approved):
I'm approving because I can't repro the behavior @zhen0 mentioned after some pretty extensive QA. That said, let's only merge if that can be confirmed separately so any repro steps/code would be helpful

pleek91:
@zhen0 maybe I can get your approval on this as well?

zhen0:
Thanks @pleek91 

I've run through several parameter versions to try to break this and can't.

zhen0 (approved):

PR #1448, opened 2023-05-24
BugFix: Unknown states crash flow run timeline

pleek91:
# Description
Closes https://github.com/PrefectHQ/prefect/issues/9689

- Adds "Skipped" state
- Fixes "Retrying" state
- Uses consistent colors for states on the timeline rather than hard coding everything to the `600` shade.


znicholasbrown (approved):
Ty @pleek91 !

billpalombi:
We need to remove the "Skipped" state - Prefect 2 does not a built in named skipped state.

PR #1449, opened 2023-05-24
BugFix: Fix `useStorage` composition destroying the storage context

pleek91:
# Description
I noticed some errors around storage on the flow run page. They were caused by returning the storage methods directly from the composition rather than wrapping them in arrow function to preserve context.


znicholasbrown (approved):
Nice. Lgtm

PR #1451, opened 2023-05-25
Enhancement: Organize and update task run details

pleek91:
# Description
Closes https://github.com/PrefectHQ/prefect/issues/8632

### DIFF
--- a/src/components/TaskRunDetails.vue
+++ b/src/components/TaskRunDetails.vue
@@ -6,11 +6,31 @@
       </template>
     </p-key-value>
 
+    <template v-if="taskRun.startTime">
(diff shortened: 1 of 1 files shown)


znicholasbrown (approved):
Approving in case it's intentional but it looks like this removes the Cache Key line

znicholasbrown (approved):
Approving in case it's intentional but it looks like this removes the Cache Key line

znicholasbrown (approved):
Approving in case it's intentional but it looks like this removes the Cache Key line

pleek91:
@znicholasbrown Cache Key is still visible it was just moved. TY!

PR #1463, opened 2023-05-31
Feature: DashboardTimeSpanFilter

pleek91:
# Description
Filter component for changing the time span that the dashboard shows. 
image


brandonreid (approved):

znicholasbrown (approved):

brandonreid:
Looking around actually, maybe the `h` and `w` should be lower case.
image
image

PR #1464, opened 2023-06-01
update work pool health indicator

brandonreid:
Updating these indicators in preparation for the new Dashboard view. The previous icons used for healthy and unhealthy appeared interactive/clickable (similar to a dismiss button or radio). This makes them more generic. The green dot is intended to be more passive than an illustrative icon. Distinction is maintained between each for accessibility (e.g. a red dot may not be visibly distinguishable from a green dot to certain users).

New healthy dark:
image

New healthy light:
image

New unhealthy dark:
image

New unhealthy light:
image

New paused dark:
image

New paused light:
image

### DIFF
--- a/src/components/WorkPoolQueueStatusIcon.vue
+++ b/src/components/WorkPoolQueueStatusIcon.vue
@@ -43,7 +43,7 @@
     icon: Icon,
   }>(() => {
     if (workPoolQueue.value?.isPaused) {
-      return { state: 'paused', name: 'Paused', icon: 'PauseIcon' }
+      return { state: 'paused', name: 'Paused', icon: 'PauseCircleIcon' }
     }
     if (healthy.value) {
       return { state: 'healthy', name: 'Healthy', icon: 'CheckCircleIcon' }
@@ -58,6 +58,7 @@
 .work-queue-status-icon { @apply
   flex
   items-center
+  cursor-help
 }
 
 .work-queue-status-icon--healthy { @apply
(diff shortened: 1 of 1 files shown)


pleek91:
On the dashboard we're going to have multiple of these side by side. Thoughts on making the healthy and others all the same size?

brandonreid:
> On the dashboard we're going to have multiple of these side by side. Thoughts on making the healthy and others all the same size?

Good call out. That was somewhat by design:
image

The `unhealthy` and `paused` states will be larger, but that will also allow them to stand out a bit more in the set and draw attention. Can always adjust if things get fugly.

src/components/WorkPoolQueueStatusIcon.vue
```diff
@@ -1,7 +1,18 @@
 <template>
-  <template v-if="workPoolQueue && workQueueStatus">
-    <p-icon :icon="status.icon" size="small" class="work-queue-status-icon" :class="classes" />
-  </template>
+  <p-tooltip
+    v-if="workPoolQueue && workQueueStatus"
+    class="work-queue-status-icon"
+    :text="`Work queue is ${status.state}`"
```
pleek91, line 5: On the dashboard it would be useful to include the name in this tooltip. Since there will be multiple work queue statuses grouped together. Maybe a prop to include the name or a slot or something to customize this message would be helpful.

pleek91:
@brandonreid the latest commit makes me wonder if we should just include the work pool queue name in the tooltip by default. We already have the name here and it would simplify this a decent bit and enforce some consistency.

brandonreid:
> @brandonreid the latest commit makes me wonder if we should just include the work pool queue name in the tooltip by default. We already have the name here and it would simplify this a decent bit and enforce some consistency.

@pleek91 that makes sense to me:

image

PR #1466, opened 2023-06-02
Enhancement: Add deployment filter before the dropdown

zhen0:
image

image

### DIFF
--- a/src/components/FlowList.vue
+++ b/src/components/FlowList.vue
@@ -21,6 +21,7 @@
 
             <div class="flow-list__header-end">
               <SearchInput v-model="search" :placeholder="localization.info.searchByFlowName" :label="localization.info.searchByFlowName" />
+              <SearchInput v-model="deploymentNameSearch" :placeholder="localization.info.searchByDeploymentName" :label="localization.info.searchByDeploymentName" />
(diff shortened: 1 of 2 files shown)


brandonreid:
When I search by a deployment in the [preview](https://deploy-preview-1466--prefect-ui-library.netlify.app/flows/flow-list?page=1&deployments.nameLike=dee), it filters the deployments in the expandable lists, but doesn't filter the list of flows. I think when I search, I expect the flows to filter out to those with the deployment as well, especially since the lists of deployments are collapsed by default. Maybe I'm the only one there?

zhen0:
Actually @brandonreid - I think it may just be the demo?

https://github.com/PrefectHQ/prefect-ui-library/assets/40272060/7e28e018-635a-4df8-b180-a36316a5ad09

znicholasbrown:
The demo app uses a bespoke filter implementation native to this library - when it comes to combined filters like this it'll sometimes give unexpected results because it's not doing an actual query, it's just filtering mocked data - for that sort of testing it's best to work against a live backend (local or otherwise)

znicholasbrown (approved):

brandonreid (approved):

PR #1467, opened 2023-06-02
Chore: Dashboard filter type and statistic component

pleek91:
# Description
Just a couple pieces to be used on the dashboard in this and other projects
image


src/components/DashboardStatistic.vue
```diff
@@ -0,0 +1,38 @@
...
+  gap-1
+  text-sm
+}
+
+.dashboard-statistic__value { @apply
+  font-semibold
+}
+
+.dashboard-statistic__meta { @apply
+  text-slate-500
```
znicholasbrown, line 36:
```suggestion
  text-foreground-200
```
This is usually better if you're looking for subdued text in all color modes

pleek91: @brandonreid ^

brandonreid: I’m definitely open minded with these details. The card boarder radiuses that I implemented were a good notch or two less than in the design files. Over all I’d advise sticking with conventions as much as possible or feeling free to use your best judgment. Design files don’t always translate to the flexible world of the browser.

When we have a fuller picture we can totally review as a team and find these happy mediums and sim to deploy them system wide. At some point, there’s certainly an initiative around arresting our design tokens from tailwind and making them semantic in a way where we don’t have to share personal mental models, but for now that’s the best we got and we should stick to it.


I’d go with @znicholasbrown’s suggestion for now. If it still seems off though, I can totally take on an audit.

znicholasbrown (approved):

PR #1468, opened 2023-06-05
BugFix: Use canceling state type instead of name

zhen0:
Uses the state type cancelling instead of cancelled for setting a flow run to cancelled. 

Quote from slack:

_We originally implemented cancellation as a Cancelled state with a Cancelling name but ran into a few edge cases and switched it to a new Cancelling state instead._


serinamarie:
The issue this PR stems from: 
https://github.com/PrefectHQ/prefect/issues/9801#issuecomment-1574369041

src/models/StateType.ts
```diff
@@ -5,6 +5,7 @@ export const stateType = [
   'pending',
   'failed',
   'cancelled',
+  'cancelling',
```
pleek91, line 8: I believe there is a utility function for mapping state names to state types that probably needs to be updated to reflect this addition.

PR #1472, opened 2023-06-06
Feature: Workspace dashboard task runs card

pleek91:
# Description
Adds a new dashboard component for task runs. 

Closes https://github.com/PrefectHQ/nebula-ui/issues/2855

image


src/components/WorkspaceDashboardTaskRunsCard.vue
```diff
@@ -0,0 +1,199 @@
...
+  h-16
+}
+
+.workspace-dashboard-task-runs-card__chart { @apply
+  absolute
+  min-h-0
+  -left-8
+  top-0
+  -right-8
+  bottom-0
```
brandonreid, line 173: I think we'll end up wanting some `bottom` amount, but we can do that when we're getting the actual layout together and see it next to the events card, etc.

brandonreid (commented):
I tried to keep up with the TaskRunHistory api types, but there's a lot that I don't have a great mental model for there.. Might be worth a review from @znicholasbrown  on that stuff?

We might see some tweaks to make around the line chart heights, edge cases. Looks good to me overall though.

src/components/WorkspaceDashboardTaskRunsCard.vue
pleek91: This bottom is relative to the container div which does respect the cards padding. So this is full width but still 16px from the bottom of the card.

brandonreid: Ah I see in the screenshot, ty

brandonreid (approved):

PR #1474, opened 2023-06-07
Feature/dashboard work pools card

brandonreid:
Adds the Work Pools card for use on the dashboard.

image


src/components/WorkPoolFlowRunCompletePercentage.vue
```diff
@@ -0,0 +1,159 @@
...
+  import { toPercent } from '@/utilities'
+
+  const props = defineProps<{
+    workPool: WorkPool,
+    filter?: WorkspaceDashboardFilter,
+  }>()
+
+  const subscriptionOptions = {
+    interval: 30000,
+  }
```
pleek91, line 30: at some point we should probably make a const for the interval the dashboard uses. I haven't even added intervals to the components I've made yet.

src/components/WorkPoolLastPolled.vue
```diff
@@ -0,0 +1,63 @@
...
+  const lastWorkerHeartbeat = computed(() => {
+    const heartbeats = workPoolWorkers.value.map(worker => worker.lastHeartbeatTime)
+
+    if (heartbeats.length === 0) {
+      return null
+    }
+
+    return max(heartbeats)
+  })
+  const lastPolled = computed(() => {
```
pleek91, line 50: Since the logic for this is copied from the WorkPoolCard we should make a composition for this. `useWorkPoolLastPolled` would be a nice abstraction that would dry up these components.

src/components/WorkPoolQueueStatusArray.vue
```diff
@@ -0,0 +1,41 @@
...
+  import { computed } from 'vue'
+  import { WorkPoolQueueStatusIcon } from '@/components'
+  import { useWorkspaceApi } from '@/compositions'
+  import { WorkPool } from '@/models'
+
+  const props = defineProps<{
+    workPool: WorkPool,
+  }>()
+
+  const maxWorkQueues = 26
```
pleek91, line 25: what's the thought here? I assume showing 26 work queues doesn't look great so why not allow 27 or 37? I vote we let this look bad to encourage better patterns.

brandonreid: So 25 would make 5 even rows of 5 in the actual implementation, this is worded poorly. I should make it 25 and use `<=`.

You can see from the percentiles Guidry pulled up:
image
Almost no one should encounter this, but for one workspace that has 454 work pools for some strange reason, hopefully this doesn't break is the real hope.. I could also bump this to 50 or something I suppose.

src/components/WorkPoolLastPolled.vue
brandonreid: I'll do that in a separate PR and use it here before I merge

src/components/WorkPoolFlowRunCompletePercentage.vue
brandonreid: I'll do this in a separate PR and use it here before I merge.

src/components/WorkPoolLastPolled.vue
brandonreid: Good call! Updated.

src/components/WorkPoolFlowRunCompletePercentage.vue
brandonreid: What's a good place to put something like that? Should I make a `useDashboardSubscriptionOptions` composition?

pleek91: Yes, that would give us a common place to do abstractions in the future

src/components/WorkPoolQueueStatusArray.vue
brandonreid: Added a limit to 50 and a fallback.

src/components/DashboardWorkPoolFlowRunCompletes.vue
```diff
@@ -0,0 +1,162 @@
...
+  const prevFlowRunFilter = computed(() => {
+    const rootStartTimeBefore = flowRunsFilter.value?.flowRuns?.startTimeBefore
+    const rootStartTimeAfter = flowRunsFilter.value?.flowRuns?.startTimeAfter
+
+    if (!rootStartTimeBefore || !rootStartTimeAfter || !props.filter?.timeSpanInSeconds) {
+      return null
+    }
+
+    const startTimeBefore = new Date(rootStartTimeBefore.getTime() - props.filter.timeSpanInSeconds * 1000)
+    const startTimeAfter = new Date(rootStartTimeAfter.getTime() - props.filter.timeSpanInSeconds * 1000)
```
pleek91, line 83: use date-fns `subSeconds` for this

src/components/DashboardWorkPoolFlowRunsTotal.vue
```diff
@@ -0,0 +1,43 @@
...
+  import { WorkspaceDashboardFilter } from '@/types'
+
+  const props = defineProps<{
+    workPool: WorkPool,
+    filter?: WorkspaceDashboardFilter,
+  }>()
+
+  const subscriptionOptions = {
+    interval: 30000,
+  }
```
pleek91, line 20: Can we use the composition you wrote? Or is that not ready to use

src/components/WorkPoolQueueStatusArray.vue
```diff
@@ -0,0 +1,52 @@
...
+
+  const maxWorkQueues = 50
+  const api = useWorkspaceApi()
+
+  const workPoolQueuesSubscription = useSubscription(
+    api.workPoolQueues.getWorkPoolQueues,
+    [props.workPool.name, { limit: maxWorkQueues }],
+  )
+  const workPoolQueues = computed(() => workPoolQueuesSubscription.response ?? [])
+  const showTooMany = computed(() => workPoolQueues.value.length >= maxWorkQueues)
```
pleek91, line 36: so since you used the limit on the request itself you can get rid of this because you'll never get more than 50 back from the api

src/compositions/useWorkPoolLastPolled.ts
```diff
@@ -0,0 +1,65 @@
...
+  options?: SubscriptionOptions,
+): UseWorkPoolLastPolled {
+  const api = useWorkspaceApi()
+  const { now } = useNow({ interval: 1000 })
+
+  const workPoolNameRef = ref(workPoolName)
+  const filterRef = ref(filter)
+
+  const finalFilters = computed<[string, WorkPoolWorkersFilter] | null>(() => {
+    if (!filterRef.value) {
```
pleek91, line 26: Don't you still need to return the workPoolName even if there are no extra filters?

pleek91 (approved):
Couple last comments but overall looks good

src/components/DashboardWorkPoolFlowRunsTotal.vue
brandonreid: I used that composition on the dashboard specific components, but left off these child components for now, since they could be used outside the dashboard. We talked about working out a composition inheritance type of utility, I figured we could tackle polling on this page more holistically in another PR. Open to pushback on that though.

PR #1486, opened 2023-06-14
Bugfix: Washed out flow run timeline view

znicholasbrown:
This PR fixes an issue where the timeline wasn't properly inverting colors it received from the `getStateTypeStyles` utility. It also passes a debounced theme ref to avoid a race condition where the theme has changed but the sample elements have not (yet).

Before:
Screenshot 2023-06-14 at 4 20 18 PM

After:
Screenshot 2023-06-14 at 4 20 23 PM


pleek91 (approved):
Nice, thanks for fixing this!

PR #1488, opened 2023-06-14
hide checkbox on disabled

brandonreid:
`disabled` was used to prevent the checkbox from animating in and being clickable. This change keeps respects that expectation, but actually removes a checkbox from view when it's disabled.

### DIFF
--- a/src/components/StateListItem.vue
+++ b/src/components/StateListItem.vue
@@ -55,7 +55,7 @@
   }>()
 
   const component = computed(() => {
-    if (!props.disabled && props.value !== undefined) {
+    if (!props.disabled || props.value !== undefined) {
       return PListItemInput
     }
 
--- a/src/components/StateListItem.vue
+++ b/src/components/StateListItem.vue
@@ -5,7 +5,7 @@
     :value="value"
     class="state-list-item"
     :class="classes"
-    disabled
+    :disabled="disabled"
   >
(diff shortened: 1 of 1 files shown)


src/components/StateListItem.vue
```diff
@@ -51,10 +51,11 @@
     value?: unknown,
     stateType?: StateType | null | undefined,
     tags?: string[] | null,
+    disabled?: boolean,
   }>()
 
   const component = computed(() => {
-    if (props.value !== undefined) {
+    if (!props.disabled && props.value !== undefined) {
```
pleek91, line 58: curious about the `value !=== undefined` piece. Do we need that if disabled controls whether or not the checkbox is shown?

brandonreid: Oh that should be an `||` danget, good eye.

pleek91: Do we even need the undefined check? Or is was that existing and we're just preserving that functionality.

brandonreid: Re in slack: updated again. Now it supports the empty array for selected as another way to disable it.. Not great, but fixes it for now. Really tough to fallow these props since they're forwarded as attrs from far up the chain in some places, so this is the safest bet for now.

pleek91 (approved):
Thanks for tackling this!

PR #1490, opened 2023-06-15
audit StateListItem selection disabling

brandonreid:
Previously, StateListItem had confusing methods for disabling selection. This PR makes selection an opt-in feature, `disable` will still hide the checkbox as well. Fallowing the paths here took a number of passes to make sure the logic made sense, some consuming components enable selection by default and have opt-out patterns.

This change will require only 1 change in the Nebula repo, none in OSS.

### DIFF
--- a/src/components/FlowRunFilteredList.vue
+++ b/src/components/FlowRunFilteredList.vue
@@ -11,7 +11,7 @@
       <StateNameSelect :selected="states" empty-message="All run states" class="flow-run-filtered-list__state-select" @update:selected="updateState" />
       <FlowRunsSort v-model="sort" class="flow-run-filtered-list__flow-runs-sort" />
     </div>
-    <FlowRunList v-model:selected="selectedFlowRuns" :flow-runs="flowRuns" :disabled="disabled || !can.delete.flow_run" @bottom="flowRunsSubscription.loadMore" />
+    <FlowRunList v-model:selected="selectedFlowRuns" :flow-runs="flowRuns" :disable-deletion="disableDeletion || !can.delete.flow_run" @bottom="flowRunsSubscription.loadMore" />
     <PEmptyResults v-if="empty">
       <template #message>
         <slot name="empty-message">
@@ -40,7 +40,7 @@
   const props = defineProps<{
     flowRunFilter: FlowRunsFilter,
     states?: PrefectStateNames[],
-    disabled?: boolean,
+    disableDeletion?: boolean,
   }>()
 
   const emit = defineEmits<{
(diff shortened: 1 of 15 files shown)


pleek91:
I like this direction. But the selectable and disable-deletion props are a little confusing (one is positive one is negative). Can we just use selectable? If you add selectable the input controls are enabled. 

Also disabled should probably disable the checkbox rather than remove it. Otherwise disabled and selectable both do the same thing and contradict if you have both props set to true.

brandonreid:
> I like this direction. But the selectable and disable-deletion props are a little confusing (one is positive one is negative). Can we just use selectable? If you add selectable the input controls are enabled.
> 
> Also disabled should probably disable the checkbox rather than remove it. Otherwise disabled and selectable both do the same thing and contradict if you have both props set to true.

Yeah I was thinking there could be a world where an item is selectable, but `disable` does like you said, disables the checkbox. In this component's implementation, it just so happens we hide the checkbox if it is disabled, but there may be a world where we decide to still show it in those cases (e.g. in some future scenario, users now have permissions to select some and not others). So I guess collapsing those into one prop felt like it was tailoring to the current styles rather than strict functionality or something. Thoughts?

pleek91:
I don't feel strongly about keeping the checkbox visible when it's disabled. But generally something disabled is still visible. 

The prop naming is the main part I'm thinking about.

brandonreid:
Really a discriminated union is sorta called for here, where `disabled` should only be an option if `selectable` is true, same with the other selection props. `selectable` is a variant prop, while `disabled` is specific to functionality.

When looking at the implementation in some places though, I think you're right.. It's causing issues. Some implementations set `selectable` to true by default, then potentially disable them all at once based on some condition, while the real intent is to show them as selectable or not.

brandonreid:
@pleek91 ok, only uses `selectable` now. Much easier to review too.

src/components/FlowRunFilteredList.vue
```diff
@@ -40,7 +40,7 @@
   const props = defineProps<{
     flowRunFilter: FlowRunsFilter,
     states?: PrefectStateNames[],
-    disabled?: boolean,
+    disableDeletion?: boolean,
```
pleek91, line 43: Can we use `selectable` here as well?

pleek91 (approved):
I didn't test this but the code looks good to me. Much better than what we had before. One last prop I think we should rename but approving preemptively.

src/components/FlowRunFilteredList.vue
brandonreid: Ya know what, I don't actually see an instance where this is being disabled anywhere. The `can` check in here is relied on. I'll just remove it.

PR #1494, opened 2023-06-15
Feature: `SchemaInput` component

znicholasbrown:
This PR introduces a component, `SchemaInput` that will subsume the JSON / form input mapping functionality for all schema-related forms. Callers pass a schema and v-model a `SchemaValues`, regardless of type (JSON vs form). The purpose is to obfuscate that mapping and allow us to use deployment parameters (and in particular the JSON input mode) more broadly without juggling that mapping and validation.

Some open questions: 

- (~blocking~ resolved) as written, the component will handle validation but won't block upstream form submission because it only emits a valid `SchemaValue`. ~This is obviously incorrect but I was struggling to figure out a way to handle this that was also a decently straightforward devex. Open to ideas but I don't want to merge this until that's solved.~ Since the component is using a `useValidation` instance, validation observers at the root can catch form submissions and disable submit buttons accordingly!

- (~not blocking~ resolved) - ~I've chosen to leave the form of input type choosing up to the caller instead of incorporating a button group into the component. The reason for this is that the `FlowRunCreate` and `QuickRunParametersModal` both (should) use slightly different options there, including a "Default" option that will show no form at all and will rely on already-set API or runtime defaults. ~ 

cc @bunchesofdonald


src/components/QuickRunParametersModal.vue
```diff
@@ -1,9 +1,7 @@
 <template>
   <p-modal v-model:showModal="internalShowModal" class="parameters-modal" title="Run Deployment">
     <p-form @submit="submit">
-      <p-content>
-        <DeploymentParameters v-model="parameters" :deployment="deployment" />
-      </p-content>
+      <SchemaInput v-model="parameters" :schema="deployment.parameterOpenApiSchema" />
```
pleek91, line 4: Should this be using the rawSchema?

src/compositions/useSyncedRecordString.ts
```diff
@@ -0,0 +1,56 @@
...
+    return false
+  }
+}
+
+/**
+ * The useSyncedRecordString composition takes a MaybeRef of SchemaValues
+ * and creates two reactive references: a string and an object.
+ * It then sets up watchers to sync changes between these two references.
+ * @param initialValues UseSyncedRecordValue
+ * @returns UseSyncedRecordString
```
znicholasbrown, line 26:
```suggestion
 * The useSyncedRecordString composition takes a record or record-parsable string
 * and creates two reactive references: a string and an object.
 * It then sets up watchers to sync changes between these two references.
 * @param initialValues UseSyncedRecordValue
 * @returns UseSyncedRecordString
```

src/components/SchemaInput.vue
pleek91, line 1: Am I correct that this component expects the schema to already be mapped? Thoughts on having it accept the unmapped "raw" schema? I'd like to get to the point where we don't map the schemas in the api layer and only map them when needed in components like this.

src/components/QuickRunParametersModal.vue
znicholasbrown: I don't think so, no - that looks like it was incorrect before since `DeploymentParameters` was stripping required properties when this form should be respecting them.

src/compositions/useOptionalPropertiesSchema.ts
```diff
@@ -0,0 +1,21 @@
+import { ComputedRef, MaybeRef, computed, ref } from 'vue'
+import { mapper } from '@/services'
+import { Schema, SchemaValues } from '@/types'
+
+export type UseOptionalPropertiesSchema = {
+  schema: ComputedRef<Schema>,
+}
+
+export function useOptionalPropertiesSchema(rawSchema: MaybeRef<SchemaValues>): UseOptionalPropertiesSchema {
```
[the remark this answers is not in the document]
znicholasbrown: Oh good catch, those types completely overlap 🤦🏻

src/components/SchemaInput.vue
```diff
@@ -0,0 +1,100 @@
...
+  })
+
+  const inputType = ref<SchemaInputType>(props.defaultInputType ?? 'json')
+  const setInputType = (value: unknown): void => {
+    if (isSchemaInputType(value)) {
+      inputType.value = value
+    }
+  }
+
+  const defaultValues = merge(getSchemaDefaultValues(props.schema), props.modelValue)
```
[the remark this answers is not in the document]
znicholasbrown: My thinking is that it never really made sense that we were doing that in the first place - components should never need to reach into the mapping layer like this. Ostensibly any `modelValue` passed to this component should already have come from a mapped source. The type for that values snippet used in `SchemaFormFieldsWithValues` is also incorrect, though overlapping, since it _says_ it's emitting type `SchemaValues` but is actually emitting `SchemaValuesRequest`.

src/components/SchemaInput.vue
znicholasbrown: I'm kind of on the other side of the fence - I don't think we should ever be using the raw schema unless necessary (like when we need to modify it in a way that would break the mapped version). Either way though I'd like us to be consistent and I don't think we're using the raw schema anywhere except the composition I'm proposing here

pleek91: I'm fine with that direction as long as we're consistent. My thought is mapping the schema is decently expensive and we're doing it when we fetch lists of deployments even when we never access/use the schema. Perhaps we can map it lazily in the model itself?

src/compositions/useOptionalPropertiesSchema.ts
```diff
@@ -0,0 +1,21 @@
...
+  const rawSchemaRef = ref(rawSchema)
+
+  const computedSchema = computed(() => {
+    rawSchemaRef.value.required = []
+
+    return mapper.map('SchemaResponse', rawSchemaRef.value, 'Schema')
```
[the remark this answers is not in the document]
znicholasbrown: Ah probably not - I thought the mapper was operating on that field for some reason. Removed that and updated the type 👍🏻

src/components/SchemaInput.vue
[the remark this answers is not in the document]
pleek91: Makes sense.

src/compositions/useSyncedRecordString.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  watch(recordRef, (newRecord) => {
+    if (isValidRecordString(stringRef.value)) {
+      const stringifiedStringRef = JSON.stringify(JSON.parse(stringRef.value))
+      const stringifiedRecord = JSON.stringify(newRecord)
+      if (stringifiedStringRef !== stringifiedRecord) {
+        stringRef.value = JSON.stringify(newRecord)
+      }
+    }
+  })
```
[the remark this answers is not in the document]
znicholasbrown: Yup that's exactly why it's doing that - if the string isn't parsable we don't have a way to compare them and we know they'll never be equal, so it only makes sense to compare fully-qualified strings. After that, we're left with differences in formatting, which for the sake of this composition shouldn't matter, which is what parsing + stringifying at the same time allows us to ignore

src/components/SchemaInput.vue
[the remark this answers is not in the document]
znicholasbrown: Eh well in theory it does but in practice not so much - I've added the mapper back in because otherwise I think this'll break any form that's passing unmapped parameters.

src/components/SchemaInput.vue
znicholasbrown: Yeah agreed 100% re: lazy eval

src/components/QuickRunParametersModal.vue
znicholasbrown: yeah actually on another look this was happening in a couple of places but should be fixed now - the only one of these forms that needs to bypass the existing schema is the DeploymentForm, since you can leave parameters unset

src/components/SchemaInput.vue
```diff
@@ -0,0 +1,100 @@
...
+  import { localization } from '@/localization'
+  import { getSchemaDefaultValues } from '@/services'
+  import { SchemaInputType, isSchemaInputType } from '@/types/schemaInput'
+  import { Schema, SchemaValues } from '@/types/schemas'
+  import { isJson, fieldRules } from '@/utilities'
+
+  const props = defineProps<{
+    modelValue: SchemaValues | null | undefined,
+    schema: Schema,
+    defaultInputType?: SchemaInputType,
```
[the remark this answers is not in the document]
znicholasbrown: Great suggestion - agreed this was a cleaner approach

src/compositions/useOptionalPropertiesSchema.ts
```diff
@@ -0,0 +1,20 @@
...
+  schema: ComputedRef<Schema>,
+}
+
+export function useOptionalPropertiesSchema(rawSchema: MaybeRef<SchemaResponse | Schema>): UseOptionalPropertiesSchema {
+  const rawSchemaRef = ref(rawSchema)
+
+  const computedSchema = computed(() => {
+    rawSchemaRef.value.required = []
+    return rawSchemaRef.value
+  })
```
pleek91, line 15: This should probably return a new schema rather than modifying the ref. Because if a ref is passed in then it'll modify the original ref which might be unexpected.

src/compositions/useSyncedRecordString.ts
```diff
@@ -0,0 +1,47 @@
...
+}
+
+/**
+ * The useSyncedRecordString composition takes a record or record-parsable string
+ * and creates two reactive references: a string and an object.
+ * It then sets up watchers to sync changes between these two references.
+ * @param initialValues UseSyncedRecordValue
+ * @returns UseSyncedRecordString
+ */
+export function useSyncedRecordString(initialValues?: UseSyncedRecordValue | string): UseSyncedRecordString {
```
pleek91, line 19: Thoughts on renaming this `useJsonRecord`? And having it accept/return 

- valid
- json
- record

src/utilities/json.ts
```diff
@@ -23,3 +25,12 @@ export function stringifyUnknownJson(value: unknown): string | null | undefined
...
 }
+
+export function isValidJsonString(value: unknown): value is string {
+  try {
+    const parsed = JSON.parse(value as string)
+    return isRecord(parsed)
+  } catch {
+    return false
+  }
+}
```
pleek91, line 36: I think this method is misleading. `'true'` is a valid json string but if passed to this method it would return false. I think `isValidJsonRecord` would be more accurate.

pleek91 (approved):
Approving as these last comments are not blocking

znicholasbrown:
Agreed on all counts @pleek91 - updated those here instead of waiting for a later PR

PR #1495, opened 2023-06-15
Feature: Workspace dashboard flow runs card

pleek91:
# Description
Adds the remaining functionality to the flow runs card component for the workspace dashboard. 

image


src/components/FlowRunStateTypeCount.vue
```diff
@@ -0,0 +1,66 @@
+<template>
+  <div class="flow-run-state-type-count">
+    <span class="flow-run-state-type-count__badge">
+      <template v-for="state in states" :key="state">
```
brandonreid, line 4: Since the red and orange of failed and cancelled is so close in color, they may be largely indistinguishable and may even appear like a mistake. We may take this out, but I do appreciate how clean it is in code. Think it's something we can revisit on a design pass once we've got the page laid out.

src/components/FlowRunStateTypeTabDescription.vue
```diff
@@ -0,0 +1,53 @@
...
+      },
+    },
+  }))
+
+  const states = computed(() => asArray(props.stateType))
+
+  const { count } = useFlowRunsCount(filter)
+  const hasCount = computed(() => count.value !== undefined)
+
+  const formatter = new Intl.ListFormat('en', { style: 'long', type: 'disjunction' })
```
brandonreid, line 37: man so many slick tricks in this repo!

src/components/FlowRunsAccordion.vue
```diff
@@ -0,0 +1,41 @@
...
+  import { Flow } from '@/models/Flow'
+  import { toMap } from '@/utilities'
+
+  const props = defineProps<{
+    filter: FlowRunsFilter,
+  }>()
+
+  const flowsFilter = computed<FlowsFilter>(() => {
+    // eslint-disable-next-line no-unused-vars
+    const { sort, limit, offset, ...filter } = props.filter
```
brandonreid, line 29: almost wonder if you want to handle this in some sort mapper or something, `FlowRunsFilterToFlowsFilter` idk.

brandonreid (dismissed):
A few comments, thoughts really. State buckets is the only thing we'll really want to address.

src/components/FlowRunStateTypeTabs.vue
```diff
@@ -0,0 +1,71 @@
...
+  import FlowRunStateTypeCount from '@/components/FlowRunStateTypeCount.vue'
+  import FlowRunStateTypeTabDescription from '@/components/FlowRunStateTypeTabDescription.vue'
+  import { FlowRunsFilter } from '@/models/Filters'
+  import { StateType } from '@/models/StateType'
+
+  const props = defineProps<{
+    filter?: FlowRunsFilter,
+  }>()
+
+  const tabStates: Record<string, StateType[]> = {
```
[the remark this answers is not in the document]
brandonreid: Ok, here's the state buckets we're hoping to achieve here for now:

Red
- failed
- crashed

Blue
- running
- pending
- paused

Green
- complete

Yellow
- scheduled
- cancelled

src/maps/filters.ts
```diff
@@ -140,7 +140,7 @@ export const mapTagFilter: MapFunction<TagFilter, TagFilterRequest> = function(s
 export const mapStateFilter: MapFunction<StateFilter, StateFilterRequest> = function(source) {
   return removeEmptyObjects({
     ...toOperator(source.operator),
-    type: toAny(source.type),
+    type: toAny(source.type?.map(type => type.toUpperCase())),
```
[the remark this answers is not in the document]
pleek91: The API expects state names to be uppercase. Unfortunate decision a long time ago (when state types were more important in the ui than they are now) we chose to map them to lowercase to use as labels and such in the ui. But that means we need to map them back to uppercase when we make api requests. Apparently we didn't have any filter request using state type cause this was just broken.

src/components/WorkspaceDashboardTaskRunsCard.vue
```diff
@@ -125,6 +113,19 @@
 
     return max * 4
   })
+
+  function getPercent(x: number | undefined, y: number | undefined): string | undefined {
```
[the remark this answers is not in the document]
pleek91: Agreed, will look for an opportunity to do so in future PRs.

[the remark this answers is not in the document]
pleek91: This was just a bugfix because this broke in some previous PR.

src/components/FlowRunStateTypeTabs.vue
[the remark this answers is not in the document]
pleek91: Updated 👍

src/components/FlowRunsAccordion.vue
pleek91: yeah this is kinda weird, I like the mapper idea but I also don't want to encourage doing this lol. So for now I'm going to leave it here.

src/components/FlowRunStateTypeCount.vue
pleek91: agreed, we can revisit 👍

src/components/FlowRunStateTypeCount.vue
```diff
@@ -0,0 +1,66 @@
...
+    },
+  }))
+
+  const states = computed(() => asArray(props.stateType))
+
+  const { count } = useFlowRunsCount(filter)
+
+  function getStateTypeClass(state: StateType): string {
+    return `bg-state-${state}-500`
+  }
```
znicholasbrown, line 41: Don't we have a utility for this?

src/components/FlowRunStateTypeTabs.vue
```diff
@@ -0,0 +1,71 @@
+<template>
+  <p-tabs v-model:selected="selected" :tabs="tabs" class="flow-run-state-type-tabs">
+    <template #heading="{ tab }">
+      <template v-if="tab">
```
znicholasbrown, line 4: Is this `v-if` check necessary? It doesn't seem like `tab` can be undefined from the type

src/components/FlowRunStateTypeTabs.vue
```diff
@@ -0,0 +1,71 @@
...
+
+  const tabStates: Record<string, StateType[]> = {
+    failed: ['failed', 'crashed'],
+    running: ['running', 'pending', 'paused'],
+    completed: ['completed'],
+    scheduled: ['scheduled', 'cancelled'],
+  }
+  const tabs = Object.keys(tabStates)
+
+  const selected = useRouteQueryParam('flow-run-state', 'failed')
```
znicholasbrown, line 37: ❤️

src/components/FlowRunsAccordion.vue
```diff
@@ -0,0 +1,41 @@
+<template>
+  <template v-if="flowIds">
+    <p-accordion :sections="flowIds" class="flow-runs-accordion">
+      <template #header="{ section: flowId, id, toggle, content, selected }">
+        <FlowRunsAccordionHeader :flow="getFlow(flowId)!" v-bind="{ id, content, toggle, filter, selected }" />
```
znicholasbrown, line 5: I think the need for a non-null assertion is a smell test that the flow lookup isn't quite correct - it doesn't seem like an in-memory lookup based on a single API request is likely to handle scale gracefully if the number of flows exceeds the default API limit

src/components/FlowRunsAccordion.vue
znicholasbrown, line 29: This is definitely a way to accomplish this but can't we instead `Omit` the fields we don't want callers to pass? That would give a much clearer expectation of what this component can and can't accept without needing to bypass the linter and create an intermediary computed

src/components/FlowRunsAccordion.vue
znicholasbrown: I might be misunderstanding what's going on here though

src/components/FlowRunsAccordionHeader.vue
```diff
@@ -0,0 +1,97 @@
...
+  import { useLastFlowRun } from '@/compositions/useLastFlowRun'
+  import { useWorkspaceApi } from '@/compositions/useWorkspaceApi'
+  import { useWorkspaceRoutes } from '@/compositions/useWorkspaceRoutes'
+  import { FlowRunsFilter } from '@/models/Filters'
+  import { Flow } from '@/models/Flow'
+  import { formatDateTimeRelative } from '@/utilities/dates'
+
+  const props = defineProps<{
+    flow: Flow,
+    id: string,
```
znicholasbrown, line 30: What's with the `id` prop and attr? I can't tell exactly what that id will be but if it's coming from the API it has a large chance of being an invalid CSS selector since selectors can't start with numbers.

znicholasbrown (commented):
A few comments but not blocking since this are a WIP

src/components/FlowRunsAccordion.vue
pleek91: Came up with a better solution for the type issue. As for the api scale, noted but if this component is trying to show more than the api limit this would be broken from a design perspective and we'd have to rethink that. So for now I'm going to leave this as is.

src/components/FlowRunsAccordion.vue
pleek91: You can Brandon commented on this which makes me second guess this approach. But basically this accordion groups flow runs by their flow. So this component accepts a `FlowRunsFilter` as a prop but then here it converts it into a `FlowsFilter` so that we can grab all the flows for the flow runs we need to display. 

So the caller can pass these things to restrict how many flow runs they want to display. But we need to omit them when getting flows. Maybe there is a clearer/better way to do so.

src/components/FlowRunsAccordionHeader.vue
pleek91: I think this is a mistake I made in PAccordion shining through naming wise. PAccordion has a slot for the section header and it binds the ids for the heading and content which need to be bound for accessibility. So this id is coming straight from PAccordion. However you are correct that since the accordion sections are uuids these might be problematic. Though since they are not being used for css I think this should be fine

https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id

src/components/FlowRunStateTypeTabs.vue
pleek91: Something about the slot types for the heading slot makes tab possibly undefined. Briefly looked into it but wasn't 5 minute fix so decided using a v-if as a type guard was the simple fix. More growing pains with slot types appearing that previously were not available.

src/components/FlowRunStateTypeCount.vue
pleek91: I made that utility that gets the background and text color based on the `state--${stateType}` classes however I realized later (in that timeline light mode bug you fixed) that the background is different in light/dark mode. Which isn't really what we want here. This is something I'd like to revisit as a whole (state colors) but I know @brandonreid also wants to revisit them from a ux perspective. So I'm thinking we can tackle the devx and ux at the same time.

znicholasbrown: Makes sense, thanks 👍🏻

src/components/FlowRunStateTypeTabs.vue
znicholasbrown: 😞 🤠

src/components/FlowRunsAccordion.vue
znicholasbrown: Since the `flowId` exists on every flow run, could we do something like: `new Set(flowRuns.map(run => run.flowId)` and make the header responsible for fetching that flow with `useFlow(id)`? Then you could pass the filter through without modifying it since it's still just for flow runs

pleek91: Hmm that's clever. But then I'd have to fetch all the flow runs, which I'm currently not doing. Only the expanded flow fetches its flow runs.

src/components/FlowRunsAccordionHeader.vue
znicholasbrown: Hmmm ok - I don't have another solution outside of changing that prop so let's address later. I think Vue uses entirely data attributes when juggling selectors so i think we're probably in the clear since we do very little of that

brandonreid (approved):

PR #1496, opened 2023-06-16
Enhancement: Add all flow runs to storage using flow run compositions

pleek91:
# Description

## This is a breaking change
I will need to prepare PRs for oss and cloud to reflect this change in order to merge and publish. 

This PR introduces a change to how we fetch flow runs via the api to make sure all flow runs are added to storage. **When reviewing start with these files**
- `useFlowRun` already added flow runs to storage
- `useFlowRuns` now adds flow runs to storage and supports a `page` argument
- `infiniteScrollCompositionFactory` introduced to create entity specific compositions. This utility replaces `usePaginatedSubscription`. This adds the needed flexibility for separate compositions with separate business logic for storage
- `useFlowRunsInfiniteScroll` - replacement for anywhere we were using `usePaginatedSubscription` for flow runs.

This PR updates all implementations of of using `api.flowRuns.getFlowRuns` in components and other compositions to use the new `useFlowRuns` composition. This ensures that all flow runs are added to storage. Which is a major milestone in the goal of being able to separate out our refresh logic for flow runs. Which would allow us to stop polling and do more targeted refreshing.


src/components/FlowRunMenu.vue
```diff
@@ -53,12 +53,12 @@
 
 <script lang="ts" setup>
   import { showToast } from '@prefecthq/prefect-design'
-  import { useSubscription, useSubscriptionWithDependencies } from '@prefecthq/vue-compositions'
-  import { computed, ref } from 'vue'
+  import { useSubscription } from '@prefecthq/vue-compositions'
```
znicholasbrown, line 56:
```suggestion
```

src/components/FlowRunMenu.vue
```diff
@@ -97,20 +97,17 @@
     return isPausedStateType(flowRun.value.stateType)
   })
 
-  const flowRunFilter = computed<Parameters<typeof api.flowRuns.getFlowRuns> | null>(() => {
+  const flowRunFilter = computed<FlowRunsFilter | null>(() => {
```
znicholasbrown, line 100: ❤️

src/compositions/useFlowRuns.ts
```diff
@@ -26,7 +32,16 @@ export function useFlowRuns(filter: FlowRunsFilter | Ref<FlowRunsFilter | null |
...
 
-    return [filterRef.value]
+    const limit = filterRef.value.limit ?? FLOW_RUN_LIMIT
+    const offset = (pageRef.value - 1) * limit
+
+    const filter: FlowRunsFilter = {
+      ...filterRef.value,
+      limit,
+      offset,
+    }
```
znicholasbrown, line 42: Now that I see this on (paper) I'm realizing this could probably be even further abstracted since it's not specific to the flow runs route - maybe something like:
```ts
const GLOBAL_API_LIMIT = 200

function useFilterPagination(page: MaybeRef<number> = 1, limit: MaybeRef<number> = GLOBAL_API_LIMIT) {
    const pageRef = ref(page)
    const limitRef = ref(limit)
    const offset = computed(() => (pageRef.value - 1) * limit.value)

    return {
      limit,
      offset,
    }
}

export function useFlowRuns(filter: FlowRunsFilter | Ref<FlowRunsFilter | null | undefined>, page: MaybeRef<number> = 1, options?: SubscriptionOptions): UseFlowRuns {
  const { limit, offset } = useFilterPagination(page, filterRef.value.limit)
}

```

It's a small change to how you're juggling the filter (you might need to create a ref for it?) but it'll let you re-use the logic elsewhere, since the API limits are only applied globally anyway

src/compositions/useSubscriptions.ts
znicholasbrown, line 1: Love this.

src/services/storage/flowRunStorage.ts
```diff
@@ -4,4 +4,6 @@ import { useStorage } from '@/services/storage/useStorage'
 
 export const flowRunStorage = new Storage<FlowRun>()
 
-export const useFlowRunStorage = useStorage(flowRunStorage)
\ No newline at end of file
+export const useFlowRunStorage = useStorage(flowRunStorage)
+
+console.log(flowRunStorage)
```
znicholasbrown, line 9:
```suggestion
```

src/utilities/infiniteScroll.ts
```diff
@@ -0,0 +1,88 @@
...
+
+    if (getCurrentInstance()) {
+      onUnmounted(() => {
+        subscriptions.unsubscribe()
+        unwatch()
+      })
+    }
+
+    // Typescript says this is `[x: string]: ComputedRef<Awaited<ReturnType<TAction>>>>` even though `property` is TProperty
+    // so this object is unnecessary from a logic perspective but allows me to use an as without casting the entire return object
```
znicholasbrown, line 75: 😢 🤠

znicholasbrown (commented):
Most of this we discussed synchronously but overall I'm super on board with this, I think it'll improve pagination consistency immensely. Just a few comments/suggestions, otherwise this can be considered provisional approval.

src/utilities/reactivity.ts
```diff
@@ -42,4 +42,20 @@ export function uniqueValueWatcher(...[source, callback, options]: Parameters<ty
...
+
+  if (isArray(source)) {
+    return source.filter(value => isValidWatchValue(value))
+  }
+
+  return []
+}
+
+function isValidWatchValue(value: unknown): value is WatchSource | WatchEffect {
+  return isRef(value) || isReactive(value) || isFunction(value)
```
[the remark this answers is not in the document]
znicholasbrown: that's fine for now, we can export later

src/compositions/useFlowRuns.ts
pleek91: Great suggestion. Updated to use this pattern.

znicholasbrown (approved):

znicholasbrown:
I've approved both downstream PRs @pleek91 - feel free to merge this whenever you see fit.

PR #1501, opened 2023-06-20
BugFix: `TypeError: crypto.randomUUID is not a function` causes page to break

### LINKED ISSUE #1497: Enhancement: Shimmed crypto module
The `crypto` module isn't available in untrusted contexts. We rely on that module for ids in several of our graphing libraries, which causes any downstream apps to break when running on a non-localhost, non-https domain (like a VM). Before support for the module was widespread we used a shim to ensure it existed - we should reintroduce that for these situations.

pleek91:
# Description
The storage service used `crypto.randomUUID` to generate random ids for checkouts. But that isn't safe to use because its only available in secure contexts. Using the `randomId` utility from prefect-design make sure this works no matter what. 

Closes https://github.com/PrefectHQ/prefect-ui-library/issues/1497

### DIFF
--- a/src/services/storage/Storage.ts
+++ b/src/services/storage/Storage.ts
@@ -1,4 +1,4 @@
-import { isDefined } from '@prefecthq/prefect-design'
+import { isDefined, randomId } from '@prefecthq/prefect-design'
 import { reactive } from 'vue'
 import { StorageItem } from '@/services/storage/StorageItem'
 
@@ -39,7 +39,7 @@ export class Storage<T extends StorageItem> {
   }
 
   public subscribe(id: string): Unsubscribe {
-    const subscriptionId = crypto.randomUUID()
+    const subscriptionId = randomId()
     const subscriptions = this.getSubscription(id)
 
     subscriptions.add(subscriptionId)


znicholasbrown (approved):

tumluliu:
Hi there, I'm using 2.10.17 prefect but still have the `TypeError: crypto.randomUUID is not a function` issue. Is this fix really included in the 2.10.17 version? I saw it from the [release note](https://github.com/PrefectHQ/prefect/blob/2.10.17/RELEASE-NOTES.md#fixes)

phill84:
Actually opening prefect in an insecure context (non-localhost via plain http) still doesn't work in release 2.10.17 which has this fix. Is prefect UI supposed to work in an insecure context? This used to work in prefect<-2.10.13.

pleek91:
@tumluliu and @phill84 you are correct. There was a downstream (but still prefect) occurrence that I had missed. We were able to reproduce and found https://github.com/PrefectHQ/prefect-design/pull/829 which should take care of this. Thanks for reporting and helping debug. We'll get this fix out with the next release on thursday

PR #1507, opened 2023-06-23
BugFix: Filter task runs on the workspace dashboard by flow run tags rather than task run tags

pleek91:
# Description
Fixes the task runs card not showing the data you'd expect when filtering by tags.


znicholasbrown (approved):
Approving but you'll probably need a rebase once #1506 is merged

PR #1510, opened 2023-06-26
BugFix: Workspace dashboard task runs not being calculated correctly

pleek91:
# Description
Fixes some logical and filtering errors with the task runs card
- Accounts for large count differences between the completed chart and the failed or crashed chart
- Fix the filtering to look for task by state rather than flow runs
- Fix merge modifying input values

Before
image

After
image

### DIFF
--- a/src/components/WorkspaceDashboardTaskRunsCard.vue
+++ b/src/components/WorkspaceDashboardTaskRunsCard.vue
@@ -122,6 +122,8 @@
     return max
   })
 
+  const MAX_ITERATIONS = 10
+
   const maxCompletedValue = computed(() => {
     const completedValues = taskRunsChartData.value.completed.map(([, y]) => y)
     const minValue = 1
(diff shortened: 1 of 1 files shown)


src/components/WorkspaceDashboardTaskRunsCard.vue
```diff
@@ -113,15 +113,57 @@
...
+    const minValue = 1
+    const maxFailed = Math.max(...failedValues, minValue)
+
+    if (maxFailed === maxValue.value) {
+      return maxFailed
+    }
+
+    let unit = 1
+
+    while (unit <= 10) {
```
znicholasbrown, line 158: Why 10? (here and in the completed computed)

znicholasbrown (approved):
1 question, non-blocking

src/components/WorkspaceDashboardTaskRunsCard.vue
pleek91: Just a nice round number lol. I should make that a const. These basically just reduce the distance between the max completed value and the max failed value so you get at least some movement in the lines. Otherwise a large complete number completely flattens a tiny failed number.

PR #1515, opened 2023-06-27
Enhancement: Load more flow runs asyn when "show more" is press on the flow runs accordion

pleek91:
# Description
Rather than navigating to the flows page load more runs directly into the accordion's content.


znicholasbrown (approved):
Oh nice, I was hoping that was where this was headed

PR #1519, opened 2023-06-28
Chore: Remove skipped state type

pleek91:
# Description
Remove the skipped state name from the ui since it is not a name prefect itself uses


brandonreid (approved):

pleek91:
> Hold on - won't this break the timeline since these'll fall through to "Unknown"?

If a user creates a custom state name called "Skipped" it will get the same treatment as any other custom state name that we are unable to associate with a state type. There was a previous issue where an unknown state name would break the timeline. That was fixed by letting the timeline use the "unknown" styles  but this state name was also mistakenly added.

znicholasbrown (approved):
Got it - thanks 👍🏻

PR #1521, opened 2023-06-28
Bugfix: `Any` and `Object` default mapping

znicholasbrown:
This PR changes the way defaults are handled for `any` and `object` properties; previously unmapped defaults were returned when a nullish value was passed to the service. Now, defaults are mapped the same as other response values

### DIFF
--- a/src/services/schemas/utilities.ts
+++ b/src/services/schemas/utilities.ts
@@ -52,8 +52,8 @@ export function getSchemaRequestValue(schema: Schema, values: SchemaValues): Sch
 /*
  * Gets a UI friendly version of a property's api friendly value. Used for mapping.
  */
-export function getSchemaPropertyResponseValue(property: SchemaProperty, value: SchemaValue, level?: number): SchemaValue {
-  const service = schemaPropertyServiceFactory(property, level ?? 0)
+export function getSchemaPropertyResponseValue(property: SchemaProperty, value: SchemaValue, level: number = 0): SchemaValue {
+  const service = schemaPropertyServiceFactory(property, level)
 
   return service.mapResponseValue(value)
 }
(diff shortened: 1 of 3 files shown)


src/services/schemas/utilities.ts
```diff
@@ -52,17 +52,17 @@ export function getSchemaRequestValue(schema: Schema, values: SchemaValues): Sch
 /*
  * Gets a UI friendly version of a property's api friendly value. Used for mapping.
  */
-export function getSchemaPropertyResponseValue(property: SchemaProperty, value: SchemaValue): SchemaValue {
-  const service = schemaPropertyServiceFactory(property, 0)
+export function getSchemaPropertyResponseValue(property: SchemaProperty, value: SchemaValue, level?: number): SchemaValue {
+  const service = schemaPropertyServiceFactory(property, level ?? 0)
```
pleek91, line 56:
```suggestion
export function getSchemaPropertyResponseValue(property: SchemaProperty, value: SchemaValue, level: number = 0): SchemaValue {
  const service = schemaPropertyServiceFactory(property, level)
```

src/services/schemas/utilities.ts
```diff
@@ -52,17 +52,17 @@ export function getSchemaRequestValue(schema: Schema, values: SchemaValues): Sch
...
   return service.mapResponseValue(value)
 }
 
 /*
  * Gets a api friendly version of a property's UI friendly value. Used for mapping.
  */
-export function getSchemaPropertyRequestValue(property: SchemaProperty, value: SchemaValue): SchemaValue {
-  const service = schemaPropertyServiceFactory(property, 0)
+export function getSchemaPropertyRequestValue(property: SchemaProperty, value: SchemaValue, level?: number): SchemaValue {
+  const service = schemaPropertyServiceFactory(property, level ?? 0)
```
pleek91, line 65:
```suggestion
export function getSchemaPropertyRequestValue(property: SchemaProperty, value: SchemaValue, level: number = 0): SchemaValue {
  const service = schemaPropertyServiceFactory(property, level)
```

pleek91 (commented):
Syntax suggestion for the utility changes

pleek91 (approved):

src/services/schemas/utilities.ts
znicholasbrown: For some reason I thought TS didn't respect default values 😮‍💨  nice suggestion

PR #1522, opened 2023-06-29
Bugfix: SchemaInput reactivity

znicholasbrown:
I'm not totally sure this will work as intended - it works locally but so does the current version. I'll need to publish this to npm and try that against a deployed version of the application.

### DIFF
--- a/src/components/SchemaInput.vue
+++ b/src/components/SchemaInput.vue
@@ -37,7 +37,7 @@
   import { merge } from 'lodash'
   import { computed, ref, watchEffect } from 'vue'
   import { SchemaFormFields } from '@/components'
-  import { useForm, useJsonRecord } from '@/compositions'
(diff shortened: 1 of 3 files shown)


src/components/SchemaInput.vue
```diff
@@ -102,7 +111,7 @@
     if (inputType.value === 'json') {
       emit('update:modelValue', mapper.map('SchemaValuesResponse', { values: record.value, schema: props.schema }, 'SchemaValues'))
```
znicholasbrown, line 112: Since these are linked, we could probably avoid this check altogether and emit the model value update in the computed setter if we want

pleek91: +1 to just using the setter for this

pleek91 (approved):
Approving so we can test this in a released version

PR #1527, opened 2023-06-30
BugFix: Fix SchemaInput reactivity and remove vee-validate from FlowRunCreateForm.vue

pleek91:
# Description
Fixes json and reactivity issues with SchemaInput where default values could not be overridden. Also fixes `useReactiveForm` which was no longer syncing values back to the form when the source ref was updated.


znicholasbrown (approved):

znicholasbrown:
Going to approve/release this so we can live test

PR #1529, opened 2023-06-30
expand first flow accordion item by default on dashboard

brandonreid:
Expands the first accordion item in the list of flows on the Flow Runs card by default.

### DIFF
--- a/src/components/FlowRunsAccordion.vue
+++ b/src/components/FlowRunsAccordion.vue
@@ -1,6 +1,6 @@
 <template>
   <template v-if="flowIds">
-    <p-accordion :sections="flowIds" class="flow-runs-accordion">
+    <p-accordion v-model:selected="accordionModel" :sections="flowIds" class="flow-runs-accordion">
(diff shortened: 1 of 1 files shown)


src/components/FlowRunsAccordion.vue
```diff
@@ -35,6 +35,25 @@
...
+    get() {
+      if (selectedAccordionItem.value) {
+        return selectedAccordionItem.value
+      }
+
+      return null
+    },
+    set(selected) {
+      selectedAccordionItem.value = selected
+    },
```
pleek91, line 54: it doesn't seem like this computed is needed? I think you could v-model `selectedAccordionItem` and everything would work the same

brandonreid: Ahh. Indeed. haha, updating

pleek91 (approved):

brandonreid:
Going to create a new PR for this. Somehow my latest commit appears locally but isn't registering on the site.

PR #1532, opened 2023-07-03
Bugfix: Recursive work pool variables

znicholasbrown:
This PR reverts some of the changes to default logic for the any and object schema services, placing this logic instead in the root property service response value fallback. 

Previously we were attempting to map the default value whenever it was being accessed, which was incorrect because the response service logic was _also_ hitting the default as a fallback, which resulted in infinite recursion in situations where the default value was null or undefined. 

This also fixes an issue where defaults, used as values for worker variables, weren't using the mapped schema when retrieving default values. This meant that form values that relied on referenced definitions (blocks in particular) were never properly populated. 

Resolves: #2984

### DIFF
--- a/src/services/schemas/properties/SchemaPropertyAny.ts
+++ b/src/services/schemas/properties/SchemaPropertyAny.ts
@@ -34,13 +34,7 @@ export class SchemaPropertyAny extends SchemaPropertyService {
 
   protected request(value: SchemaValue): unknown {
     if (this.has('anyOf') || this.has('allOf')) {
-      const reference = this.referenceRequest(value)
-
-      if (reference === undefined) {
-        return undefined
-      }
-
-      return reference
+      return this.referenceRequest(value)
     }
 
     return parseUnknownJson(value)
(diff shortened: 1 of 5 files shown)


src/services/schemas/properties/SchemaPropertyAny.ts
```diff
@@ -8,19 +8,19 @@ import { parseUnknownJson, stringifyUnknownJson } from '@/utilities/json'
...
-      return ''
-    }
-
-    if (this.has('anyOf') || this.has('allOf')) {
-      return this.getDefaultValueForFirstDefinition()
+      return stringifyUnknownJson(defaultValue) ?? ''
     }
 
-    return null
+    return defaultValue ?? {}
```
pleek91, line 23: Changing the fallback value (if there is no default value) from `null` to `{}` seems risky to me. Is returning an empty object specifically needed?

znicholasbrown: Hm shouldn't be, no - reverted

pleek91 (approved):

PR #1539, opened 2023-07-05
use useInterval on work pools card

brandonreid:


src/components/WorkPoolQueueStatusArray.vue
```diff
@@ -28,9 +28,10 @@
   const maxWorkQueues = 50
   const api = useWorkspaceApi()
 
+  const options = useInterval({ interval: 30000 })
   const workPoolQueuesSubscription = useSubscription(
     api.workPoolQueues.getWorkPoolQueues,
-    [props.workPool.name, { limit: maxWorkQueues + 1 }],
+    [props.workPool.name, { ...options, limit: maxWorkQueues + 1 }],
```
pleek91, line 34: I don't think this is the correct placement for the options. In fact, I don't think the limit piece is correct either? 
image

pleek91 (approved):

PR #1540, opened 2023-07-05
Feature/average lateness to workpools

brandonreid:
Screenshot 2023-07-05 at 12 06 02 PM

### DIFF
--- a/src/components/WorkPoolAverageLateTime.vue
+++ b/src/components/WorkPoolAverageLateTime.vue
@@ -1,7 +1,7 @@
 <template>
   <span class="work-pool-average-late-time" :class="classes">
     <template v-if="averageLateness">
-      {{ averageLateness }}
+      {{ secondsToApproximateString(averageLateness) }}
     </template>
     <template v-else>
       N/A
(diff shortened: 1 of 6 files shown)


src/services/WorkspaceFlowRunsApi.ts
```diff
@@ -64,6 +65,13 @@ export class WorkspaceFlowRunsApi extends WorkspaceApi implements IWorkspaceFlow
     return mapper.map('FlowRunHistoryResponse', data, 'RunHistory')
   }
 
+  public async getFlowRunsAverageLateness(filter: FlowRunsFilter): Promise<string | null> {
+    const request = mapper.map('FlowRunsFilter', filter, 'FlowRunsFilterRequest')
+    const { data } = await this.post<number>('/lateness', request)
+
+    return data > 0 ? secondsToApproximateString(data) : null
```
pleek91, line 72: I would suggest this endpoint not return a formatted string. Doing so removes the possibility of doing any sort of math or calculation (or different formatting). Implementation should be responsible for formatting the seconds of lateness into a readable string.

src/components/WorkPoolAverageLateTime.vue
```diff
@@ -0,0 +1,58 @@
...
+        name: ['Late'],
+      },
+    },
+  }))
+
+  const averageLatenessSubscription = useSubscription(api.flowRuns.getFlowRunsAverageLateness, [lateFlowRunsFilter], subscriptionOptions)
+  const averageLateness = computed(() => averageLatenessSubscription.response
+    ? secondsToApproximateString(averageLatenessSubscription.response)
+    : null,
+  )
```
pleek91, line 47: nit: You don't really need to format this here and use a ternary. You can just return the response directly and in the template where you're already checking if there is lateness you can format the seconds like `{{ secondsToApproximateString(averageLateness) }}`

pleek91 (approved):

src/components/WorkPoolAverageLateTime.vue
brandonreid: Ahh true.. might as well cleanup the extra ternary

PR #1541, opened 2023-07-05
Feature: subscription interval provide/inject using `useInterval`

pleek91:
# Description
Adds a new injection key and composition for injecting the interval for a subscription. This means at the component level we can provide default polling per implementation without having to prop drill. This is specifically being implemented for the workspace dashboard were we want the whole page to poll at a specific interval.


brandonreid (approved):

pleek91:
@znicholasbrown I'd also like your eyes on this from a pattern perspective. Might be some risk you see that I didn't think of.

znicholasbrown (approved):
I think this is fine - the only issue I forsee is the lack of granularity; any subscription that relies on the injection necessarily has the same interval, which fits the dashboard use case but isn't particularly flexible elsewhere (or even on the dashboard if groups of components needed to refresh at different intervals but still want to avoid the drilling piece)

pleek91:
@znicholasbrown my thought being generally we want all subscriptions within a card/list item/page to interval at the same rate. So this provides a way to set that interval at any level. I think this still allows for a component to have a dedicated interval if needed but generally we haven't needed that.

znicholasbrown:
Hm ok yeah generally agreed and this is a marked improvement regardless 👍🏻

PR #1543, opened 2023-07-05
adding is_push_pool to api

WillRaphaelson:


src/models/api/WorkPoolResponse.ts
```diff
@@ -14,6 +14,7 @@ export type WorkPoolResponse = {
   description: string | null,
   type: string,
   is_paused: boolean | null,
+  is_push_pool: boolean | null,
```
pleek91, line 17: Nit: this doesn't change anything functionally but since this actually won't exist in oss I would change this slightly
```suggestion
  is_push_pool?: boolean,
```

pleek91 (approved):

PR #1545, opened 2023-07-05
Enhancement: Sort flow runs into bars efficiently

pleek91:
# Description
I believe this has the same functionality as the branch this PR is too. Sorts flow runs more efficiently without looping or sorting multiple times. @brandonreid could you please review and verify this has the correct functionality.


brandonreid (approved):
Very nice!

PR #1554, opened 2023-07-06
BugFix: Schema date-time inputs broken

### LINKED ISSUE #10158: Unable to select correct date/time for deployment flow run parameters
### First check

### Bug summary

My team is attempting to use the date selection tool to set deployment flow run parameters. The selection is not behaving as expected - the date/time being clicked isn't always what is selected, and attempting to retry by clicking the correct value is causing the other date in our params to decrement by a single day each time. The desired selected times are a week apart.

### Reproduction

https://github.com/PrefectHQ/prefect/assets/19937265/f0aef1ea-115f-4c0a-94e4-89a2b9f8b484

### Error

^ See recording in reproduction section.

### Browers

- [X] Chrome
- [X] Firefox
- [ ] Safari
- [ ] Edge

### Prefect version

_No response_

### Additional context

_No response_

pleek91:
# Description
Closes https://github.com/PrefectHQ/prefect/issues/10158

Updates the way we're mapping, editing, and comparing dates for schemas. 

- The string property service only deals with UTC values for date and date-time formats rather than trying to account for timezones
- `date` formats no longer consider timezones
- `date-time` formats take into account the user's timezone only when displaying and inputting dates
- Fixes a bug where we were comparing mapped defaults with unmapped values when making value requests.

### DIFF
--- a/src/services/schemas/properties/SchemaPropertyService.ts
+++ b/src/services/schemas/properties/SchemaPropertyService.ts
@@ -82,13 +82,11 @@ export abstract class SchemaPropertyService {
   }
 
   public mapRequestValue(value: SchemaValue): SchemaValue | undefined {
-    const mappedValue = this.request(value)
-
-    if (this.isDefaultValue(mappedValue)) {
+    if (this.isDefaultValue(value)) {
       return undefined
     }
 
-    return mappedValue
+    return this.request(value)
   }
 
   public getDefaultValue(): SchemaValue {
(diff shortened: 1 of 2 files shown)


src/services/schemas/properties/SchemaPropertyService.ts
```diff
@@ -82,13 +82,11 @@ export abstract class SchemaPropertyService {
...
   public mapRequestValue(value: SchemaValue): SchemaValue | undefined {
-    const mappedValue = this.request(value)
-
-    if (this.isDefaultValue(mappedValue)) {
+    if (this.isDefaultValue(value)) {
       return undefined
     }
 
-    return mappedValue
+    return this.request(value)
```
pleek91, line 89: This is where we were comparing a mapped value with an unmapped value. Most of the time that was fine because we were stringifying to compare. But comparing two unmapped values produces a more accurate comparison.

znicholasbrown (approved):

PR #1559, opened 2023-07-07
Enhancement: Use bottom placement for flow run scatter plot popover

### LINKED ISSUE #1460: Obtrusive Flow Runs chart mouseover.
It is not possible to gracefully mouseover all the chart's dots, specifically from left to right because the popup get's in the way. See attached.

image

**Recommended fix:** Move the popup above or below the cursor position.

pleek91:
# Description
Fixes an issue where horizontally scrubbing through flow runs was difficult due to the popover appearing directly to the right of the flow run point on the chart. 

Closes https://github.com/PrefectHQ/prefect-ui-library/issues/1460

Also removed the custom logic for closing the popover in favor of the built in functionality which didn't exist when this component was first created. 

Also added a big a padding between the popover and the flow run point so even if the popover opens in the direction of the next point if they are right next to each other the next point should be visible.

### DIFF
--- a/src/components/FlowRunPopOver.vue
+++ b/src/components/FlowRunPopOver.vue
@@ -18,7 +18,7 @@
     flowRunId: string,
   }>()
 
-  const placement = [positions.bottom, positions.top, positions.right, positions.left]
+  const placement = [positions.top, positions.bottom, positions.right, positions.left]
 </script>
 
 <style>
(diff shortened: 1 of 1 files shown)


src/components/FlowRunPopOver.vue
```diff
@@ -84,4 +37,9 @@
   w-full
   h-full
 }
+
+.flow-run-pop-over__content { @apply
+  p-2
+  pointer-events-none
```
znicholasbrown, line 43: Won't this prevent interacting with the contents of the popover? So I wouldn't be able to hover a point and then go to the run associated with it?

pleek91: hmm I feel like you are correct, I wanted to add some spacing between the trigger and the content but not have it as part of the interaction. But I think you're right this will cascade to the content itself. I'll double check

zhen0:
@pleek91 - shall we close this one?

PR #1560, opened 2023-07-08
Bugfix: Json validation

znicholasbrown:
Removes JSON validation from schema JSON inputs in favor of value mapping instead - all values are now parsed but will fall back to strings instead, which are always valid JSON. Also fixes an issue where parsing was leading to quotes getting swallowed incorrectly.

### DIFF
--- a/src/services/schemas/properties/SchemaPropertyService.ts
+++ b/src/services/schemas/properties/SchemaPropertyService.ts
@@ -1,5 +1,4 @@
 import { SelectOption } from '@prefecthq/prefect-design'
-import { JsonInput } from '@/index'
 import { InvalidSchemaValueError } from '@/models/InvalidSchemaValueError'
 import { getSchemaPropertyAttrs, getSchemaPropertyComponentWithDefaultProps, getSchemaPropertyDefaultValidators, schemaPropertyComponentWithProps, SchemaPropertyComponentWithProps } from '@/services/schemas/utilities'
 import { schemaHas, SchemaProperty, SchemaPropertyInputAttrs, SchemaPropertyMeta, SchemaValue } from '@/types/schemas'
@@ -7,7 +6,7 @@ import { Require } from '@/types/utilities'
 import { sameValue } from '@/utilities'
 import { isNumberArray, isStringArray } from '@/utilities/arrays'
 import { ComponentDefinition } from '@/utilities/components'
-import { fieldRules, isJson, ValidationMethod, ValidationMethodFactory } from '@/utilities/validation'
+import { fieldRules, ValidationMethod, ValidationMethodFactory } from '@/utilities/validation'
 
 export type SchemaPropertyServiceSource = {
   property: SchemaProperty,
--- a/src/services/schemas/properties/SchemaPropertyService.ts
+++ b/src/services/schemas/properties/SchemaPropertyService.ts
@@ -106,10 +106,6 @@ export abstract class SchemaPropertyService {
     const defaults = getSchemaPropertyDefaultValidators(this.property, required)
     const validators = fieldRules(title, ...this.validators)
 
-    if (this.componentIs(JsonInput)) {
-      validators.push(isJson(title))
-    }
-
     return [...validators, ...defaults]
   }
 
--- a/src/utilities/json.ts
+++ b/src/utilities/json.ts
@@ -33,7 +33,7 @@ export function stringifyUnknownJson(value: unknown): string | null | undefined 
   const parsed = parseUnknownJson(value)
 
   if (isString(parsed)) {
-    return value as string
+    return parsed
   }
 
   return JSON.stringify(parsed)
--- a/src/utilities/json.ts
+++ b/src/utilities/json.ts
@@ -14,7 +14,7 @@ export function parseUnknownJson(value: unknown): unknown {
     try {
       const parsed = JSON.parse(value)
 
-      // If the parsed value is a record, we return it.
+      // If the parsed value isn't a string, we return it.
       if (!isString(parsed)) {
         return parsed
       }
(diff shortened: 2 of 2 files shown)


prefectcboyd:
I tested against the fix using OSS Prefect - some of the test cases I evaluated; admittedly not all of these cases are valid JSON, but they produced the expected behavior and allowed me to submit the flow runs where they were appropriate.

Selecting the "Format" button when the JSON was indeed valid, produced the expected results.

Selecting "Format" with invalid JSON did nothing, but also no longer produces the "invalid JSON" - as I understand this __is__ expected in this case. 

Additionally, tested similar functionality in JSON fields in work pools, as well as blocks, to the same effect. 

In short - tests appear successful, lgtm!
```
"test"
["test"]
{"test"}
{"test": "value"}
{"test": value}
{"test": true}
{"test": null}
{"test": {}}
{"test": []}
{"test": [1,3,5,7,"9"]}
""
test
{"test": [
"list"
]
}
```
image

src/services/schemas/properties/SchemaPropertyService.ts
```diff
@@ -106,10 +106,6 @@ export abstract class SchemaPropertyService {
     const defaults = getSchemaPropertyDefaultValidators(this.property, required)
     const validators = fieldRules(title, ...this.validators)
 
-    if (this.componentIs(JsonInput)) {
-      validators.push(isJson(title))
-    }
-
```
pleek91, line 112: Has this ever been part of the problem? If the value from the input itself is not json the mappers are likely to fail and post a valid json string to the server. Which means errors at runtime rather than at form submission.

znicholasbrown: Yeah this is something I'd like to address longer-term. Because of the way we're mapping request values, JSON inputs would require that bare text be wrapped in quotes to be valid (according to this validator, which is technically correct). However, that means that when the values are sent to the server, they're stringified, which means this: `"test"` becomes `"\"test\""` - which is obviously incorrect. The obvious solution, which is what we were doing before, is to parse that value on the way out so that `"\"test\""` is unwrapped to `"test"`; the problem with that is that we're feeding those request values directly back into the form and parsing them again, which is why quotes were disappearing.

pleek91: Ah okay makes sense.

pleek91 (approved):

jbbrunello:
Issue is **not** resolved, the following test cases still fail:
- input value `"test_string"` (which is valid JSON) is automatically translated as "\"test_string\"" which is still valid JSON but is **not** the parameter value I need to pass to my flow (if I wanted to include escaped double quotes into my parameter value, I would have typed them in) - **I NEED THE FORMATTER TO LEAVE MY INPUT VALUE ALONE IF IT'S VALID JSON**
- input value `"\"test_string\""` (which **is** valid JSON) is automatically translated as `"\"\\\"test_string\\\"\""` which is **not** the parameter value I need to pass to my flow  - again,  **I NEED THE FORMATTER TO LEAVE MY INPUT VALUE ALONE IF IT'S VALID JSON**

For completeness:
- input value `test_string` (which is **not** valid JSON) is automatically translated as `"test_string"`, which may be correct but is uncalled for in my opinion: I did **not** click on the "Format" button
- input value `12345` (which is valid JSON) is thankfully left alone

Screenshot of my inputs:
image

Screenshot of the translated JSON:
image

prefectcboyd:
Hi @jbbrunello - this appears to be less a continuance of the original issue, and more a display issue specific when flipping between "custom" and "JSON" parameters.

For example:
I can submit the following string (which is JSON) - `"12345"`.
image

You are correct it is interpolated within the JSON UI as :
image

Running the flow produces the expected behavior though:
```
09:04:39.139 | INFO    | prefect.flow_runs.worker - Completed submission of flow run '48406b6e-dc7d-4af8-a4b3-7a9c448e0edc'
09:04:42.199 | INFO    | Flow run 'lambda3-reina' - Finished in state Completed()
"12345"
09:04:42.832 | INFO    | prefect.flow_runs.worker - Process 32528 exited cleanly.
```

For better completeness sake, I mimicked your inputs:
```
from prefect import flow
@flow
def hello_flow(
    user=None,
    debug=None,
    limit=None,
    ref_date=None,
    force_calc=None,
):
    print (f"{type(user)} and {user}")
    print (f"{type(debug)} and {debug}")
    print (f"{type(limit)} and {limit}")
    print (f"{type(ref_date)} and {ref_date}")
    print (f"{type(force_calc)} and {force_calc}")
    
    print("Hello, Prefect!")

if __name__ == "__main__":
    hello_flow()
```

I am submitting this parameter set:
```
user = "test_string"
debug = 12345
limit = "12345"
ref_date = "\"test_string\""
```

You can see here how the JSON formatter is rendering it (which is escaping):
image

The output on running however does appear correct to me - the values are rendered literally as I submitted them, not as they are displayed.

```
<class 'str'> and "test_string"
<class 'int'> and 12345
<class 'str'> and "12345"
<class 'str'> and "\"test_string\""
<class 'NoneType'> and None
Hello, Prefect!
```

Can you clarify if this is in fact submitting improperly formatted parameters to the flows, or just the display in the UI is incorrect?

jbbrunello:
Hi @znicholasbrown , fair question: the flows are in fact getting what is displayed in the JSON tab, which is preventing us to launch our flows which are expecting **a string containing only numeric characters**. 

For example, if I type `"20230630" ` in the `ref_date` parameter in the Custom tab:
image
 
it gets reformatted as { "ref_date": "\"20230630\"" } in the JSON tab:
image

and, if I then click on Submit, the flow run fails because it's getting a double-quote enclosed string, not a string containing ony numeric character as expected:
image

image

On the other hand, if I edit directly the parameters in the JSON tab:
image

and then click on Submit, the flow run fails as well because the string gets reformatted as a numeric:
image

image

Of course there are many long-term ways to address this (type-hinting, or simply passing the date in ISO format, e.g. 2023-06-30 which cannot be misinterpreted as an int), but we urgently need a - even temporary - workaround **through the web UI** (lauching flow runs through CLI is not an option).

PR #1563, opened 2023-07-10
BugFix: Schema parameters not respecting strings for certain values

### LINKED ISSUE #1561: Flow JSON Input Tab applying unwanted formatting
This seems to be a follow-on effect to #1555 and #1560 where formatting is rectified.

Issue:
Pasting a literal JSON input in the JSON inputs field of a flow submission applies unexpected formatting, instead of passing the input.

Using the following JSON object:
```
{
    "user": "John",
    "debug": "true",
    "limit": 12345,
    "ref_date": "12345",
    "force_calc": "\"12345\""
}
```

Running locally (repro at the bottom) produces:
```
Type of the item with key user:John is: <class 'str'>
Type of the item with key debug:true is: <class 'str'>
Type of the item with key limit:12345 is: <class 'int'>
Type of the item with key ref_date:12345 is: <class 'str'>
Type of the item with key force_calc:"12345" is: <class 'str'>
```

Note - ref_date `"12345"` is unwrapped to a string `12345`.

When I paste this same literal JSON block into the UI JSON tab:
image

The JSON block becomes interpreted instead, and the `"ref_date": "12345"` is unwrapped to `"ref_date": 12345` -
image

The output becomes:
```
09:46:07.599 | INFO    | Flow run 'mu4-gilrain' - <class 'str'> and John
09:46:07.608 | INFO    | Flow run 'mu4-gilrain' - <class 'bool'> and True
09:46:07.612 | INFO    | Flow run 'mu4-gilrain' - <class 'int'> and 12345
09:46:07.615 | INFO    | Flow run 'mu4-gilrain' - <class 'int'> and 12345
09:46:07.616 | INFO    | Flow run 'mu4-gilrain' - <class 'str'> and "12345"
09:46:07.617 | INFO    | Flow run 'mu4-gilrain' - Hello, Prefect!
```

pleek91:
# Description
Fixes an issue specifically for schema properties of type `Any` where we were incorrectly converting values like `"123"` (a string) into `123` (an int). 

Closes https://github.com/PrefectHQ/prefect-ui-library/issues/1561

This introduces two new json utility methods: `jsonSafeParse` and `jsonSafeStringify` with unit tests to prove their functionality. These new utilities are used in the `SchemaPropertyAny` service replacing the flawed `parseUnknownJson` and `stringifyUnknown` json utilities. 

This also changes the syncing behavior of the `SchemaInput` component so we're no longer syncing the `form` and `json` tab values as the user types. Instead this happens when switching between tabs which is not only more performant but also is much safer.

### DIFF
--- a/src/components/SchemaInput.vue
+++ b/src/components/SchemaInput.vue
@@ -99,19 +99,6 @@
     return mapper.map('SchemaValuesResponse', { values, schema: props.schema }, 'SchemaValues')
   }
 
-  function shouldSync(values: SchemaValues, json: string): boolean {
-    try {
-      const valuesString = JSON.stringify(values)
-      const jsonRequest = JSON.parse(json)
-      const mappedJson = toSchemaValues(jsonRequest)
-      const jsonString = JSON.stringify(mappedJson)
-
-      return valuesString !== jsonString
-    } catch {
-      return false
-    }
-  }
-
   const { validate: validateReactiveForm, errors: reactiveFormErrors } = useReactiveForm(values, {
     initialValues: values.value,
   })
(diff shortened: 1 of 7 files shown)


src/components/SchemaInput.vue
```diff
@@ -122,19 +110,35 @@
...
 
-      json.value = stringify(mappedValues)
+    if (success) {
+      values.value = toSchemaValues(parsed as SchemaValues)
+    }
+  }
+
+  watch(inputType, (newType, oldType) => {
+    if (newType === 'form') {
+      return syncJsonToForm
```
znicholasbrown, line 129:
```suggestion
      return syncJsonToForm()
```

znicholasbrown: I assume the intent is to call this

znicholasbrown (approved):
just 1 comment @pleek91

src/components/SchemaInput.vue
pleek91: Correct, good catch

znicholasbrown (approved):

PR #1567, opened 2023-07-11
enhancement: adding count of tasks in a running state to the dashboard card

WillRaphaelson:
This adds a blue count of running tasks, but does not add them to the line chart, because the line chart is cumulative and running is not a terminal state. it looks like this.

image


pleek91 (commented):
From a code perspective this LGTM. @brandonreid and @billpalombi curious if you have any thoughts on this from a ux perspective. I think this card was originally designed to show terminal states over time. But seems like seeing how many tasks are running (at least on the dashboard somewhere) is a good idea.

brandonreid:
Daaaang @WillRaphaelson, legit.

Yeah this is actually pretty illuminating. The Total count includes running as do the ratios. I'm going to make a PR into this one with the fallowing updates:

- Move running to the top, remove the ratio for it.
- Make the ratio of the completed and failed/crashed to be a comparison of each other.
- Also going to make the failed count gray if 0 so it's not unnecessarily red

brandonreid:
I see now that the ratio was not including running before. I do see merit in running appearing here though, so will still implement these changes.

brandonreid (approved):

PR #1574, opened 2023-07-12
anchor task runs chart to bottom

brandonreid:
This positions the chart absolute to the bottom so that it will align with the Events chart nicer (I'll update the events chart with the same styles for consistency):

Before, the events card is taller so the baselines aren't lined up:
Screenshot 2023-07-12 at 5 30 00 PM

I also noticed the left/right positioning of the chart pushed it out of bounds of the card, seemingly cutting off information (@pleek91 unless this was intentional?)
Screenshot 2023-07-12 at 5 30 11 PM

After, chart is aligned absolute to the bottom:
Screenshot 2023-07-12 at 5 30 26 PM

It's also contained within the card:
Screenshot 2023-07-12 at 5 30 48 PM

UPDATE:
Also reduced the bottom padding affordance, as the graph is cumulative, it's designed to go up and to the right, which actually lends pretty reliable space for the graph to encroach on the text's space. This was actually a part of the original design.
Screenshot 2023-07-12 at 5 43 37 PM

### DIFF
--- a/src/components/WorkspaceDashboardTaskRunsCard.vue
+++ b/src/components/WorkspaceDashboardTaskRunsCard.vue
@@ -222,6 +222,8 @@
 .workspace-dashboard-task-runs-card__summary { @apply
   grid
   gap-1
+  relative
+  z-10
 }
 
 .workspace-dashboard-task-runs-card__statistic--completed .dashboard-statistic__value { @apply
--- a/src/components/WorkspaceDashboardTaskRunsCard.vue
+++ b/src/components/WorkspaceDashboardTaskRunsCard.vue
@@ -216,7 +216,7 @@
   gap-4
   auto-rows-max
   overflow-hidden
-  pb-20
+  pb-8
 }
 
 .workspace-dashboard-task-runs-card__summary { @apply
--- a/src/components/WorkspaceDashboardTaskRunsCard.vue
+++ b/src/components/WorkspaceDashboardTaskRunsCard.vue
@@ -216,6 +216,7 @@
   gap-4
   auto-rows-max
   overflow-hidden
+  pb-20
 }
 
 .workspace-dashboard-task-runs-card__summary { @apply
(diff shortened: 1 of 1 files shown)


src/components/WorkspaceDashboardTaskRunsCard.vue
```diff
@@ -216,11 +216,14 @@
   gap-4
   auto-rows-max
   overflow-hidden
+  pb-8
```
pleek91, line 219: Is this a magic number or is this based on another value?  I expected `pb-16` to match the `h-16` of the chart since it is now `absolute`.

src/components/WorkspaceDashboardTaskRunsCard.vue
```diff
@@ -237,17 +240,20 @@
...
 }
 
 .workspace-dashboard-task-runs-card__chart { @apply
   absolute
   min-h-0
-  -left-8
+  left-0
   top-0
-  -right-8
+  right-0
```
pleek91, line 256: good catch. It was always intended to span the width of the card and not to be wider than the card itself.

pleek91 (approved):

src/components/WorkspaceDashboardTaskRunsCard.vue
brandonreid: So it's actually the card's default padding plus "some more". I added an update to this ticket's description. TLDR: the original design anticipates the chart grows up and to the right, so there's always a gap below the text, we don't need to actually clear the total potential height of the graph. Lmk if I should add a comment?

In the event that there is a collision, the text will wrest on top, I'm hoping this is pretty rare. Here I manually moved the graph to the left to force a collision to see how bad it could  and what would be affected:
Screenshot 2023-07-12 at 5 52 02 PM

I could reduce that potential overlap a bit.. Curious to hear your thoughts on that.

pleek91: Makes sense. I originally allowed for the overlap but after putting all the pieces on the dashboard it never did because the task run card would always match the height of the taller event card. But there are certain screen sizes where they are not locked together and OSS doesn't have events at all. So I think this makes sense. 

No comment needed.

PR #1576, opened 2023-07-13
fix padding on task runs card

brandonreid:
This PR undoes a negative gap on the Task Runs card.

Last night I introduced a UI change that backfired in the morning. Essentially because the graph on the Task Runs and Events graphs (not changed yet) are cumulative and will typically go up and to the right, a negative gap between the text and the graph is seen as acceptable, allowing the graph to curve up and around the text in most cases. However, this morning I woke to this and felt it's not worth:

Before:
image

After:
image

### DIFF
--- a/src/components/WorkspaceDashboardTaskRunsCard.vue
+++ b/src/components/WorkspaceDashboardTaskRunsCard.vue
@@ -216,14 +216,12 @@
   gap-4
   auto-rows-max
   overflow-hidden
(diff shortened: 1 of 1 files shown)


zhen0 (approved):
Thanks @brandonreid

znicholasbrown (approved):

zhen0 (approved):
Thanks @brandonreid

PR #1580, opened 2023-07-14
sort flows on the flow runs card

brandonreid:
Sorts the flows newest to oldest on the Flow Runs card:

Before:
Screenshot 2023-07-14 at 7 35 04 AM

After:
Screenshot 2023-07-14 at 7 34 07 AM

### DIFF
--- a/src/components/FlowRunsAccordion.vue
+++ b/src/components/FlowRunsAccordion.vue
@@ -29,7 +29,10 @@
     // eslint-disable-next-line no-unused-vars
     const { sort, limit, offset, ...filter } = props.filter
 
-    return filter
+    return {
+      ...filter,
+      sort: 'UPDATED_DESC',
+    }
   })
 
   const options = useInterval()


pleek91 (commented):
Won't this mean that the order of this list might change each time it refreshes? 

This does make sense to order this way but wondering if a busy workspace is going to have stuff jumping around. 

Also, surprised that the updated timestamp gets set when a flow runs.

brandonreid:
> Won't this mean that the order of this list might change each time it refreshes?
> 
> This does make sense to order this way but wondering if a busy workspace is going to have stuff jumping around.
> 
> Also, surprised that the updated timestamp gets set when a flow runs.

Hmm. Good points for sure. I think for the failed/crashed especially, the sorting is pretty key. Hopefully user's don't have too many of those happening normally. The Completed tab or others could get a bit more jumpy. I'm thinking we go for it and see if we get push back here. The nature of the dashboard is to surface the helpful thing from the bunch and likely navigate to it, hopefully if they want a more stable list they use other pages.

pleek91 (approved):

PR #1581, opened 2023-07-14
Fix/work pool card items layout

brandonreid:
Rearranges some content on the work pool list items for the work pools card so that things have room to breath:

After:
Screenshot 2023-07-14 at 10 03 15 AM

After with late runs:
Screenshot 2023-07-14 at 10 02 20 AM

Before:
Screenshot 2023-07-14 at 8 24 29 AM

### DIFF
--- a/src/components/DashboardWorkPoolCard.vue
+++ b/src/components/DashboardWorkPoolCard.vue
@@ -20,9 +20,11 @@
         <WorkPoolQueueStatusArray :work-pool="workPool" />
       </DashboardWorkPoolCardDetail>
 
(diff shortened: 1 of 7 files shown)


pleek91 (approved):
I don't feel strongly but combining lateness and late runs feels a little forced. I wonder if making late runs the primary statistic with total lateness being the extra piece of detail would make that feel more natural.

brandonreid:
> I don't feel strongly but combining lateness and late runs feels a little forced. I wonder if making late runs the primary statistic with total lateness being the extra piece of detail would make that feel more natural.

Ah I dig that idea. I'll mess that rn.

PR #1582, opened 2023-07-14
hide failed line in Task Runs card if none

brandonreid:
Felt weird to have a flat red line when task runs have ran 100% as Completed

Screenshot 2023-07-14 at 9 12 32 AM
Screenshot 2023-07-14 at 9 13 42 AM

### DIFF
--- a/src/components/WorkspaceDashboardTaskRunsCard.vue
+++ b/src/components/WorkspaceDashboardTaskRunsCard.vue
@@ -11,7 +11,7 @@
     </div>
 
     <div class="workspace-dashboard-task-runs-card__chart-container">
-      <LineChart v-if="taskRunsChartData.hasFailed" :data="taskRunsChartData.failed" :options="{ maxValue: maxFailedValue }" class="workspace-dashboard-task-runs-card__chart workspace-dashboard-task-runs-card__chart--failed" />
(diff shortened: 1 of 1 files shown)


src/components/WorkspaceDashboardTaskRunsCard.vue
```diff
@@ -11,7 +11,7 @@
     </div>
 
     <div class="workspace-dashboard-task-runs-card__chart-container">
-      <LineChart :data="taskRunsChartData.failed" :options="{ maxValue: maxFailedValue }" class="workspace-dashboard-task-runs-card__chart workspace-dashboard-task-runs-card__chart--failed" />
+      <LineChart v-if="taskRunsChartData.hasFailed" :data="taskRunsChartData.failed" :options="{ maxValue: maxFailedValue }" class="workspace-dashboard-task-runs-card__chart workspace-dashboard-task-runs-card__chart--failed" />
```
pleek91, line 14: Nit: you don't need this new boolean if you just check that failed > 0 same as with the count statistic above.

brandonreid: ah danget, will update

pleek91 (approved):

PR #1583, opened 2023-07-14
Enhancement: Remove the term "Work Pool Queue" in favor of just "Work Queue"

pleek91:
# Description
We no longer refer to work queues in a work pool as "work pool queues" but just "work queues". Updating the language to reflect that.


jamiezieziula (approved):
ty! 🚀

brandonreid (approved):

PR #1592, opened 2023-07-17
make some dashboard components reusable, add FlowRunHistoryCard

brandonreid:
This makes some of the dashboard cards generic and adds the FlowRunHistoryCard component.

A nebula PR is ready, I'll make a OSS PR along with it once this is approved.

Here's what it looks like in Nebula local:
image

Feedback on some of the naming conventions would be great!


src/components/SimpleTimeSpanFilter.vue
pleek91, line 1: nit: I don't think the prefix simple adds and value here and I would remove it.

src/components/WorkspaceDashboardFlowRunsCard.vue
```diff
@@ -15,7 +15,7 @@
 
 <script lang="ts" setup>
   import { computed } from 'vue'
-  import DashboardStatistic from '@/components/DashboardStatistic.vue'
+  import ValueKeyStatistic from '@/components/ValueKeyStatistic.vue'
```
pleek91, line 18: linting warning

src/types/flow.ts
```diff
@@ -0,0 +1,4 @@
+export type FlowStatsFilter = {
```
pleek91, line 1: Trying to figure out where this is used outside of the mappers. I'm assuming this is used at implementation on the actual flow page?

brandonreid: It's only used in mappers, but in nebula there's an additional `mapFlowStatsFilterTo**EventsFilter**` that uses it

pleek91: Yep, makes sense 👍

pleek91 (approved):

PR #1598, opened 2023-07-18
fix dashboard and flow stats polling

brandonreid:
Removes the end date requirements for the dashboard and Flow Stats related mappers. That way when polling happens, they'll always fetch data up to the latest moment.

### DIFF
--- a/src/components/FlowRunsBarChart.vue
+++ b/src/components/FlowRunsBarChart.vue
@@ -1,5 +1,5 @@
 <template>
-  <div ref="chart" class="flow-runs-bar-chart" :class="classes.root" :style="styles.root" @mouseleave="close">
+  <div ref="chart" class="flow-runs-bar-chart" :class="classes.root" @mouseleave="close">
     <template v-for="(flowRun, index) in barFlowRuns" :key="getKey(flowRun, index)">
       <p-pop-over class="flow-runs-bar-chart__bar-container" :to="chart" :placement="placement" :group="group" auto-close>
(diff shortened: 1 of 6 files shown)


src/components/FlowRunsBarChart.vue
```diff
@@ -163,9 +160,10 @@
 <style>
 .flow-runs-bar-chart { @apply
   relative
-  grid
-  items-end
-  grid-rows-1;
+  w-full
+  h-full
+  flex
```
brandonreid, line 165: I switched this to flex to remove the need to calculate columns, plus in a situation where for some reason more bars are trying to render than the width allows then the bars won't start wrapping, they'll just overflow to the side.

pleek91 (changes requested):
Correct me if I'm wrong. But doesn't this change make it so as you leave the dashboard open the time span you're looking at grows? So 8 hours turns into 9 after leaving it open for an hour.

brandonreid:
> Correct me if I'm wrong. But doesn't this change make it so as you leave the dashboard open the time span you're looking at grows? So 8 hours turns into 9 after leaving it open for an hour.

That's correct. I called that out in the #raft-workspace-dashboard channel. It shou

pleek91:
I'm going to close this PR in favor of implementing this change https://github.com/PrefectHQ/prefect-ui-library/pull/1602

That PR doesn't fix this but allows us to fix it without having the growing time span. I'll work on that today or tomorrow.

PR #1601, opened 2023-07-19
BugFix: Update useWorkPoolLastPolled to not accept a filter

pleek91:
# Description
The `useWorkPoolLastPolled` composition accepted a filter which was being used to limit the work queues. But this defeats the point of knowing then the work pool was last polled. Removing that filter to get an accurate time

### DIFF
--- a/src/components/DashboardWorkPoolCard.vue
+++ b/src/components/DashboardWorkPoolCard.vue
@@ -13,7 +13,7 @@
     </div>
     <dl class="dashboard-work-pool-card__details">
       <DashboardWorkPoolCardDetail label="Polled">
-        <WorkPoolLastPolled :work-pool="workPool" :filter="filter" />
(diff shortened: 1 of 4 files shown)


brandonreid (approved):
I think you'll want to also cleanup the filter being passed in [here](https://github.com/PrefectHQ/prefect-ui-library/blob/a2ee94677542c62dccf4e79957963eaa67eb63d7/src/components/DashboardWorkPoolCard.vue#L16), otherwise lgtm! Thanks for catching that.

pleek91:
> I think you'll want to also cleanup the filter being passed in [here](https://github.com/PrefectHQ/prefect-ui-library/blob/a2ee94677542c62dccf4e79957963eaa67eb63d7/src/components/DashboardWorkPoolCard.vue#L16), otherwise lgtm! Thanks for catching that.

Nice catch. Removed!

PR #1602, opened 2023-07-19
Enhancement: Update all entity compositions to accept getters and use getters for parameters rather than computed properties

pleek91:
# Description
## Motivation
On the workspace dashboard we have a time span like 8hr, 24h, 1w. That span does not change each time a poll happens. So the longer a user stays on the page the more outdated their data becomes. Allowing compositions to accept getters means that the parameters of an action can be dynamic and update each time the subscription polls. 

## Changes
I went ahead and updated all of our compositions to follow the same pattern. We had a mix if different patterns. But almost every change in this PR is updating the argument from something like `string | Ref<string | null | undefined>` to `MaybeRefOrGetter<string | null | undefined>`. And then changing the parameters computed property to be a getter. 

## Implementation
The implementation difference is subtle. 
```javascript
// computed
const filter = computed(() => mapper.map('DashboardFilters', dashboardFilter, 'FlowRunsFilter'))
```
versus
```javascript
// getter
const filter = () => mapper.map('DashboardFilters', dashboardFilter, 'FlowRunsFilter')
```

Both have the same reactivity when used with `useFlowRuns`. However since getters are evaluated each time they are referenced the getter version will run the mapper each time the subscription is executed. The mapper in the examples converts a time span (like 8 hours) into start and end dates. Using a computed, those dates never get re-calculated. However using a getter means the dates will be calculated each time the subscription is executed.

### DIFF
--- a/src/compositions/useBlockDocument.ts
+++ b/src/compositions/useBlockDocument.ts
@@ -1,29 +1,32 @@
 import { useSubscriptionWithDependencies } from '@prefecthq/vue-compositions'
-import { computed, ref, Ref } from 'vue'
+import { computed, MaybeRefOrGetter, toRef, toValue } from 'vue'
 import { useCan } from '@/compositions/useCan'
 import { useWorkspaceApi } from '@/compositions/useWorkspaceApi'
 import { WorkspaceBlockDocumentsApi } from '@/services/WorkspaceBlockDocumentsApi'
+import { Getter } from '@/types/reactivity'
 import { UseEntitySubscription } from '@/types/useEntitySubscription'
 
 export type UseBlockDocument = UseEntitySubscription<WorkspaceBlockDocumentsApi['getBlockDocument'], 'blockDocument'>
 
-export function useBlockDocument(blockDocumentId: string | Ref<string | null | null>): UseBlockDocument {
-  const id = ref(blockDocumentId)
+export function useBlockDocument(blockDocumentId: MaybeRefOrGetter<string | null | null>): UseBlockDocument {
   const api = useWorkspaceApi()
   const can = useCan()
 
-  const parameters = computed<[string] | null>(() => {
-    if (!id.value) {
-      return null
-    }
-
+  const getter: Getter<[string] | null> = () => {
     if (!can.read.block) {
       return null
     }
 
-    return [id.value]
-  })
+    const id = toValue(blockDocumentId)
 
+    if (!id) {
+      return null
+    }
+
+    return [id]
+  }
+
+  const parameters = toRef(getter)
   const subscription = useSubscriptionWithDependencies(api.blockDocuments.getBlockDocument, parameters)
   const blockDocument = computed(() => subscription.response)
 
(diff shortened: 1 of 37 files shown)


brandonreid (approved):
Reviewed but this may be worth a second pair of eyes. Fingers crossed that typescript has our backs here, haha. Thanks for making this enhancement!

src/compositions/useBlockDocument.ts
```diff
@@ -1,29 +1,32 @@
...
+export function useBlockDocument(blockDocumentId: MaybeRefOrGetter<string | null | null>): UseBlockDocument {
   const api = useWorkspaceApi()
   const can = useCan()
 
-  const parameters = computed<[string] | null>(() => {
-    if (!id.value) {
+  const getter: Getter<[string] | null> = () => {
+    if (!can.read.block) {
       return null
     }
```
znicholasbrown, line 18: I'm assuming the switch in checks for this are to avoid making an extra `toValue` call that we don't need to?

pleek91: Correct, and I think it makes sense in general to do permission checks first.

src/compositions/useFlowRunsCount.ts
```diff
@@ -1,31 +1,39 @@
 import { SubscriptionOptions, UseSubscription, useSubscriptionWithDependencies } from '@prefecthq/vue-compositions'
-import { ComputedRef, Ref, computed, ref } from 'vue'
+import merge from 'lodash.merge'
```
znicholasbrown, line 2: Here and elsewhere we should be using bucket imports for lodash
```suggestion
import { merge } from 'lodash'
```
https://lodash.com/per-method-packages

src/types/reactivity.ts
```diff
@@ -1,7 +1,9 @@
 import { Ref } from 'vue'
 import { AnyArray, AnyRecord } from '@/types/any'
 
+export type Getter<T> = () => T
```
znicholasbrown, line 4: Oh no wonder I couldn't import this type from Vue lol

znicholasbrown (approved):
Minor comments, overall LGTM

src/compositions/useBlockSchema.ts
```diff
@@ -1,22 +1,26 @@
...
 import { useWorkspaceApi } from '@/compositions/useWorkspaceApi'
 import { WorkspaceBlockSchemasApi } from '@/services/WorkspaceBlockSchemasApi'
+import { Getter } from '@/types/reactivity'
 import { UseEntitySubscription } from '@/types/useEntitySubscription'
 
 export type UseBlockSchema = UseEntitySubscription<WorkspaceBlockSchemasApi['getBlockSchema'], 'blockSchema'>
 
-export function useBlockSchema(blockSchemaId: string | Ref<string | null | null>): UseBlockSchema {
-  const id = ref(blockSchemaId)
+export function useBlockSchema(blockSchemaId: MaybeRefOrGetter<string | null | null>): UseBlockSchema {
```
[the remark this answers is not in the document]
pleek91: good catch! You are correct.

src/compositions/useFlowRunsCount.ts
pleek91: I would agree if we had lodash itself installed and I understand that in v5 the individual packages will be deprecated and removed. But currently we have individual v4 packages installed and I actually believe the import you're suggesting here will break in production.

src/types/reactivity.ts
pleek91: I was very surprised they didn't export this.

PR #1603, opened 2023-07-20
Fix: Hide the link to work queues for push work pools

zhen0:
Push work pools do not use work queues and the work queues tab is hidden on the work pool page for any push type work pool.  However, we are still showing the work queue as default and displaying a link to a work queue in the deployment and flow run pages (and flow run list item).  This PR adds a check for whether a work pook is a push type work pool and, if it's a push type pool, hides the link to a work queue. 

Current:

image

Updated:

image


src/components/DeploymentDetails.vue
```diff
@@ -107,6 +107,9 @@
   const api = useWorkspaceApi()
   const updateScheduleLoading = ref(false)
 
+  const { workPool } = useWorkPool(props.deployment.workPoolName ?? '')
+  const isPushWorkPool = computed(() => workPool.value?.isPushPool)
```
pleek91, line 111: Nit: we really don't need a computed for this. `workPool.isPushPool` can be used directly

src/components/FlowRunWorkQueue.vue
```diff
@@ -18,6 +19,8 @@
   }>()
 
   const isNotTerminal = computed(() => props.flowRunState && !isTerminalStateType(props.flowRunState))
+  const { workPool } = useWorkPool(props.workPoolName ?? '')
+  const isPushWorkPool = computed(() => workPool.value?.isPushPool)
```
pleek91, line 23: Nit: same suggestion here to just use `workPool.isPushPool` directly.

src/compositions/useWorkPool.ts
pleek91, line 1: ❤️

pleek91 (approved):
I'd remove those computed properties but otherwise this looks good to me!

zhen0:
@pleek91 - double checking the computed check is what you were expecting before I merge!

src/components/FlowRunWorkQueue.vue
```diff
@@ -18,6 +19,7 @@
   }>()
 
   const isNotTerminal = computed(() => props.flowRunState && !isTerminalStateType(props.flowRunState))
+  const { workPool } = useWorkPool(props.workPoolName ?? '')
```
pleek91, line 22: Looking at this closer this probably works but might not work quite as expected. You'll probably want to do

```suggestion
  const { workPoolName } = toRefs(props)
  const { workPool } = useWorkPool(workPoolName)
```

Which will retain reactivity even if the prop starts as undefined but changes.

pleek91 (approved):
One last suggestion but feel free to merge.

PR #1604, opened 2023-07-20
Enhancement: Make the work pools wizard more push work pool friendly

zhen0:
Follow on to #1603 

Currently the first step in the work pool create wizard is to add details.  This PR switches the order around so that details are last. This:
1. Makes the work pool creation wizard consistent with the automations wizard
2. Allows us to hide the concurrency input on the details step for push work pools

This PR also adds the logic to hide the input for push work pools.


pleek91 (approved):
Nice, LGTM!

PR #1613, opened 2023-07-25
BugFix: Variable names with multiple underscores invalid in ui when creating a new variable

pleek91:
# Description
Updates all the components and compositions used in the dashboard to accepts and pass getters rather than refs for filters. This way the dashboard filters can update with each polling interval so the 24h time span constantly updates.


src/components/DashboardWorkPoolFlowRunCompleteness.vue
pleek91, line 1: This isn't completely new. Git didn't follow the rename well.

pleek91:
@znicholasbrown please let me know if this is to large and I can break it up into separate PRs per card.

src/compositions/useTaskRunsCount.ts
```diff
@@ -1,14 +1,15 @@
...
 import { TaskRunsFilter } from '@/models/Filters'
 import { WorkspaceTaskRunsApi } from '@/services/WorkspaceTaskRunsApi'
 import { Getter } from '@/types/reactivity'
 import { UseEntitySubscription } from '@/types/useEntitySubscription'
 
-export type UseTaskRunsCount = UseEntitySubscription<WorkspaceTaskRunsApi['getTaskRunsCount'], 'taskRunsCount'>
+export type UseTaskRunsCount = UseEntitySubscription<WorkspaceTaskRunsApi['getTaskRunsCount'], 'count'>
 
-export function useTaskRunsCount(flowRunId: MaybeRefOrGetter<string | null | undefined>): UseTaskRunsCount {
+export function useTaskRunsCount(filter: MaybeRefOrGetter<TaskRunsFilter | null | undefined>, options?: SubscriptionOptions): UseTaskRunsCount {
```
pleek91, line 12: Technically this is a breaking change. However this composition is yet to be used outside of ui-library so this can be released without breaking oss or cloud.

src/compositions/useTaskRunsHistory.ts
```diff
@@ -0,0 +1,45 @@
+import { SubscriptionOptions, UseSubscription, useSubscriptionWithDependencies } from '@prefecthq/vue-compositions'
+import merge from 'lodash.merge'
+import { computed, ComputedRef, MaybeRefOrGetter, toRef, toValue } from 'vue'
+import { TaskRunHistory } from '..'
```
znicholasbrown, line 4:
```suggestion
import { TaskRunHistory } from '@/models'
```
I think this is the correct import but I'm not totally sure

src/compositions/useTaskRunsCount.ts
```diff
@@ -17,30 +18,30 @@ export function useTaskRunsCount(flowRunId: MaybeRefOrGetter<string | null | und
...
   const parameters = toRef(getter)
-  const subscription = useSubscriptionWithDependencies(api.taskRuns.getTaskRunsCount, parameters)
-  const taskRunsCount = computed(() => subscription.response)
+  const subscription = useSubscriptionWithDependencies(api.taskRuns.getTaskRunsCount, parameters, options)
+  const count = computed(() => subscription.response)
 
   return {
     subscription,
-    taskRunsCount,
+    count,
```
znicholasbrown, line 45: ❤️

src/components/WorkPoolAverageLateTime.vue
```diff
@@ -1,39 +1,41 @@
...
+  const lateFlowRunsFilter: Getter<FlowRunsFilter> = () => {
+    const filter: FlowRunsFilter = {
+      workPools: {
+        name: [props.workPool.name],
       },
-    },
-  }))
+      flowRuns: {
+        state: {
+          name: ['Late'],
```
znicholasbrown, line 30: Am I correct in assuming the API accepts and we have use cases for passing flow run states that _aren't_ late to the average lateness composition? If not, it might make sense to bake this into the composition

src/components/PageHeadingFlowRun.vue
```diff
@@ -85,12 +83,15 @@
...
+  const { flowRun } = useFlowRun(() => props.flowRunId, { interval: 30000 })
 
   const isPending = computed(() => flowRun.value?.stateType ? isPendingStateType(flowRun.value.stateType) : true)
-  const { taskRunsCount } = useTaskRunsCount(flowRunId)
+
+  const { count: taskRunsCount } = useTaskRunsCount(() => ({
+    flowRuns: {
+      id: [props.flowRunId],
+    },
+  }))
```
znicholasbrown, line 94: The anonymous function makes this a little tougher to read than I'd hope (the one on line 86 is fine)

src/components/FlowRunsAccordion.vue
```diff
@@ -2,41 +2,42 @@
...
   import { toMap } from '@/utilities'
 
   const props = defineProps<{
-    filter: FlowRunsFilter,
+    filter: MaybeGetter<FlowRunsFilter>,
   }>()
 
-  const flowsFilter = computed<FlowsFilter>(() => {
+  const flowRunsFilter = toRef(props.filter)
+  const flowsFilterGetter: Getter<FlowsFilter> = () => {
```
znicholasbrown, line 30: I'm not sure about this rename - we don't use variableGetter elsewhere and I'm not sure it's necessary. For consistency I'd prefer we retained `flowsFilter` instead

src/components/FlowRunStateTypeTabs.vue
```diff
@@ -54,19 +56,19 @@
...
-    const types = getTabStates(tab)
-
-    return {
-      ...props.filter,
-      flowRuns: {
-        ...props.filter?.flowRuns,
-        state: {
-          ...props.filter?.flowRuns?.state,
-          type: types,
+  function getStateTypeFilterGetter(tab: string): Getter<FlowRunsFilter> {
```
znicholasbrown, line 59: And in direct conflict with my other comment, I think this name change is fine since it's describing the return value lol

src/components/FlowRunListItem.vue
```diff
@@ -65,8 +65,11 @@
   const tags = computed(() => props.flowRun.tags)
   const value = computed(() => props.flowRun.id)
 
-  const flowRunId = computed(() => props.flowRun.id)
-  const { taskRunsCount } = useTaskRunsCount(flowRunId)
+  const { count: taskRunsCount } = useTaskRunsCount(() => ({
```
znicholasbrown, line 68: Same here wrt anonymous function not helping readability

znicholasbrown (approved):
Added a few comments wrt passing anonymous functions that return nested objects but overall I don't have any concerns with the changes

src/components/PageHeadingFlowRun.vue
pleek91: Personally for small filters like this I prefer this and find it reads nicely. But if it gets much larger than this, yeah I'd wanna break it out.

src/components/WorkPoolAverageLateTime.vue
pleek91: Hmm good catch. I'm actually not sure this is correct at all. But I'll check on it

src/components/PageHeadingFlowRun.vue
znicholasbrown: Ack - fine by me

src/components/FlowRunsAccordion.vue
pleek91: yeah I don't disagree. Will update.

src/components/FlowRunStateTypeTabs.vue
pleek91: Lol same cause it returns a getter.

src/components/FlowRunListItem.vue
pleek91: Talked about this in slack. In github it is kinda hard to parse. But in vscode it looks nice. For now we'll keep these but might revisit

image

src/components/WorkPoolAverageLateTime.vue
pleek91: This shouldn't even have the late state filter. So removed that. Thanks for calling that out!

PR #1618, opened 2023-07-31
Enhancement: Add ability to select a work queue on a custom flow run

### LINKED ISSUE #1615: Expose work queue name on flow run create
Recently the ability to specify a work queue name for an individual flow run was [exposed in the API](https://github.com/PrefectHQ/prefect/pull/10276) (both OSS and Cloud).  We should reflect this ability on the flow run create form in the UI as well; ideally it work exactly as the "tags" form works: a list of available work queue names are presented based on the deployment's work pool, and the ability to create a new one is also there (the backend will create a new work queue if the name provided doesn't already exist, so the UI doesn't need to take that responsibility on).

zhen0:
https://github.com/PrefectHQ/prefect-ui-library/assets/40272060/d6e3b1b0-387c-43b3-9e84-a201d0d316a5

Closes https://github.com/PrefectHQ/prefect-ui-library/issues/1615

@cicdw - to keep scope small, this PR makes use of our existing WorkPoolQueueComboBox component. That means a user can select an existing work queue but can't create a new one.  I'm happy to create a new issue for a follow up where we use tags so a new work queue can be passed to the api if you think the ability to create a work queue from this form input is needed.


cicdw (approved):
nice, this looks great!

src/components/FlowRunCreateForm.vue
```diff
@@ -149,6 +154,11 @@
   const name = ref<string>(generateRandomName())
   const parameters = ref<SchemaValues>(combinedParameters.value)
   const stateMessage = ref<string>('')
+  const workQueueName = ref<string | null>(props.deployment.workQueueName)
+  const workPoolName = ref<string | null>(props.deployment.workPoolName)
+
+  const { workPool } = useWorkPool(props.deployment.workPoolName ?? '')
+  const workQueueComboboxLabel = computed(() => `Work Queue for ${workPoolName.value} (Optional)`)
```
znicholasbrown, line 161: Double checking - is `Work Queue` correct for the label here or should it be `Work queue`?

znicholasbrown (approved):
1 requested change, otherwise lgtm @zhen0

src/components/FlowRunCreateForm.vue
zhen0: I had to go back and check the style guide! I think labels should be using title case which means that all words expect for articles and short prepositions (which would include for) should be upper case. That also led me to add the capitalization of 'Optional'

src/components/FlowRunCreateForm.vue
```diff
@@ -26,6 +26,10 @@
         <p-tags-input v-model="tags" :options="deploymentTags" />
       </p-label>
 
+      <p-label v-if="workPoolName && !workPool?.isPushPool" :label="workQueueComboboxLabel">
```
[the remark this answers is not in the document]
zhen0: This is here to work with the WorkPoolQueueComboBox which has `workPoolName` as a required prop so throws an error if we're not confirming it's there.  And we need the check on pushPool because they do have work queues but we don't want to show them.

zhen0:
Sorry for the delay on this @znicholasbrown but should be good for re-review

znicholasbrown (approved):
all good, thank you @zhen0 !

PR #1622, opened 2023-08-04
Removes workers beta message

EmilRex:
Removes the "Workers are currently in beta" message on workers table view since workers are now GA.


EmilRex:
@brandonreid Yep, nice catch!

brandonreid (approved):

PR #1637, opened 2023-08-15
replace background color tokens

brandonreid:


src/components/NotificationForm.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <p-form class="notification-form" @submit="submit" @cancel="cancel">
+  <p-form class="notification-form p-background" @submit="submit" @cancel="cancel">
```
znicholasbrown, line 2: Here and elsewhere, please put additional classes within the containing BEM class, e.g.:

```css
.notification-form { @apply
  p-background
}

Comment (pleek91) [reply src/components/NotificationForm.vue:2]: So we can't do that for the p-background class because we're using actual class selectors like `.p-background .p-background` to do the "automatic layers". And using `@apply` removes the class and just adds the background color directly (which would be incorrect for any layers). 

But I would like to find a way to do this that works with `@apply`. Haven't had a viable idea yet.

Comment (brandonreid) [reply src/components/NotificationForm.vue:2]: Yeah so this is a bit of a snowflake in the strategy. You can read about it [here](https://deploy-preview-912--prefect-design.netlify.app/design-tokens#backgrounds-and-borders).

Basically what we have is 3 total background levels (plus 'code' and 'floating'). The root bg color is level 0. Then by adding this p-background class you'll get bg-1 or bg-2, under the hood is a simple `.p-background { bg-1}` and `.p-background .p-background { bg-2}`. This makes the background assignment automatic and is pretty handy for components that may be in a card sometimes, and other times not.

In day-to-day use, hopefully p-background isn't required too often.

File: src/styles/states.css [hunk 2]
```diff
@@ -55,6 +55,6 @@
 }
 
 .state--unknown { @apply
-  bg-background-300
-  text-foreground-700
+  text-foreground-700;
+  background-color: var(--p-color-sentiment-neutral);
```
pleek91, line 59: isn't there a tailwind utility for sentiment neutral?

brandonreid: There totally is, will update

src/components/NotificationForm.vue
znicholasbrown: Oof ok, thanks for the explanation. I don't have any ideas at the moment but I'm seeing this in a lot of PRs. I'll think about how we could handle that, there might not be a CSS-only solution with that setup

pleek91: For now I think the trade off of having a special class (negative) for "automatic background colors" (positive) is worth it. And FWIW this is the css only solution. Just not a tailwind solution lol

znicholasbrown (approved):

PR #1638, opened 2023-08-17
replace all instances of foreground tokens

brandonreid:


src/styles/states.css
```diff
@@ -54,7 +54,6 @@
   dark:text-state-crashed-50
 }
 
-.state--unknown { @apply
-  text-foreground-700;
```
brandonreid, line 58: All these states will be hit up with their own tokens once the build is stable

src/components/SearchInput.vue
```diff
@@ -45,7 +45,7 @@
 }
 
 .search__icon { @apply
-  text-foreground-400
-  ml-2
+  ml-2;
+  color: var(--p-color-input-placeholder);
```
pleek91, line 49: should there be a utility class for placeholder?

pleek91 (approved):

src/components/SearchInput.vue
brandonreid: So this is a component level token, in general I've kept those as css variables. If placeholder keeps popping up for some reason, we could consider making that more generic.

PR #1640, opened 2023-08-17
Update buttons

brandonreid:


src/components/FlowListItem.vue
```diff
@@ -156,12 +155,14 @@
...
-  transition-transform
   mr-2
 }
 
-.flow-list-item__content-toggle--expanded { @apply
+.flow-list-item__content-toggle > * { @apply
+  transition-transform
+}
+
+.flow-list-item__content-toggle--expanded > * { @apply
```
pleek91, line 165: Curious why this selector needed to be updated like this. I'd like to avoid using `*` if possible. Also the transition should be inherited right?

src/components/PageHeadingApiKeys.vue
```diff
@@ -2,7 +2,7 @@
   <page-heading class="page-heading-api-keys" :crumbs="crumbs">
     <template #after-crumbs>
       <!-- Note: This doesn't do anything yet -->
-      <p-button inset size="xs" icon="PlusIcon" />
+      <p-button small icon="PlusIcon" />
```
pleek91, line 5: I wonder if this component is even used judging by the comment for this button

pleek91 (commented):
Coule questions. Overall LGTM. Love that this PR removed more than it added

src/components/FlowListItem.vue
brandonreid: Yeah just trying to target the icon in a sustainable way.. I should probably just target `.p-icon`, huh? That way it's all good if extra containers get added in p-button or something

pleek91: Yeah that would be better. This is one case where having a slot for the icon might be preferred so that at implementation the icon can have its own class. I like the devx of the icon props. But having slots for them might be nice for stuff like this. 

For now updating this to target the `p-icon` itself would be an improvement 👍

src/components/PageHeadingApiKeys.vue
brandonreid: I'm sure it's not. As I review each component individually (once I get a build working), I might try and check the usage of a lot of these components and see what can be deleted.

pleek91 (approved):

PR #1642, opened 2023-08-17
Fix page heading

brandonreid:


src/components/PageHeading.vue
```diff
@@ -15,19 +15,14 @@
...
-      required: true,
-    },
-    size: {
-      type: String as PropType<Size>,
-      default: 'xl',
-    },
-  })
+  const props = defineProps<{
+    crumbs: BreadCrumbs,
+    size?: Size,
```
pleek91, line 24: Before there was a default of `xl` which I don't think is accounted for now.

brandonreid: image

pleek91 (approved):

PR #1644, opened 2023-08-18
General sweep

brandonreid:
I ran through every component in the /components folder and checked for color adjustments. Fortunately most components don't have any color styles.

After this, I'll need to address state colors, tags, and a short list of components I'll address individually.


src/components/FlowList.vue
```diff
@@ -230,20 +229,21 @@
...
+  mt-4
+}
+
+.flow-list__filter-button--filter-open { @apply
+  cursor-pointer
 }
 
-.flow-list__filter-button--filter-active .p-icon { @apply
-  text-link
+.flow-list__filter-button--filter-active:not(.flow-list__filter-button--filter-open) .p-icon { @apply
```
pleek91, line 246: high specificity selectors like this are a code smell IMO. Anyway we can avoid this? Seems like a third dynamic class could be used here.

src/components/PythonInput.vue
pleek91, line 1: Because it was unused?

pleek91 (approved):

src/components/PythonInput.vue
brandonreid: Yeah, marked deprecated and I searched here/nebula/oss, no usage.

PR #1646, opened 2023-08-18
enhancement: adding a flow name to deployment page crumbs

### LINKED ISSUE #10226: Include Flow name in Deployments title
### First check

### Prefect Version

2.x

Inside the Deployments menu, when a specific deployment is selected the title reads
`Deployments / <deployment_name>`.

This is a bit under-informative, because the deployment name must be seen in the context of the flow name that the user selected. So we have to glance a the side menu or list of flow runs to understand what we are actually looking at.

image

Include the flow name in the title, like `Deployments / <flow_name> / <deployment_name>`. (Preferred)

Or alternatively below the title similar to Flow runs:
image

The flow name may be removed from the sidebar or the list of flow runs.

### Example Use

_No response_

### Additional context

_No response_

WillRaphaelson:
closes https://github.com/PrefectHQ/prefect/issues/10226

### DIFF
--- a/src/components/PageHeadingDeployment.vue
+++ b/src/components/PageHeadingDeployment.vue
@@ -17,10 +17,11 @@
   import { DeploymentMenu, DeploymentToggle, PageHeading, RunMenu } from '@/components'
   import { useWorkspaceRoutes } from '@/compositions'
   import { useCan } from '@/compositions/useCan'
-  import { Deployment } from '@/models'
+  import { Deployment, Flow } from '@/models'
 
   const props = defineProps<{
     deployment: Deployment,
+    flow: Flow,
   }>()
 
   const can = useCan()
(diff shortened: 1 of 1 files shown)


src/components/PageHeadingDeployment.vue
```diff
@@ -17,10 +17,11 @@
   import { DeploymentMenu, DeploymentToggle, PageHeading, RunMenu } from '@/components'
   import { useWorkspaceRoutes } from '@/compositions'
   import { useCan } from '@/compositions/useCan'
-  import { Deployment } from '@/models'
+  import { Deployment, Flow } from '@/models'
 
   const props = defineProps<{
     deployment: Deployment,
+    flow: Flow,
```
pleek91, line 24: This works but means this is a breaking change. What we'd generally do is grab the flowId from the deployment and fetch it right within this component like `const { flow } = useFlow(() => props.deployment.flowId)`. 

That would also make this not a breaking change.

PR #1647, opened 2023-08-18
Enhancement: Add create and response api models for v2 concurrency limits

zhen0:
Adds the basic models for v2 concurrency. May need updating/adding to as we build out the feature.


pleek91 (approved):
Having V2 models is interesting. Do we anticipate having many V2 names things for this?

PR #1648, opened 2023-08-20
Enhancement: Adds API for Concurrency V2

zhen0:
Adds the mappers and api calls for concurrency V2


zhen0:
Note - the build is broken because of the demo api section - to keep the PR size down I'm adding that in #1650 - which fixes the build.  (If you prefer feel free to go straight to that one for review).

PR #1649, opened 2023-08-21
Enhancement: Add workspace routes for concurrency v2

zhen0:
Adds initial routes for the v2 concurrency page(s)


src/router/index.ts
```diff
@@ -247,6 +247,24 @@ export function createWorkspaceRouteRecords(components: Partial<WorkspaceRouteCo
...
+          path: '',
+          component: components.concurrencyLimitsV2,
+        },
+        {
+          name: 'workspace.concurrency-limits-v2.concurrency-limit',
+          path: 'concurrency-limit-v2/:concurrencyLimitId',
+          component: components.concurrencyLimitV2,
+        },
+      ],
+    },
```
znicholasbrown, line 267: Do we plan to keep v1 concurrency stuff around while v2 is released? If not, I don't think it makes sense to introduce these as "v2", we should replace existing components instead

znicholasbrown (approved):
I have a question that might be blocking but approving in case it's not

src/router/index.ts
zhen0: Yeah good question - for now I think we will need to keep the existing components around since users will need code changes to switch to v2.  The api is using the v2 convention but we may want to update this naming once we have a clearer idea of how v2 concurrency will work in the ui.

PR #1652, opened 2023-08-21
Cleanup unused css files

brandonreid:
Since removing the deprecated [PythonInput component](https://github.com/PrefectHQ/prefect-ui-library/blob/main/src/components/PythonInput.vue), I ran a find-all for these root class names and found nothing. I believe all instances that may have been are now replaced with `p-code-input`.


znicholasbrown (approved):
thanks @brandonreid

PR #1664, opened 2023-08-24
bump dependencies and fixes

brandonreid:


src/components/MarketingBanner.vue
```diff
@@ -74,31 +73,9 @@
   flex-shrink-0
 }
 
-.marketing-banner__background-image {
```
brandonreid, line 77: Removing these for now. Looks plain, but Karim is thinking through different motifs.

pleek91 (approved):

PR #1666, opened 2023-08-24
add pull steps to deployment well

WillRaphaelson:
add pull steps to deployment well

### DIFF
--- a/src/components/DeploymentDetails.vue
+++ b/src/components/DeploymentDetails.vue
@@ -77,6 +77,12 @@
       :alternate="alternate"
     />
 
+    <p-key-value v-if="deployment.pullSteps" label="Pull Steps" :alternate="alternate">
+      <template #value>
+        {{ deployment.pullSteps }}
+      </template>
+    </p-key-value>
+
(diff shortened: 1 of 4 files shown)


src/components/DeploymentDetails.vue
```diff
@@ -77,6 +77,12 @@
       :alternate="alternate"
     />
 
+    <p-key-value v-if="deployment.pullSteps" label="Pull Steps" :alternate="alternate">
+      <template #value>
+        {{ deployment.pullSteps }}
```
pleek91, line 82: How does this look when its rendered? `pullSteps` is an array so I'm guessing this might just look like json?

WillRaphaelson: yeah thats correct, it looks like this

image

pleek91: Could we use a p-code-highlight and show it as json? There's a `stringify` utility that will convert that array to formatted json.

WillRaphaelson:
for posterity the p-code didnt look good and craig resolved to refactor this into a tab in the near future

image

PR #1684, opened 2023-08-30
Resolve v2 feedback

brandonreid:


src/compositions/useOffsetStickyRootMargin.ts
pleek91, line 1: nit: I'd just return margin like `const { margin } = useOffsetStickyRootMargin()`.

pleek91: This composition also could use a doc string or something to explain what its for.

pleek91 (approved):

PR #1692, opened 2023-08-31
Enhancement: Updates to the v2 concurrency limits table

zhen0:
Adds slot decay and active fields to the v2 concurrency limits table and a toggle to set a limit to active/inactive

image


src/components/ConcurrencyLimitsV2Table.vue
```diff
@@ -6,10 +6,15 @@
...
       </p-link>
     </template>
-    <template #active-task-runs="{ row }">
+
+    <template #active-slots="{ row }">
       {{ row.activeSlots }}
     </template>
 
+    <template #active="{ row }">
+      <ConcurrencyLimitV2Toggle :limit="row" @update="concurrencyLimitSubscription.refresh()" />
```
pleek91, line 15: It looks like the toggle is automatically refreshing this subscription. Is this needed? I think this is going to cause the subscription to be refreshed twice.

pleek91 (approved):
One concern but otherwise LGTM!

PR #1694, opened 2023-08-31
Fix: Add truncate to log rows for task run names

zhen0:
Before:
image

After:
image
image

Resolves https://github.com/PrefectHQ/nebula-ui/issues/3170

### DIFF
--- a/src/components/LogRow.vue
+++ b/src/components/LogRow.vue
@@ -62,7 +62,7 @@
   shrink-0
   text-right
   pl-1
-  overflow-auto
+  truncate
 }
 
 .log-row__logger { @apply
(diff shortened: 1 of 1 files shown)


src/components/LogRow.vue
```diff
@@ -62,6 +62,7 @@
   shrink-0
   text-right
   pl-1
+  overflow-auto
```
brandonreid, line 65: I believe tailwind has a `truncate` utility for adding overflow ellipsis styles. If instead you want the user to be able to scroll this small area, then g2g, might get weird on devices with bulky scroll bars though.

brandonreid (approved):

pleek91 (approved):
Nice. Works for me

PR #1702, opened 2023-09-01
Remove descriptions from state tabs

cicdw:
This PR removes the descriptions at the top of the flow run lists on the dashboard - they added a small amount of clutter and we should only use sentences sparingly in the UI

Screenshot to validate my CSS on the accordion view:

image


src/components/FlowRunsAccordion.vue
```diff
@@ -56,4 +65,10 @@
 
     return flow
   }
-</script>
\ No newline at end of file
+</script>
+
+<style>
+.p-accordion__section:first-child .flow-runs-accordion-header:first-of-type { @apply
+  border-transparent
```
pleek91, line 72: Rather than turning it transparent let's remove it. I think it's border-0 or border-none

pleek91 (approved):
For context these descriptions let the user know what type of runs they can expect to see in each tab. Without these descriptions the tiny colors are the only indicator of what to expect which isn't very accessible. Generally only relying on colors or icons should be avoided. 

One suggestion about the border. But otherwise from a code perspective this looks fine.

cicdw:
Updated 👍 

That's a slippery slope though - the whole point of a UI is to present things with visual display; we have colors, state names, other visual cues, and even documentation to guide a user.  Maybe there's a case to be made for headers here, but complete sentences are bad headers.

Also, and maybe more importantly, that perspective is implicitly optimizing for education - but by definition, being a new user is a temporary situation and the majority of time spent in the UI will be by not-new users.  As such, no persistent and integrated feature should _ever_ be motivated for only a new user; every display should be equally valuable to a seasoned user.  More often than not descriptive text like this doesn't aid a seasoned user and at best takes up valuable real estate or at worst is a visual distraction.

PR #1703, opened 2023-09-01
Type -> Name -> Config

cicdw:
Small change to swap the order of steps 2 & 3 in work pool creation - there are many work pool types that don't require advanced config, and it can feel a little overwhelming to see all of that right after you pick a type.  So now it goes Type -> Name and Description -> Job Config


pleek91 (approved):
Assuming the order doesn't matter from a data/logic perspective.

cicdw:
Nope, the fields are all independent between the two steps :+1:

PR #1704, opened 2023-09-01
Expose initName as a prop for FlowRunCreateForm

cicdw:
Small PR that exposes a new optional prop on FlowRunCreateForm called `initName`; when provided, is set as the run name (and it can still be overwritten / randomly generated with the button).

We can leverage this in the last onboarding step to be suggestive of how someone might use this field, and I wouldn't be surprised if came up with other uses for it as well.

### DIFF
--- a/src/components/FlowRunCreateForm.vue
+++ b/src/components/FlowRunCreateForm.vue
@@ -109,7 +109,7 @@
     deployment: Deployment,
     // these must be the unmapped SchemaValues
     parameters?: SchemaValues,
-    initName?: string,
+    name?: string,
   }>()
 
   const generateRandomName = (): string => {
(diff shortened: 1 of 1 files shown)


src/components/FlowRunCreateForm.vue
```diff
@@ -109,6 +109,7 @@
     deployment: Deployment,
     // these must be the unmapped SchemaValues
     parameters?: SchemaValues,
+    initName?: string,
```
pleek91, line 112: Can we just call this prop "name"? We haven't used an "init" prefix anywhere else and the parameters prop does the same thing and doesn't have a prefix.

pleek91 (approved):
Requested a name change to the prop but code wise LGTM

src/components/FlowRunCreateForm.vue
cicdw: Ah I guess I thought it would conflict with the const called `name` but that doesn't make sense - good call, will update

cicdw: updated 👍

pleek91: Thanks!

PR #1705, opened 2023-09-01
Simplify flow run create form

cicdw:
This PR reduces the visual clutter when creating a run for a deployment:
- removes "General" as a heading
- only displays "Run Name" and "Parameters" by default (the most commonly figured fields)
- all other options (tags, retries, work queues, start times, message) are hidden under a new collapsable section called "Additional Options"

This simplifies the experience of creating a run for the majority of use cases while still preserving the full functionality of options for more complex situations.

Screenshot of the new default look:
image

Screenshot of the expanded view for a deployment with no work pool (couldn't capture everything):
image

As an added benefit, the third step in the onboarding tutorial looks nice and clean:
image

### DIFF
--- a/src/components/FlowRunCreateForm.vue
+++ b/src/components/FlowRunCreateForm.vue
@@ -1,11 +1,8 @@
 <template>
   <p-form class="flow-run-create-form p-background" @submit="submit">
     <p-content class="flow-run-create-form__section">
-      <h3 class="flow-run-create-form__section-header">
-        General
-      </h3>
 
-      <p-label label="Name (Optional)">
+      <p-label label="Run Name">
         <p-text-input v-model="name">
           <template #append>
             <p-button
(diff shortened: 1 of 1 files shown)


billpalombi:
I think the now/later toggle under "start" should be above the fold, not buried in additional options, otherwise I like these changes.

cicdw:
@billpalombi updated 👍

billpalombi (approved):
Love it

src/components/FlowRunCreateForm.vue
```diff
@@ -18,33 +15,23 @@
...
+              <p-button-group v-model="parametersInputType" :options="parametersInputTypeOptions" small />
+            </template>
+          </SchemaInput>
+        </p-content>
+      </template>
 
-      <h3 class="flow-run-create-form__section-header">
-        Start
-      </h3>
+      <h3>Start</h3>
```
pleek91, line 34: I feel like this doesn't need to be an `h3` anymore now that there isn't a general heading. Should we just wrap the `p-button-group` in a `p-label`? That would be more inline with the rest of the form

cicdw: Yea let me give that a shot

cicdw: So that makes the text here smaller, and @billpalombi had a request that it be sized the same as the `Parameters` heading (which is how I chose h3); open to achieving that however you think is best

pleek91: ah okay no that's fine leave it as is

pleek91 (approved):

PR #1710, opened 2023-09-01
Use slots and classes rather than targeting the accordion classes directly

pleek91:
# Description
The simplified run form looks great. But its directly targeting some classes from `PAccordion`. This means if another accordion shows up on the page at the same time (even in another feature) it'll get these styles. 

One way to solve that would be to add the `.flow-run-create-form` selector to each `.p-accordion` selector like `.flow-run-create-form .p-accordion__heading`.

But there was also a little css that we have components for to help us keep things standardized (`p-content` for spacing between things) and there is also a slot for the `heading` of an accordion section. So I also implemented those and removed the need for targeting accordion selectors at all.

### DIFF
--- a/src/components/FlowRunCreateForm.vue
+++ b/src/components/FlowRunCreateForm.vue
@@ -1,7 +1,6 @@
 <template>
   <p-form class="flow-run-create-form p-background" @submit="submit">
     <p-content class="flow-run-create-form__section">
-
       <p-label label="Run Name">
         <p-text-input v-model="name">
           <template #append>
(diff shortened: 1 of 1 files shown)


src/components/FlowRunCreateForm.vue
```diff
@@ -47,27 +46,34 @@
       </template>
 
       <p-accordion :sections="['Additional Options']">
+        <template #heading="{ section }">
```
cicdw, line 49: at some point I'd love to understand this syntax; I couldn't figure out in the flow runs accordion example how section / flowID / etc got unpacked

cicdw:
Awesome thank you for helping me better understand the patterns here @pleek91 - this makes sense!

PR #1711, opened 2023-09-01
Enhancement: Advanced to next step on selection of work pool type

cicdw:
This PR auto-advances to Step 2 of the work pool creation wizard once a user makes a selection of work pool type on Step 1; returning to Step 1 to update the selection is transparent and easy if the user made a mistake in their choice.


pleek91 (approved):
Nice, great little enhancement

PR #1712, opened 2023-09-03
Enhancement: Better error displays

cicdw:
This PR introduces a new utility for parsing errors returned from the API; we use the `details` field in many of our error handling responses as a way to communicate useful information to the user, but the UI has never displayed those messages to users.  This leads to confusing situations that put the user in a very difficult and hard to debug state.

With this PR, if the backend returns a `detail` field we will use and display it, otherwise we fall back to the current error messages that we've been displaying.

There are still failure modes for which this won't help, but we can handle many (though never all) of those in the backend with global error handlers that populate the `detail` field on the response.

If you want to see this in action, try to create a block or a work pool that starts with the name "prefect"

### DIFF
--- a/src/utilities/errors.ts
+++ b/src/utilities/errors.ts
@@ -1,19 +1,16 @@
-export function getErrorMessage(error: unknown, defaultErrorMessage: string): string {
-  let errorJson
+import axios from 'axios'
+import { isRecord, isString } from '@/utilities'
 
-  if (typeof error === 'string') {
-    try {
-      errorJson = JSON.parse(error)
-    } catch (parseError) {
-      // Error is not a valid JSON string
-      return defaultErrorMessage
-    }
-  } else if (typeof error === 'object' && error !== null) {
-    errorJson = error
-  } else {
+export function getErrorMessage(error: unknown, defaultErrorMessage: string): string {
+  if (!axios.isAxiosError(error)) {
     return defaultErrorMessage
   }
 
-  const detail = JSON.parse(errorJson?.request?.response ?? '')?.detail
-  return detail || defaultErrorMessage
+  const data = error.response?.data
+
+  if (isRecord(data) && isString(data.detail)) {
+    return data.detail
+  }
+
+  return defaultErrorMessage
 }
--- a/src/utilities/errors.ts
+++ b/src/utilities/errors.ts
@@ -1,7 +1,7 @@
 import axios from 'axios'
 import { isRecord, isString } from '@/utilities'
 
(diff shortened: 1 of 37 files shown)


src/utilities/errors.ts
pleek91, line 1: Are the error from the api sometimes strings and sometimes objects?

pleek91:
Nice, I like the simplicity of this. I wasn't able to get an error from the api in staging by creating a block that started with `prefect`. Is there some trick to that?

src/utilities/errors.ts
cicdw: Hmm good question - I might have overthought that logic; I guess they _should_ always be the same and I'd expect them to be an `object` but I thought I did see an instance of `string` when I was testing

cicdw:
Ah yea, my message was unclear - it's block _types_ that don't allow for a "prefect" prefix, not block documents.

src/utilities/errors.ts
pleek91: Okay so took me a while to figure this out cause I was really confused by this method. But finally got it sorted. Few things are going on here. 
1. Any error thrown from the api layer in the ui is going to come from axios and should never be a string. 
2. You can look at the request object and find the json value of the response (the `JSON.parse(errorJson?.request?.response)` part). But the parsed response can be found at `error.response.data`

So I believe this method could be written like
```
import axios from 'axios'
import { isRecord, isString } from '@/utilities'

export function getErrorMessage(error: unknown, defaultErrorMessage: string): string {
  if (!axios.isAxiosError(error)) {
    return defaultErrorMessage
  }

  const data = error.response?.data

  if (isRecord(data) && isString(data.details)) {
    return data.details
  }

  return defaultErrorMessage
}
```

pleek91: This also matches up with what I'm seeing if I try and create a work pool with the name "prefect"

cicdw: Gotcha that makes sense and is more direct; I can update in the next 24 hours - if you'd like to get this in sooner feel free to update my branch and merge also

cicdw: @pleek91 the new function definition didn't work for me for some reason - I still got the opaque "Failed to create work pool" message.  Because the function signature is stable would you be open to merging as-is and I can spend some time refining it when I can spend more time investigating?

pleek91 (approved):

src/utilities/errors.ts
pleek91: I had `details` (s) instead of `detail`.

pleek91:
Cleaned this up a little per our conversation

znicholasbrown (approved):
wonderful

PR #1714, opened 2023-09-05
Adds worker status to worker table

desertaxle:
- Adds display of worker status to worker table for a given work pool
- Updates worker last heartbeat time to be displayed as a relative time

### Examples
Screenshot 2023-09-05 at 3 12 11 PM

Screenshot 2023-09-05 at 3 13 19 PM


src/components/WorkerStatusBadge.vue
```diff
@@ -0,0 +1,32 @@
...
+
+  const props = defineProps<{
+    worker: WorkPoolWorker,
+  }>()
+
+  const workerStatus = computed(() => props.worker.status)
+
+  const label = computed(() => {
+    return capitalize(props.worker.status)
+  })
```
pleek91, line 24: nit: I'm trying to reduce the number of computed properties we're using because they have some overhead associated with them. This is fine as is but you could also pass `worker.status` directly into `WorkerStatusIcon` and add `capitalize(worker.status)` directly to the template rather than using computed properties.

src/components/WorkerStatusBadge.vue
```diff
@@ -0,0 +1,32 @@
...
+</script>
+
+<style>
+.worker-status-badge { @apply
+  text-xs
+  pl-1
+}
```
pleek91, line 31: @brandonreid should we be customizing tag styles? I assume this is fine

src/components/WorkersTable.vue
```diff
@@ -11,8 +11,12 @@
         <span>{{ row.name }}</span>
       </template>
 
-      <template #last-polled="{ value }">
-        <span>{{ formatDateTimeNumeric(value) }}</span>
+      <template #last-seen="{ value }">
+        <span>{{ formatDateTimeRelative(value) }}</span>
```
pleek91, line 15: We do have a composition called `useNow` which you can use with `formateDateTimeRelative` to get a dynamically updating time. We used that for work pools [here](https://github.com/PrefectHQ/prefect-ui-library/blob/main/src/compositions/useWorkPoolLastPolled.ts#L53)

Might be cool to do the same here.

pleek91 (approved):
Few super minor suggestions but overall looks good!

src/components/WorkerStatusBadge.vue
desertaxle: I removed the custom tag styling since it made the badges a little easier to read.

pleek91 (approved):

PR #1715, opened 2023-09-05
Maintenance: Update empty states to match cloud

zhen0:
https://github.com/PrefectHQ/nebula-ui/pull/3251 updated the copy in flow, flow run and deployment empty states in cloud.  This PR brings the changes into the ui library so OSS and Cloud language is consistent.  It also lays the way for updating Cloud to use ui-library empty states and remove duplicate components.


pleek91 (approved):
Thank you!

PR #1717, opened 2023-09-06
Add display of work pool status

desertaxle:
Adds display of work pool status in the following locations:
- Work pool details
- Work pool card
- Dashboard work pool card

The work pool status icon has a tooltip that displays an explanation for each status:
- **Ready:** Work pool has at least one online worker ready to execute work.
- **Not ready:** Work pool does not have any online workers ready to execute work.
- **Paused:** Work pool is paused. No work will be executed.

#### Work pool details

Screenshot 2023-09-05 at 9 24 04 PM

#### Work pool cards

Screenshot 2023-09-05 at 9 24 24 PM

#### Dashboard work pool cards

Screenshot 2023-09-05 at 9 24 46 PM


src/components/WorkPoolStatusIcon.vue
```diff
@@ -0,0 +1,66 @@
+<template v-if="status">
```
pleek91, line 1: 1. Is this v-if is needed?
2. Does this work on the root template? If so TIL

src/components/WorkPoolDetails.vue
```diff
@@ -1,5 +1,16 @@
 <template>
+  <!-- NEED WORK POOL BADGE AND ICON -->
+
+
```
pleek91, line 4: ^

src/components/WorkPoolStatusIcon.vue
desertaxle: Yep, the `v-if` is needed because push work pools will not display a status because the user won't need to host any workers. Push work pools will get a status of `null` from the API.

That `v-if` didn't work on the root template, so I moved it to the `p-tooltip`. Thanks for the catch!

pleek91: aw man, I was hoping it worked on that root template tag!

pleek91 (approved):

PR #1719, opened 2023-09-07
Fix: Simplify and correct global concurrency toggle logic

zhen0:


src/components/ConcurrencyLimitV2Toggle.vue
```diff
@@ -42,7 +42,7 @@
     loading.value = true
 
     try {
-      await api.concurrencyV2Limits.updateConcurrencyV2Limit(props.limit.id, { active: !value })
+      await api.concurrencyV2Limits.updateConcurrencyV2Limit(props.limit.id, { active: value })
```
pleek91, line 45: This part doesn't make sense to me. Pressing the toggle should trigger the set with the correct value shouldn't it?

pleek91 (approved):

PR #1728, opened 2023-09-08
Enhancement: Expose state-list-item default slot in flow run list item

znicholasbrown:
Invisible change, only used by downstream apps


brandonreid (approved):
Feels like `after` should do the job? Don't totally see the need but trust there is one.

pleek91:
Also wondering why this would be needed

znicholasbrown:
Ah you know what, if the default slot isn't flexed the `after` slot works perfectly for this, I was confused about how that was being used

PR #1739, opened 2023-09-12
add an opt in to render large flow run graphs

### LINKED ISSUE #10610: Prefect UI flow run digram with large number of tasks is unresponsive
### First check

### Prefect Version

2.x

The current implementation of the Prefect UI flow run diagrams is causing performance issues when dealing with a large amount of data. As the diagram loads, it slows down the browser and may even cause it to become unresponsive or freeze. 

Requiring the user to click a button before displaying the flow run diagram if it has maybe over N number of tasks.

### Example Use

_No response_

### Additional context

_No response_

brandonreid:
Resolves https://github.com/PrefectHQ/prefect/issues/10610

If a graph has more than 2k nodes, prevent the graph from drawing unless the user opts in to prevent potentially bogging down and browser crashing.

image

### DIFF
--- a/src/components/FlowRunTimeline.vue
+++ b/src/components/FlowRunTimeline.vue
@@ -7,7 +7,7 @@
     aria-label="Flow run timeline graph"
     :style="{ height }"
   >
-    <div class="flow-run-timeline__wrapper">
+    <div v-if="!requireLargeRenderConfirmation" class="flow-run-timeline__wrapper">
       <p-button
         v-if="isFullscreen"
         class="flow-run-timeline__fullscreen-exit"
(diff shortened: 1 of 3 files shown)


pleek91 (approved):
Nice enhancement! Might be nice to move that confirmation into its own component just to keep this already large component from getting to big. But overall this looks good!

brandonreid:
> Nice enhancement! Might be nice to move that confirmation into its own component just to keep this already large component from getting to big. But overall this looks good!

@pleek91 I went ahead and fallowed your advice there. Will merge since I'm just moving it to another file.

PR #1745, opened 2023-09-13
add ability to filter FlowList by has deployment or not

brandonreid:
Related to https://github.com/PrefectHQ/prefect/issues/9517

I temporarily enabled this screen and linked it up locally for testing in staging:

https://github.com/PrefectHQ/prefect-ui-library/assets/6776415/0bc7e0f2-9214-42c3-9542-b859efdac554


pleek91:
I could be wrong but isn't this page turned off in prod? Like don't users see the flows table still?

brandonreid:
> I could be wrong but isn't this page turned off in prod? Like don't users see the flows table still?

In Nebula, that's correct, but in OSS we actually use this page and still have the separate Deployments page. Idk why exactly.. This functionality should probably be in Nebula as well though, now that you mention it.. So I won't call this ticket resolved quite yet.

znicholasbrown:
That screen should probably be hidden altogether in OSS in favor of the existing deployments/flows since it's been removed from the cloud env; if we're keen to keep it and to introduce this functionality, I think the label needs updating; "Deployments" make it seem like you're filtering deployments when you're really filtering out flows that have / don't have deployments, which i don't think is altogether clear

pleek91:
Oh weird, I assumed when we reverted this in cloud we also revered it in OSS. That feels very off if I upgrade from server to cloud and now my flows page is significantly different.

zhen0:
@brandonreid - do we still intend to make this happen?  Or should we close this PR?

brandonreid:
> @brandonreid - do we still intend to make this happen? Or should we close this PR?

Let's close it. We've got plans to combine the two versions into one better one. 👍

PR #1747, opened 2023-09-14
Enhancement: Deployment combobox searches and paginates via the api

pleek91:
# Description
Refactors the DeploymentCombobox component to hit the api for filtering the results. Also paginates if you scroll past the first 50 results.


src/components/UseDeploymentSlot.vue
pleek91, line 1: I'm hesitant to introduce this. Components like this often get over used and make the templates of other components larger and more complex when there should really be a second component. But I literally just need to display the `{{ deployment.name }}` for a given deployment and seemed like a waste to make a component to just do that.

src/compositions/useDeploymentsInfiniteScroll.ts
pleek91, line 1: Im planning on abstracting this into a factory in a future PR. There is very little about this that is deployment specific.

pleek91: This is a much better version/pattern from what exists in [useFlowRunInfiniteScroll](https://github.com/PrefectHQ/prefect-ui-library/blob/main/src/compositions/useFlowRunsInfiniteScroll.ts). I originally copied that here but then ran into a few bugs.

src/maps/filters.ts
```diff
@@ -57,9 +57,9 @@ function toIsNull(value?: boolean): IsNull | undefined {
   return { is_null_: value }
 }
 
-function toLike(value?: string): Like | undefined {
-  if (typeof value === 'undefined') {
-    return value
+function toLike(value?: string | null): Like | undefined {
+  if (!value) {
+    return undefined
```
pleek91, line 62: I noticed that when binding a `_like` value to one of our inputs we can end up with both `null` and `''` as empty values. This change makes sure no falsey values are sent to the api which makes for more accurate subscription caching.

src/utilities/arrays.ts
```diff
@@ -120,7 +120,7 @@ export function asSingle<T>(value: T | T[]): T {
   return value
 }
 
-export function repeat<T>(count: number, method: (index?: number) => T): T[] {
+export function repeat<T>(count: number, method: (index: number) => T): T[] {
```
pleek91, line 123: Just fixing the type here.

brandonreid (approved):
That `UseDeploymentSlot` component is a pretty slick pattern. Lgtm!

znicholasbrown (approved):

PR #1748, opened 2023-09-15
Enhancement: Schema fields without positions should sink

znicholasbrown:
Previously we were giving all schema properties without position keys a position of `0`, which would place them before any ordered properties. This PR changes that behavior to move them to the bottom instead.


pleek91 (approved):
Nice use of `Infinity`!

PR #1749, opened 2023-09-15
Fix flows don't trigger custom filter options.

collincchoy:
# Description
This looks like a bug to me. If this is actually intended then feel free to close and ignore. 

The functionality to filter by flow(s) _works_ and even selecting pre-saved filters that included flows correctly sets the filter, but what doesn't is being able to save new filters using **only** flows and the UI doesn't correctly map pre-saved filters that include flows so when selected a saved filter that includes flows, the SavedFilter select shows "Custom" instead of the selected filter. 

## Steps to reproduce:
1. Go to Flow Runs
image

2. Select at least 1 flow to filter by

| Actual | Expected |
| --- | --- |
| image | image |

3. Save a filter that includes flows and then select that filter. Notice the filtering _works_ but the dropdown shows "Custom" instead of the filter you just selected

| Actual | Expected |
| --- | --- |
| image | image |

This also means I can't save filters that only include flows
| Actual | Expected |
| --- | --- |
| image | image |

### DIFF
--- a/src/components/SavedFilters.vue
+++ b/src/components/SavedFilters.vue
@@ -43,7 +43,7 @@
     get() {
       const inRoute: SavedSearchFilter = {
         state: filter.flowRuns.state.name ?? [],
-        flow: filter.flows.name ?? [],
+        flow: filter.flows.id ?? [],
         deployment: filter.deployments.id ?? [],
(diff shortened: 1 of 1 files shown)


pleek91 (approved):
Nice find! TY!

PR #1751, opened 2023-09-18
Enhancement: Abstract api pagination into a usePagination composition

pleek91:
# Description
Building on top of the `useDeploymentsInfiniteScroll` composition (which is deleted in this PR) to create an abstracted `usePagination` composition which supports both `page` (one page at a time) and `infinite` (infinite scroll) pagination. 

## The `usePagination` composition
This composition works on its own but is intended to be wrapped by a entity specific composition. So `useDeployments` calls `usePagination`. 

`usePagination` accepts
- `fetchMethod` - A method that returns `Promise<any[]>`
- `fetchParameters` - A method that returns an array of parameters for the `fetchMethod`
- `countMethod` - A method that returns `Promise<number>` to get the total count of values that could be returned by the `fetchMethod`
- `countParameters` - A method that returns an array of parameters for the `countMethod`
- `options` - `SubscriptionOptions & { mode: 'page' | 'infinite' }`

## Implementation
See `useDeployments` for an example of implementation. The final devx makes this possible
```typescript
const { 
  deployments,
  total,
  page,
  pages, 
  next, 
  previous, 
  subscriptions 
} = useDeployments(filter, { mode: 'infinite' })
```

### Next Steps
This pattern will be carried over to the other `useThings` compositions (flows, flow runs, etc)


src/compositions/useSubscriptions.ts
pleek91, line 1: This change is just to allow `useSubscriptions` to accept subscriptions of different types and give you an accurate return type. Previously we've only used this composition on multiple subscriptions of the same type.

src/compositions/usePagination.ts
```diff
@@ -0,0 +1,189 @@
...
+
+  const countSubscriptionParameters = computed(() => {
+    if (page.value) {
+      return toValue(countParametersGetter)
+    }
+
+    return null
+  })
+  const countSubscription = useSubscriptionWithDependencies(countMethod, countSubscriptionParameters, options)
+  const total = computed(() => countSubscription.response ?? 0)
```
pleek91, line 88: Worth pointing out that because this composition returns `pages` and `total` you end up hitting the `countMethod` even if you never use that count.

src/components/DeploymentCombobox.vue
pleek91, line 1: This component is an example of where we previously used a special infinite scroll composition but now we can just use the normal `useDeployments` composition and set the `mode` to `infinite`.

src/components/DeploymentList.vue
pleek91, line 1: This component is an example of where we were previously using two separate compositions to get and count deployments as well having custom offset logic. Much simpler using the updated `useDeployment` composition which includes the pagination information by default.

src/components/DeploymentTagsInput.vue
```diff
@@ -35,5 +35,5 @@
   })
 
   const { deployments } = useDeployments(filter)
-  const tags = computed(() => unique(deployments.value?.flatMap(deployment => deployment.tags ?? []) ?? []))
+  const tags = computed(() => unique(deployments.value.flatMap(deployment => deployment.tags ?? [])))
```
pleek91, line 38: Always returning an array cleans this up too

src/compositions/useDeployments.ts
```diff
@@ -1,55 +1,45 @@
...
-export type UseDeployments = UseEntitySubscription<WorkspaceDeploymentsApi['getDeployments'], 'deployments'>
+export type UseDeployments = UsePaginationEntity<
+WorkspaceDeploymentsApi['getDeployments'],
+WorkspaceDeploymentsApi['getDeploymentsCount'],
+'deployments'
+>
 
-export function useDeployments(filter: MaybeRefOrGetter<DeploymentsFilter>): UseDeployments
-export function useDeployments(deploymentIds: MaybeRefOrGetter<string[] | null | undefined>): UseDeployments
-export function useDeployments(filterOrDeploymentIds?: MaybeRefOrGetter<string[] | DeploymentsFilter | null | undefined>): UseDeployments {
```
pleek91, line 15: Decided to remove the overload here since we weren't using it anywhere. Its also simple enough to use a getter for this at implementation like
```typescript
useDeployments(() => ({
 deployments: {
    id: ['one', 'two']
  }
})

File: src/compositions/usePagination.ts [hunk 4]
```diff
@@ -0,0 +1,202 @@
...
+type PaginationParameters = [filter?: PaginationFilter, ...any[]]
+type PaginationFetchAction = (...parameters: PaginationParameters) => Promise<unknown[]>
+type PaginationCountAction = (...parameters: PaginationParameters) => Promise<number>
+
+type FetchSubscriptionAction<
+  TFetch extends PaginationFetchAction
+> = (parameters: Parameters<TFetch>[]) => Promise<Awaited<ReturnType<TFetch>>>
+
+export type PaginationOptions = SubscriptionOptions & {
+  mode?: 'page' | 'infinite',
```
znicholasbrown, line 24: The `infinite` piece here is in my mind purely part of the UX and not really a difference in the way results are fetched; outside of changing the way results are given back to the caller (flat map + result building) is there anything stopping us from merging these two modes? That would simplify this code in a few places

znicholasbrown (commented):
A couple of questions, overall I like this direction a lot! Happy to see this abstraction coming together

src/compositions/usePagination.ts
```diff
@@ -0,0 +1,202 @@
...
+  page?: MaybeRef<number>,
+}
+
+export type UsePaginationParameters<
+  TFetch extends PaginationFetchAction,
+  TFetchParameters extends Getter<Parameters<TFetch> | null>,
```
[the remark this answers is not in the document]
pleek91: I went that route originally because it was just simpler. But that might have been an issue with a previous iteration. 

I feel like using this composition with a static filter isn't super useful. But we could allow for it if needed.

src/compositions/usePagination.ts
pleek91: When implementing only 1 of 3 places in this PR used the count. Just for reference.

src/compositions/usePagination.ts
```diff
@@ -0,0 +1,202 @@
...
+  function next(): void {
+    if (mode === 'page') {
+      page.value++
+    }
+
+    const shouldLoadNextPage = page.value * getLimit() <= results.value.length
+
+    if (shouldLoadNextPage) {
+      page.value++
+    }
```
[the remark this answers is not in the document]
pleek91: I think I'm missing a return in that first if statement. Good catch

src/compositions/usePagination.ts
pleek91: Um `page` returns just the single page of results. Vs `infinite` returns all the pages. 

Also `page` mode starts at page 1 (immediately fetched results) vs `infinite` starts at page 0 and lets the implementation dictate when to fetch results. 

So I don't think they can be combined. Might be missing something

src/compositions/usePagination.ts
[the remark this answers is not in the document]
pleek91: Fixed

znicholasbrown (approved):

src/compositions/usePagination.ts
pleek91: Discussed async.

PR #1753, opened 2023-09-18
Add a user-customizable default flow runs filter (read)

collincchoy:
This PR alters `<SavedFilters />` to automatically set the _"default"_ flow runs filter from a pre-configured value from local storage. 

Here _"default"_ is interpreted as "no selected filter" a.k.a. no filter query params. 

A follow-up PR will add the ability to save a filter to this new localstorage slot for use here. As-is, changes in this PR should have no notable user-facing effects without manual localstorage manipulation.


src/compositions/filters.ts
```diff
@@ -631,6 +632,12 @@ export function useRecentFlowRunsFilterFromRoute(defaultValue: MaybeReactive<Flo
   return response
 }
 
+const customDefaultFlowRunsFilterKey = getCacheKey('prefect-ui-library-custom-default-flow-runs-filter')
+
+export function useCustomDefaultFlowRunsFilter(): ReturnType<typeof useLocalStorage<FlowRunsFilter | null>> {
+  return useLocalStorage<FlowRunsFilter>(customDefaultFlowRunsFilterKey)
+}
+
```
pleek91, line 640: nit: this seems specific to the flow runs page vs the rest of this file which is generic of implementation. I'd move this to a new file.

src/utilities/savedFilters.ts
```diff
@@ -28,6 +28,11 @@ export function isSameFilter(filterA: SavedSearchFilter, filterB: SavedSearchFil
   return true
 }
 
+export function isEmptyFilter(filter: SavedSearchFilter): boolean {
+  const emptyFilters = { state: [], flow: [], deployment: [], workPool: [], tag: [] } satisfies SavedSearchFilter
+  return isSameFilter(filter, emptyFilters) && filter.startDate === undefined && filter.endDate === undefined
```
pleek91, line 33: I wonder why `isSameFilter` doesn't check the dates 🤔

src/components/SavedFilters.vue
```diff
@@ -39,17 +39,26 @@
...
+    startDate: filter.flowRuns.expectedStartTimeAfter != undefined ? String(filter.flowRuns.expectedStartTimeAfter) : undefined,
+    endDate: filter.flowRuns.expectedStartTimeBefore != undefined ? String(filter.flowRuns.expectedStartTimeBefore) : undefined,
+  }))
+
+  const { value: myCustomDefaultFilter } = useCustomDefaultFlowRunsFilter()
+  watch(filterInRoute, (newValue) => {
+    if (myCustomDefaultFilter.value !== null && isEmptyFilter(newValue)) {
+      setFilters(myCustomDefaultFilter.value)
+    }
+  }, { immediate: true })
```
pleek91, line 57: Won't this mean a user can never get back the the prefect defined default view of one week? Maybe that's intended. 

I'm wondering if it would be better to preserve the options in the menu but `onMount` check if there is a user defined default and select that. But then let the user choose the "Default view" if they wanted to. 

Maybe we should change "Default View"  to show up as something else 🤔 its just a filter for the last week. Maybe we should label it as such.

collincchoy: it's not totally explicit, but there's a slight distinction between _selecting_ the "default view" and navigating to `/flow_runs`. On navigation to `/flow_runs`, the `isSameFilter` ignores checking the start/end dates in the route and calls it "default". On selecting "default view", the route gains the implicit start/end dates that define the oneWeekSavedSearch default filter. 

I'm taking advantage of that so that an empty route maps to a custom default filter if one exists else fallback to current behavior. Otherwise, the "default view" option is still pickable since it has its own route query params it adds. 

> Maybe we should change "Default View" to show up as something else 🤔 its just a filter for the last week. Maybe we should label it as such.

Yeah I'm toying with that right now. atm thinking of maybe a default tag or just some `(default)` kind of text on the current default option and replacing what currently reads as "default view" as "Last week" or something.

pleek91: sounds good, thanks for clearing that up 👍

src/components/SavedFilters.vue
```diff
@@ -39,17 +39,26 @@
...
+    workPool: filter.workPools.name ?? [],
+    tag: filter.flowRuns.tags.name ?? [],
+    startDate: filter.flowRuns.expectedStartTimeAfter != undefined ? String(filter.flowRuns.expectedStartTimeAfter) : undefined,
+    endDate: filter.flowRuns.expectedStartTimeBefore != undefined ? String(filter.flowRuns.expectedStartTimeBefore) : undefined,
+  }))
+
+  const { value: myCustomDefaultFilter } = useCustomDefaultFlowRunsFilter()
+  watch(filterInRoute, (newValue) => {
+    if (myCustomDefaultFilter.value !== null && isEmptyFilter(newValue)) {
+      setFilters(myCustomDefaultFilter.value)
```
pleek91, line 55: This `setFilters` does a `router.push` under the hood. We probably want to find a way to do a `router.replace` for this so we don't end up with double history

src/utilities/savedFilters.ts
collincchoy: Yeah I updated that at first. my best guess is the way it's written atm leaves a convenient way for the `isSameFilter(oneWeekSavedSearch, emptyFilter) == true`. but since I'm changing that a little... it _might_ make sense to just extend the isSameFilter? 

🤔

src/components/SavedFilters.vue
collincchoy: 👀 
ah yeah that feels like having a custom default as-is would probably get you into a bit of a loop trying to click back

src/utilities/savedFilters.ts
pleek91: Could be a nice improvement. Your explanation sounds right.

src/components/SavedFilters.vue
collincchoy: ah I did try adding a beforeEnter/nav guard to basically do that. that... worked but this seemed maybe a little better. 

are you thinking like [here](https://github.com/PrefectHQ/prefect-ui-library/blob/0fa90a173b1abdcde6722cdc6097e1b105e2f019/src/router/routes.ts#L13-L14)?

pleek91: Correct, so anywhere that method is called it returns a url with the filters already. I THINK since you're storing the filters themselves in local storage you could grab and apply them just fine. Just an idea, I don't have much opinion here.

collincchoy: Hmmm iirc I also went away from the nav guard b/c the guard didn't rerun on query-param-only route changes so clicking on Flow Runs from the nav menu after already having filters wouldn't trigger my custom default. I think this suggestion would fix that but on the other side navigating directly to /flow_runs somehow would probably circumvent a check here? 

oh also I was struggling to figure out how to go from `const filter: FlowRunsFilter` -> `router.push({..., query: dotFlatten(filter))` w/o the `useRouteQueryParams` magic that's under the hood of the `setFilters` here

pleek91: Hmm right, yeah that wouldn't be as simple as I'm I was thinking because there isn't a utility for converting the filters to just query params. 

Might be a nice enhancement to that stuff. I'll look into that if you think it would be helpful here

src/utilities/savedFilters.ts
collincchoy: think I'm gonna leave this as-is for now so as to not break existing functionality. my brain is rolling over a little trying to figure out how to best fold startdate/enddate into `isSameFilter` w/o breaking the current behavior where the dates are ignored and either an empty filter _or_ the explicit filter match the current default view.

src/components/SavedFilters.vue
collincchoy: Thinking of kicking the can on this for now - not ideal but breaking back just for user-defined custom view for _now_ seems _okayyyy_. It'd be _neat_ if `setFilters` allowed me to just overwrite (which would under-the-hood replace) and looked into the route query compositions and seems non-trivial. 

alternatively hooking into router configurations here feels less correct to me than having the `<SavedFilter />` own the responsibility of having/setting a custom, default saved filter. 

may address in a follow-up PR.

collincchoy: resolved in [8dfa717](https://github.com/PrefectHQ/prefect-ui-library/pull/1753/commits/8dfa717157135268ce8ae21f056db6e94cfdf1af)

package.json
```diff
@@ -64,8 +64,8 @@
     },
     "peerDependencies": {
         "@prefecthq/prefect-design": "^2.0.8",
-        "@prefecthq/vue-charts": "^2.0.2",
-        "@prefecthq/vue-compositions": "^1.5.9",
+        "@prefecthq/vue-charts": "^2.0.3",
+        "@prefecthq/vue-compositions": "^1.6.2",
```
collincchoy, line 68: I'm using a new utility from vue-compositions@1.6.2 so ui-library now requires a minimum version of 1.6.2

separately but related, vue-charts@2.0.2 had peerDependencies: { @prefecthq/vue-compositions: 1.5.9 } pinned exactly whereas 2.0.3 does a looser version constraint so I'm going ahead and bumping that here as well. 🤔 

Recommend checking out the package-lock.json changes - they're pretty minimal but noting that they're there.

package-lock.json
```diff
@@ -1326,10 +1326,32 @@
...
+            },
+            "peerDependenciesMeta": {
+                "lodash.debounce": {
+                    "optional": true
+                },
+                "vue-router": {
+                    "optional": true
+                }
+            }
+        },
```
collincchoy, line 1350: 🤨

package-lock.json
```diff
@@ -8191,12 +8213,23 @@
...
+            "dependencies": {
+                "@prefecthq/vue-compositions": {
+                    "version": "1.5.9",
+                    "resolved": "https://registry.npmjs.org/@prefecthq/vue-compositions/-/vue-compositions-1.5.9.tgz",
+                    "integrity": "sha512-2W5bOxcLTvl+EjsEMyB7yj1Hp4ElebvHjlF0zHBYSlzSQZ6Bh3jkKoDjVVz8RfxkQ1+e25M/B7ZXjzV4T/NPNA==",
+                    "peer": true,
+                    "requires": {
+                        "jsdom": "^22.0.0"
+                    }
+                }
```
collincchoy, line 8226: 🤨

collincchoy (commented):
Hm don't these blocks seem strange?

package-lock.json
collincchoy: Oh this one is for prefect-design. 
image

feel like that one's probably fine.

pleek91 (approved):

PR #1759, opened 2023-09-19
Enhancement: Update FlowCombobox to search and paginate via the api

### LINKED ISSUE #9805: All the flows not being shown under the filter selection in UI when large number of flows exist
### First check

- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

It seems that the flows filter in Flow Runs page has a limit of 50 flows in it and are sorted in ascending order.
If we have more that 50 flows deployed, we are unable to select the 51th flow in the drop down selection.

Even typing in the name of the flow in the search bar does not bring up the flow.

### Reproduction

```python3
1. Deploy over 50 flows; name not starting with `z`
2. Deploy another flow; name starting with `z`
3. On the UI, move to "/flow-runs" page
4. Attempt to filter to the flow which had the name starting with `z`
```

### Error

_No response_

### Versions

```Text
Version:             2.10.7
API version:         0.8.4
Python version:      3.10.8
Git commit:          5869f3fe
Built:               Thu, May 4, 2023 2:24 PM
Profile:             local
Server type:         server
```

### Additional context

_No response_

pleek91:
# Description
Same as https://github.com/PrefectHQ/prefect-ui-library/pull/1747 did for deployments this updates the FlowCombobox to use the api

Closes https://github.com/PrefectHQ/prefect/issues/9805


src/compositions/usePagination.ts
```diff
@@ -184,7 +184,14 @@ export function usePagination<
...
   watch(fetchParametersGetter, () => {
-    page.value = 1
+    const newPageValue = 1
+    const manuallySetParameters = page.value === newPageValue
+
+    page.value = newPageValue
+
+    if (manuallySetParameters) {
+      setFetchSubscriptionParameters()
+    }
```
pleek91, line 194: I noticed that if the page value doesn't actually change the parameters would not get set because the page watcher wouldn't get triggered. This fixes that

znicholasbrown: i'm surprised this is needed for something so simple but i don't have any other suggestions

znicholasbrown (approved):

src/compositions/usePagination.ts
pleek91: I believe its because vue doesn't rigger effects when assigning the same value and since `1 === 1` it doesn't trigger the watcher.

PR #1760, opened 2023-09-20
Maintenance: Add the ability to set a default sort other than DESC for the flow list

zhen0:
Currently the flow run filtered list sets the default as 'START_TIME_DESC'. This PR adds the ability to pass a prop to the list so that a developer can set a different default.  This would allow us to set the scheduled flow runs list to have a default of ''START_TIME_ASC' and so show imminent runs first.

### DIFF
--- a/src/components/FlowRunFilteredList.vue
+++ b/src/components/FlowRunFilteredList.vue
@@ -40,6 +40,7 @@
     flowRunFilter: FlowRunsFilter,
     states?: PrefectStateNames[],
     disableDeletion?: boolean,
+    sort?: FlowRunSortValues,
   }>()
 
   const emit = defineEmits<{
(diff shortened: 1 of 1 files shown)


src/components/FlowRunFilteredList.vue
```diff
@@ -57,7 +58,7 @@
     emit('update:states', states.value)
   }
 
-  const sort = ref<FlowRunSortValues>('START_TIME_DESC')
+  const sort = ref<FlowRunSortValues>(props.sort ?? 'START_TIME_DESC')
```
pleek91, line 61: Rather than adding an additional prop. We should just grab the sort from the filter since sort is already part of that type. Otherwise it could get confusing if you have a sort on your filter and it gets ignored.

zhen0: Yeah good point.  I think your point about potentially splitting these out into a new component makes sense!

PR #1762, opened 2023-09-20
Feature: Users can set custom default flow runs filter

collincchoy:
## Description
This PR adds onto #1753 and allows a user to set/unset("remove") saved filters as their custom, default filter.

## 📸 Screenshots
| **Before** | **After** |
| ---- | ---- |
| image | image |

| Flow to change my default filter | |
| --- | --- |
| Pick my desired filter | image |
| In the kebab menu, pick the new "Set as default" option | image |
| It now says ... (default) | image |
| To remove/unset it* | image |
| **UPDATED** What if someone else deleted the filter I have set as the default? | image |

\* Instead of "removing" which will set the default back to the system default (Past week), you can also just pick another filter and "Set as default". For example, manually set "Past week" as default again.

## Issues
This is to support https://github.com/PrefectHQ/prefect/issues/10540 downstream


src/components/SavedFilters.vue
```diff
@@ -52,7 +52,15 @@
...
 
   const router = useRouter()
-  const { value: myCustomDefaultFilter } = useCustomDefaultFlowRunsFilter()
+  const { value: mySavedCustomDefaultFilter, asFlowRunsFilter: myCustomDefaultFilter } = useCustomDefaultFlowRunsFilter()
+  const nameOfDefaultFilter = computed(() => {
+    const customDefault = mySavedCustomDefaultFilter.value
+    if (!savedSearches.value.length || !customDefault) {
+      return systemDefaultSavedSearch.name
+    }
+    // TODO: handle case where customDefault is not in savedSearches
```
collincchoy, line 61: I'm thinking of prompting a user that their saved default filter might not exist anymore - [clear it] or [save it as a new filter]. 

Need to think through where/how to do that w/o just introducing a side effect into this computed ref.

src/maps/index.ts
```diff
@@ -127,7 +127,7 @@ export const maps = {
   RunHistory: { FlowRunHistoryResponse: mapRunHistoryToFlowRunHistoryResponse, DivergingBarChartItem: mapRunHistoryToDivergingBarChartItem },
   SavedSearchCreate: { SavedSearchCreateRequest: mapSavedSearchCreateToSavedSearchCreateRequest },
   SavedSearchesFilter: { SavedSearchesFilterRequest: mapSavedSearchesFilter },
-  SavedSearchFilter: { FlowRunFilters: mapSavedSearchFilterToFlowRunFilters },
+  SavedSearchFilter: { FlowRunsFilter: mapSavedSearchFilterToFlowRunsFilter },
```
collincchoy, line 130: oh sweet maps.index.ts typos, how we love you so. 🎵

src/components/SavedFilters.vue
collincchoy: eh. popping a toast after deleting from localstorage for now.

src/components/SavedFiltersMenu.vue
```diff
@@ -32,11 +42,13 @@
...
   import { useCan } from '@/compositions/useCan'
+  import { useCustomDefaultFlowRunsFilter } from '@/compositions/useCustomDefaultFlowRunsFilter'
+  import { SavedFlowRunsSearch } from '@/compositions/useSavedFlowRunsSearches'
   import { SavedSearch } from '@/models/SavedSearch'
-  import { customSavedSearch } from '@/utilities/savedFilters'
+  import { customSavedSearch, systemDefaultSavedSearch } from '@/utilities/savedFilters'
 
   const props = defineProps<{
-    savedSearch: SavedSearch | null,
+    savedSearch: SavedFlowRunsSearch | null,
```
collincchoy, line 51: This type change just adds an `isDefault` property onto the object. trying to keep `SavedSearch` as a one-to-one api/backend entity and then this composition wraps that to include the (atm) FE-only user-customizable default and to also be more context-aware as saved _flow runs_ search.

src/compositions/useCustomDefaultFlowRunsFilter.ts
```diff
@@ -1,8 +1,27 @@
...
 const customDefaultFlowRunsFilterKey = getCacheKey('prefect-ui-library-custom-default-flow-runs-filter')
 
-export function useCustomDefaultFlowRunsFilter(): ReturnType<typeof useLocalStorage<FlowRunsFilter | null>> {
-  return useLocalStorage<FlowRunsFilter>(customDefaultFlowRunsFilterKey)
+type UseCustomDefaultFlowRunsFilter = {
+  value: Ref<SavedSearchFilter | null>,
+  set: (value: SavedSearchFilter) => void,
+  remove: () => void,
+  asFlowRunsFilter: ComputedRef<FlowRunsFilter | null>,
+}
```
pleek91, line 12: This composition is super simple and I like it. But I found the naming a little confusing. I think I got thrown off by the composition being "FlowRunsFilter" but `value` being a `SavedSearchFilter`. 

The only suggestion I have would be to rename this to `useCustomDefaultSavedSearchFilter`.

src/compositions/useCustomDefaultFlowRunsFilter.ts
```diff
@@ -1,8 +1,27 @@
...
+
+const { value, set } = useLocalStorage<SavedSearchFilter>(customDefaultFlowRunsFilterKey)
+
+export function useCustomDefaultFlowRunsFilter(): UseCustomDefaultFlowRunsFilter {
```
pleek91, line 16: I'm not 100% sure but it looks like where you're using this you check if this has a null value and then you coalesce the system default. Could this just automatically return the system default if the local storage value is null? It could also return an `isCustom` boolean so implementations would know.

src/components/SavedFiltersMenu.vue
```diff
@@ -54,9 +66,33 @@
     },
   })
 
-  const canSave = computed(() => internalSavedSearch.value?.name === customSavedSearch.name && can.create.saved_search)
+  const isCustomUnsavedFilter = computed(() => internalSavedSearch.value?.name === customSavedSearch.name)
+  const canSave = computed(() => isCustomUnsavedFilter.value && can.create.saved_search)
   const canDelete = computed(() => internalSavedSearch.value?.id && can.delete.saved_search)
 
+  const canToggleDefault = computed(() => {
+    // can't set the default to an unsaved filter. save it first
```
pleek91, line 74: Technically there's no reason why we couldn't let users set a custom unsaved filter as their default right? 

I wonder if we should just let users set anything as their default. I imaging that would cause issues with the toast when filter was deleted. But maybe we're overthinking that? I liked it when I saw you demo it but I almost thing letting users save any filters as their default and not just saved filters might be better even if we lose the toast.

src/compositions/useSavedFlowRunsSearches.ts
```diff
@@ -0,0 +1,61 @@
...
+  const defaultFlowRunsFilter = computed(() => mapper.map('SavedSearchFilter', myDefaultSavedFilter.value, 'FlowRunsFilter'))
+
+  const matchedCustomDefaultSavedSearch = computed(() => savedSearches.value.find(search => search.name !== customSavedSearch.name && isSameFilter(search.filters, myDefaultSavedFilter.value)))
+  const savedFlowRunsSearches = computed(() => savedSearches.value.map(search => ({ ...search, isDefault: matchedCustomDefaultSavedSearch.value?.name === search.name })))
+
+  // handle if customDefault is not in savedSearches i.e. someone else deleted it
+  watch([matchedCustomDefaultSavedSearch, savedSearchesSubscription], ([found, subscription]) => {
+    if (!found && subscription.executed) {
+      showToast('Your custom default flow runs filter was deleted by another user. Resave it or select a new default.')
+      removeDefault()
```
pleek91, line 37: Now that I see this toast implementation I think we should remove this. Even if the saved version of a users custom default is deleted. We can keep loading their custom default from local storage indefinitely. 

The only thing we'd need to account for is if we load a custom default (that doesn't match a saved search) and then the user changes the filters. We'll end up with two "Custom" filters. Which honestly is probably fine. But maybe we have a special name for custom default filters that don't match a saved search.

src/compositions/useSavedFlowRunsSearches.ts
```diff
@@ -0,0 +1,61 @@
...
+    }
+  })
+
+  async function deleteSavedFlowRunsSearch(savedSearchId: string): Promise<void> {
```
pleek91, line 41: This is interesting. We haven't wrapped api calls in composition methods like this before but I like how it automatically refreshes the subscription.

src/compositions/useSavedFlowRunsSearches.ts
```diff
@@ -0,0 +1,61 @@
...
+  async function createSavedFlowRunsSearch(search: SavedSearchCreate): Promise<SavedSearch> {
+    const savedSearch = await api.savedSearches.createSavedSearch(search)
+    savedSearchesSubscription.refresh()
+    return savedSearch
+  }
+  return {
+    savedFlowRunsSearches,
+    createSavedFlowRunsSearch,
+    deleteSavedFlowRunsSearch,
+    default: defaultFlowRunsFilter,
```
pleek91, line 58: I'd just return `defaultFlowRunsFilter` here since js/ts is going to make us alias it when destructuring since `default` is a keyword.

src/utilities/savedFilters.ts
```diff
@@ -71,7 +71,9 @@ export const excludeScheduledSavedSearch = new SavedSearch({
   filters: noScheduleFilter,
 })
 
-export const defaultSavesSearches = [
+export const systemDefaultSavedSearch = oneWeekSavedSearch
+
+export const builtinSavedSearches = [
```
pleek91, line 76: nit: I like the `system` keyword and might carry that into this array and call them `systemSavedSearches`.

pleek91 (commented):
This is going to be so nice for users! Shaping up nicely.

src/components/SavedFiltersMenu.vue
collincchoy: Ah that's an interesting thought 🤔 . 

especially because I'm saving the entire _filter_ itself and not just the name (which at some point I thought to myself how much easier that might've been).

pleek91: Yeah the name would have been easier! But since you've got the whole filter, we could just use it right? And if it happens to match up with a saved filter we'll show that as the selected and default filter.

src/compositions/useCustomDefaultFlowRunsFilter.ts
collincchoy: ah yeah. that makes sense. 

definitely was/still-getting-ahold-of the overarching data models here (savedsearch, savedFilters, flowrunsfilter, etc.) 
I think it _was_ the flow runs filter at one point hence the name and the awkward `asFlowRunsFilter` converter, but I realized i needed the whole SavedSearch. 

I think a source of confusion has been that the components are all for "SavedFilters" which seem to actually primarily deal w/ "SavedSearch" but it atm it seems coupled to flow runs specifically.

pleek91: I was realizing that myself when I was looking at this. Maybe a cleanup PR after this to unify the names of everything would be nice.

src/components/SavedFilters.vue
collincchoy: Reworked this to instead surface this as "Unsaved (default)" which is only visible locally to that user until they change their default _or_ save the unsaved (default) with a name.

src/compositions/useSavedFlowRunsSearches.ts
collincchoy: 👍 

image

pleek91: Nice, I like that. "Unsaved (default)" makes sense to me 👍

src/compositions/useSavedFlowRunsSearches.ts
```diff
@@ -0,0 +1,62 @@
...
+      } satisfies SavedFlowRunsSearch
+      // insert before the system default so that the readonly custom option is always at the top
+      const systemDefaultIndex = all.findIndex(savedSearch => savedSearch.name === systemDefaultSavedSearch.name)
+      all.splice(systemDefaultIndex, 0, unsavedDefault)
+    }
+    return all
+  })
+
+  async function deleteSavedFlowRunsSearch(savedSearchId: string): Promise<void> {
+    await api.savedSearches.deleteSavedSearch(savedSearchId)
```
collincchoy, line 48: If I delete a saved search that is my own local default, should that immediately remove it as my local default as well? 

atm I've settled to leave the user's version alone such that the removal is more of "oops I didn't want to share this w/ other folks but keep it just for me". I could see that being a bit odd to take extra steps to fully remove it, but it seems better than being optimistically destructive?

pleek91: I can see this both ways. A user might expect if they delete a filter that is currently their default to have their default reset. But I think since if someone else deletes the filter they get to keep their default it makes sense to keep it consistent.

src/components/SavedFilters.vue
```diff
@@ -11,36 +11,44 @@
...
 
+  onBeforeRouteUpdate((to, from, next) => {
+    if (isEmptyObject(to.query)) {
+      const defaultFlowRunsSavedSearchFilter = useDefaultSavedSearchFilter()
+      const asFlowRunsFilter = mapper.map('SavedSearchFilter', defaultFlowRunsSavedSearchFilter.value.value, 'FlowRunsFilter')
+      const asQueryParams = getQueryForFlowRunsFilter(asFlowRunsFilter)
+      return next({ ...to, query: asQueryParams })
+    }
+    return next()
+  })
```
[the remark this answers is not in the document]
collincchoy: okay _or_ challenge accepted - option 3 (#1775)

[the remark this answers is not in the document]
collincchoy: coolio resolved via #1775 which is now merged into this branch.

src/compositions/useDefaultSavedSearchFilter.ts
```diff
@@ -0,0 +1,44 @@
...
+  remove: () => void,
+  isCustom: ComputedRef<boolean>,
+}
+
+const { value: custom, set: setCustom } = useLocalStorage<SavedSearchFilter>(customDefaultFlowRunsFilterKey)
+
+/**
+ * Flow runs filter specifically
+ * @returns
+ */
```
pleek91, line 19: Does this need to be updated or does @returns do something magically here

collincchoy: oh oops. just a half-written doc string lmao don't mind me...

pleek91 (approved):
Nice job on a tricky problem!

PR #1764, opened 2023-09-21
Adds ability to toggle deployment parameter schema enforcement

desertaxle:
Adds the ability to toggle parameter schema enforcement via the UI. When toggled on, the parameter schema will be used to verify supplied deployment parameters when creating/updating a deployment or creating a flow run from a deployment. Operations will be rejected if the supplied parameters do not conform to the parameter schema of the deployment in question.

I'd also like to display if a deployment has parameter enforcement turned on in the deployment display view, but I'm sure the best place to put that information.

#### Screenshot
Screenshot 2023-09-21 at 9 45 58 AM

### DIFF
--- a/src/components/ParametersTable.vue
+++ b/src/components/ParametersTable.vue
@@ -1,7 +1,7 @@
 <template>
   <div class="deployment-parameters-table">
     <div class="deployment-parameters-table__search">
-      <ResultsCount :count="filtered.length" label="parameters" />
+      <ResultsCount :count="filtered.length" :label="toPluralString('parameter', filtered.length)" />
       <SearchInput v-model="searchTerm" placeholder="Search parameters" label="Search parameters" />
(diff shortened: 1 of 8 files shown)


src/components/ParametersTable.vue
```diff
@@ -1,7 +1,7 @@
 <template>
   <div class="deployment-parameters-table">
     <div class="deployment-parameters-table__search">
-      <ResultsCount :count="filtered.length" label="parameters" />
+      <ResultsCount :count="filtered.length" :label="toPluralString('parameter', filtered.length)" />
```
pleek91, line 4: ResultsCount already pluralizes the string. Pass the singular label and it will run it through toPluralString I believe

pleek91 (approved):
One note but otherwise looks good!

PR #1766, opened 2023-09-21
Update Global Concurrency Validation

zhen0:
Adds extra validation around the creation of global concurrency limits. Inspired by https://github.com/PrefectHQ/prefect/issues/10779 (but that issue will be fully closed by api validation).


pleek91:
@zhen0 and I discussed writing a validation rule `isGreaterThanOrEqual` to user here so we can have a reusable rule for this sort of thing.

PR #1768, opened 2023-09-21
Chore: Clean up deployment and flow tables

pleek91:
# Description
Everything works the same just much simpler and takes advantage of some newer components and compositions we have.


zhen0 (approved):

src/components/DeploymentsTable.vue
```diff
@@ -2,31 +2,23 @@
...
+        <ResultsCount v-if="selectedDeployments.length == 0" label="Deployment" :count="total" />
+        <SelectedCount v-else :count="selectedDeployments.length" />
 
-          <DeploymentsDeleteButton v-if="can.delete.deployment" :selected="selectedDeployments" @delete="deleteDeployments" />
-        </div>
+        <DeploymentsDeleteButton v-if="can.delete.deployment" small :selected="selectedDeployments" @delete="deleteDeployments" />
       </template>
 
       <template #header-end>
-        <div class="deployments-table__header-end">
```
collincchoy, line 14: It looks like at smaller screens this used to wrap this section rather than overflow?

| Before | After |
| ---- | ---- |
| image | image |

collincchoy: Ah it's also losing a tiny bit of margin between the header start/end (because the margin was on the end class)

image


vs

image

src/components/FlowsTable.vue
```diff
@@ -2,20 +2,16 @@
...
+        <ResultsCount v-if="selectedFlows.length == 0" label="Flow" :count="total" />
+        <SelectedCount v-else :count="selectedFlows.length" />
 
-          <FlowsDeleteButton v-if="can.delete.flow" :selected="selectedFlows" @delete="deleteFlows" />
-        </div>
+        <FlowsDeleteButton v-if="can.delete.flow" :selected="selectedFlows" small @delete="deleteFlows" />
       </template>
 
       <template #header-end>
-        <div class="flows-table__header-end">
```
collincchoy, line 14: same comments here as for deployments table

src/components/DeploymentsTable.vue
```diff
@@ -87,43 +79,44 @@
...
+  import { SearchInput, ResultsCount, DeploymentToggle, FlowRouterLink, DeploymentsDeleteButton, SelectedCount } from '@/components'
+  import DeploymentTagsInput from '@/components/DeploymentTagsInput.vue'
+  import { useWorkspaceRoutes, useCan, useDeploymentsFilterFromRoute, useComponent, useOffsetStickyRootMargin, useDeployments } from '@/compositions'
   import { Deployment, isRRuleSchedule, Schedule } from '@/models'
   import { DeploymentsFilter } from '@/models/Filters'
   import { deploymentSortOptions } from '@/types/SortOptionTypes'
 
   const props = defineProps<{
     filter?: DeploymentsFilter,
-    hideFlowFilter?: boolean,
```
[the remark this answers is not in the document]
pleek91: Correct 👍

src/components/DeploymentsTable.vue
```diff
@@ -159,30 +152,21 @@
...
-  const model = computed({
+
+  const selectAllValue = computed({
     get() {
       return selectedDeployments.value.length === deployments.value.length
     },
     set(value: boolean) {
-      selectAllDeployments(value)
+      if (value) {
+        selectedDeployments.value = [...deployments.value.map(deployment => deployment.id)]
```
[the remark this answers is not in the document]
pleek91: Ah good call. Ty

src/components/DeploymentsTable.vue
pleek91: Ah good catch. Meant to fix this in the table layout component. Will address

collincchoy: it feels _maybe_ like PLayoutTable's header should have a little flex gap on it?

pleek91: Agreed

pleek91: PR for that here https://github.com/PrefectHQ/prefect-design/pull/972

pleek91: Bumped prefect-design to latest to bring in that change 👍

collincchoy (approved):
dope

src/components/DeploymentsTable.vue
collincchoy: cool. i went and looked at some other p-layout-tables (just in nebula) and looks like only these tables had margin explicitly in the children so elsewhere, we should just be _gaining_ some nice space at denser heads

PR #1773, opened 2023-09-22
BugFix: Use lodash.merge to avoid mutating filter props

pleek91:
# Description
Several components accept a filter and then spread that filter into a `useSomethingFilter` composition as the base filter. But those compositions (specifically the route versions) directly modify the base filter. This seems to work find when navigating through the app but I noticed that if there were filters in the url and you refreshed the page the page would be broken and there would be an error in the console. Some digging revealed that error is from vue trying to mutate the filters prop, but props are readonly so the proxy trap returned false. 

I was able to fix this by using `merge` to merge props.filter and any additional filters the component applies onto an empty object. This makes sure the values from props.filter are applied without attempting to write to props.filter. And from my testing all the filters work as expected even after a page refresh. 

Before
image

After
image

### DIFF
--- a/src/components/DeploymentsTable.vue
+++ b/src/components/DeploymentsTable.vue
@@ -81,6 +81,7 @@
 <script lang="ts" setup>
   import { PTable, PTagWrapper, PEmptyResults, PLink, TableColumn } from '@prefecthq/prefect-design'
   import { NumberRouteParam, useDebouncedRef, useRouteQueryParam } from '@prefecthq/vue-compositions'
+  import merge from 'lodash.merge'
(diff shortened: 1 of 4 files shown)


brandonreid (approved):
Nice, this [showed up in logrocket](https://app.logrocket.com/qetfiw/cloud-ui-crv2i/issues/needs-triage?from=issues%2Fneeds-triage&showMostSevereIssues=true) as well btw, good deal of folks encountering it.

image

pleek91:
@brandonreid we should probably leverage log rockets error collecting more and get a head of these.

PR #1777, opened 2023-09-25
Chore: `useFlowRuns` cleanup

pleek91:
# Description
This PR cleans up some various uses of fetching flow runs. It also implements https://github.com/PrefectHQ/prefect/pull/10806 rather than fetching task runs first. Now `useFlowRuns` is used consistently anywhere we are fetching flow runs.


src/components/FlowRunsAccordionContent.vue
```diff
@@ -24,20 +23,22 @@
 
   const limit = computed(() => props.flowRunLimit ?? 3)
 
-  const filter = computed<FlowRunsFilter>(() => ({
+  const filter = (): FlowRunsFilter => ({
```
collincchoy, line 26: Is the refactors from computed getters to just getters for simplicity and/or perf? otherwise why?

collincchoy (approved):
just a non-blocking Q. 

lgtm

src/components/FlowRunsAccordionContent.vue
pleek91: Correct less code and more performant. Originally our compositions accepted `MaybeRef` but since vue introduced the `MaybeRefOrGetter` and `toValue` I've been trying to update all the compositions to accept getters and then remove unnecessary computed properties like this.

PR #1788, opened 2023-09-29
BugFix: Make sure infinite mode is used for the infinite scroller

pleek91:
# Description
Fixes subflows getting into an infinite loop trying to load


pleek91:
Skipping code review to get this out

PR #1794, opened 2023-10-03
Bugfix: Ability to run test suite

### LINKED ISSUE #1772: Unit tests don't work when something from `vue` is imported
# Description
When writing a unit test if you import something from vue or import something that has a dependency on vue you'll get an error when running the test. Sometimes its different but you'll get something like this.

image

We successfully run unit tests in vue-compositions so there's something unique to this project that is causing this to fail.

znicholasbrown:
This PR addresses #1772 by:

- bumping the node version the package should use from 16.5.0 > 18.18.0
- adding an explicit vitest configuration ~file~ section that allows us to specify globals and environments specifically for tests
- fixing a circular import in the array utilities (one of the utilities was relying on the mocker... which also relies on the utilities 😅)

Resolves #1772


src/utilities/arrays.ts
```diff
@@ -122,11 +121,11 @@ export function repeat<T>(count: number, method: (index: number) => T): T[] {
 export function some<T>(source: T[], min?: number, max?: number): T[] {
   const minArg = min ?? 1
   const maxArg = max ?? source.length
-  const count = mocker.create('number', [minArg, maxArg])
+  const count = uniform(minArg, maxArg)
```
pleek91, line 124: ahhhhhhhhhhhhh

vitest.config.ts
pleek91, line 1: I'm surprised we need this. Most of this is duplicate from the vite.config.ts isn't it? Couldn't we add the `test` section into the root config?

znicholasbrown: Yeah we definitely could, I broke this out while I was still figuring out what exactly was misconfigured - if that's your preference I'll just add a new section to the other config

pleek91: I think adding the test section to the other config is preferred so that we don't have the manager all the other settings in two places

znicholasbrown: yeah fair enough - updated!

pleek91 (approved):

PR #1796, opened 2023-10-05
Add an object-level `can` to deployment objects

collincchoy:
This PR adds a compatibility layer which adds a property to the deployment object for verifying access to a _specific_ deployment object as opposed to the global can, resource-type level scopes.

> **Note**
> Not all relevant permission checks are being swapped over to use this new property yet in this PR. Follow-up to come with the less-trivial checks that need to be changed.


src/components/FlowRunMenu.vue
```diff
@@ -1,6 +1,7 @@
 <template>
   <p-icon-button-menu>
     <template #default>
+      <!-- todo: this PR(https://github.com/PrefectHQ/prefect-ui-library/pull/1743) will only partially work once object acls are on -->
```
collincchoy, line 4: oops looks like this comment bled into my PR before I got around to addressing it. 

for context, I've marked several places like here where `can.*.deployment` checks are happening but no deployment object is available in-context. Some places are probably _fine_ - for example, `can.view.deployment` checks since we decided that `view` is a pre-requisite to gaining additional permissions via an ACL. Other places need some thought like this one. 

I'm thinking of having a follow-up PR with those places slash going through them all separately after this PR passes the "yeah this approach all checks out"

src/models/Deployment.ts
```diff
@@ -88,6 +91,7 @@ export class Deployment implements IDeployment {
     this.workPoolName = deployment.workPoolName
     this.enforceParameterSchema = deployment.enforceParameterSchema
     this.pullSteps = deployment.pullSteps
+    this.can = deployment.can ?? createDeploymentCan()
```
pleek91, line 94: Wonder if there's some way we can safeguard agains this being used accidentally in cloud. Perhaps we could add an `if cloud` type check and throw an error if the dummy checks are used?

pleek91:
I wonder if we can improve the syntax a bit. I still feel like `deployment.can` sounds like we're checking the deployments permissions rather than the actors permissions for that deployment. Possibilities

```
// simpliest
deployment.actor.can.run

// kinda annoying 
canActor(deployment, 'run')

// kind like this one, but not sure how it would work with the existing can object which has can.run.deployment as a generic scope check. 
can.run.deployment(deployement)

// or this. But I find this confusing because `can.run.deployment` also is related to the current actor. 
can.actor.run.deployment(deployment)
```

src/models/Deployment.ts
collincchoy: is there some cloud boolean somewhere already? 

I feel _slightly_ not concerned about this only because in cloud, you shouldn't have to create a deployment can anywhere - (🤞 atleast right now...) it should just be on deployment instantiation. 

i feel like the root of the smell is really that we're not least privilege by default - ~~we're instead open access by default so where it _feels_ like ideally we should always fallback to no priv,~~
hmm i guess to your point, maybe the constructor _should_ require the can? whereas I let it be optional for easier compatibility...

pleek91: I realized when reviewing the cloud pr for this that this wouldn't be possible anyway. We don't instantiate these classes anywhere outside of the service methods and in cloud the service method automatically sets this. So I'm not worried about this anymore.

pleek91: I don't feel strongly about requiring the can be passed in but it would provide a safe guard if someone does `new Deployment()` somewhere. Might be best?

collincchoy:
> I wonder if we can improve the syntax a bit. I still feel like `deployment.can` sounds like we're checking the deployments permissions rather than the actors permissions for that deployment. Possibilities
> 
> ```
> // simplest
> deployment.actor.can.run
> 
> // kinda annoying 
> canActor(deployment, 'run')
> 
> // kind like this one, but not sure how it would work with the existing can object which has can.run.deployment as a generic scope check. 
> can.run.deployment(deployment)
> 
> // or this. But I find this confusing because `can.run.deployment` also is related to the current actor. 
> can.actor.run.deployment(deployment)
> ```

yeah. agreed it's not quite... idiomatically ideal. on the other hand, it's pretty close/familiar to the other `can` which is nice. I like the latter 2 the most but atm there's complexity in having a parameterized, at-run-time type checks like that.

pleek91 (approved):

PR #1797, opened 2023-10-05
Switch over to use `obj.can.x` in place of `can.x.objType`

collincchoy:
~~I need to look into these checks.~~ #1796 added a new `can` object onto deployments which factor in deployment ACLs in addition to the current actor's workspace scopes and so ideally all `can.x.deployments` shift over to using the new deployment can. 

In #1796, I swapped over all of the easy swaps where either:
* a specific deployment is already in scope by which I could switch `can.update.deployments -> deployment.can.update` or
* it's just a check to `can.view.deployments` since the ACLs require atleast view permission at the workspace level in order to take additive effect

These are the leftovers from my <kbd>CMD</kbd> + <kbd>SHIFT</kbd> + <kbd>F</kbd> which don't have an obvious deployment object already in scope ~~that I need to figure out the most appropriate alt. course of action for.~~


src/components/DeploymentsTable.vue
```diff
@@ -30,6 +31,7 @@
           <p-link :to="routes.deployment(row.id)">
             <span>{{ row.name }}</span>
           </p-link>
+          <span>{{ JSON.stringify(row.can) }}</span>
```
collincchoy, line 34: ok sorry not this line. this is just me testing. 😭

src/components/FlowRunMenu.vue
```diff
@@ -1,7 +1,7 @@
 <template>
   <p-icon-button-menu>
     <template #default>
-      <!-- todo: this PR(https://github.com/PrefectHQ/prefect-ui-library/pull/1743) will only partially work once object acls are on -->
+      <!-- todo: audit [cchoy] this PR(https://github.com/PrefectHQ/prefect-ui-library/pull/1743) will only partially work once object acls are on -->
```
collincchoy, line 4: For context, looking at what this PR solves is a good example of what I'm trying to fix now. So this past PR added this `can.run.deployment` check so that the button won't render if the current user cannot actually run deployments.

However, now that object ACLs are around and they can grant _additional_ permissions to an actor beyond what they may otherwise have at the workspace level, this check could now evaluate to false when the actor actually _can_ run the specific deployment that's in context because they're explicitly given run access via the object acl but don't have `can.run.deployment`

src/compositions/useBlockDocument.ts
```diff
@@ -13,6 +13,7 @@ export function useBlockDocument(blockDocumentId: MaybeRefOrGetter<string | null
   const can = useCan()
 
   const getter: Getter<[string] | null> = () => {
+    // todo: audit [cchoy]
```
collincchoy, line 16: I'm not sure what these checks are necessary for 🤔 shouldn't we just let the backend handle this check.

hmmmm either way all of these can.read.* checks in compositions should be _okay_ since you must have can.read in order to get any escalated priveleges to the objects via acl anyways.

src/components/ScheduleFieldset.vue
```diff
@@ -4,6 +4,7 @@
       {{ internalValue.toString({ verbose: true }) }}
     </div>
 
+    <!-- todo: audit [cchoy] -->
```
collincchoy, line 7: not quite sure what to do with this one 🤔

src/components/DeploymentsTable.vue
```diff
@@ -5,6 +5,7 @@
         <ResultsCount v-if="selectedDeployments.length == 0" label="Deployment" :count="total" />
         <SelectedCount v-else :count="selectedDeployments.length" />
 
+        <!-- todo: audit [cchoy] -->
```
collincchoy, line 8: this is kinda tricky along with the other check further down. 

perhaps in practice this is actually _okay_ - if I don't have workspace permissions to all deployments, but I have manage on a specific one via an ACL - maybe it's _okay_ that we don't let those folks delete that one specific one. just... might be non-intuitive. 

otherwise yeah this would take some thinking. Like we'd maybe have to show the checkbox column `if some(deployments.map(d => d.can.delete)` and then in each row, disable the checkbox if `!deployment.can.delete`.

collincchoy: nvm while writing down this comment I had a thought / figure-this-out. 

addressed in 04a9adeb28be4ff87e0bbbb4f8a6727161ec9357

collincchoy:
Note to reviewer:

There are some todo: audit comments intentionally left in here. Idea is to figure out what to do in those places in this PR and then open to remove comments once resolved.

src/compositions/useBlockDocument.ts
collincchoy: mm to that extent, I'm going to remove these

src/components/FlowRunMenu.vue
pleek91: For this I'd would just get the deployment and check it. We can get individual deployments and let the subscriptions and the request batcher take care of caching and optimizations.

src/components/ScheduleFieldset.vue
pleek91: I'm not sure where this is used (on my phone or I'd look) but I feel like this component should just accept a deployment rather than a schedule. I'll have to look at why it's accepting just a schedule

src/compositions/useBlockDocument.ts
pleek91: These were here to just avoid making a call we knew was going to fail. 

Since that's no longer the case it might make sense to remove these permission checks in compositions that fetch objects with acls.

pleek91 (commented):
Overall this looks good to me. I think the schedule component is the only one that might need some other refactoring to make it work.

src/components/FlowRunMenu.vue
```diff
@@ -1,8 +1,7 @@
 <template>
   <p-icon-button-menu>
     <template #default>
-      <!-- todo: audit [cchoy] this PR(https://github.com/PrefectHQ/prefect-ui-library/pull/1743) will only partially work once object acls are on -->
-      <p-overflow-menu-item v-if="flowRun?.deploymentId && can.run.deployment" label="Copy to new run" :to="routes.deploymentFlowRunCreate(flowRun.deploymentId, flowRun.parameters)" />
+      <p-overflow-menu-item v-if="flowRun?.deploymentId && deployment?.can.run" label="Copy to new run" :to="routes.deploymentFlowRunCreate(flowRun.deploymentId, flowRun.parameters)" />
```
pleek91, line 4: nit: I don't think we need the `flowRun?.deploymentId` check here. `deployment?.can.run` will be undefined if there is no deploymentId right?

src/components/ScheduleFieldset.vue
```diff
@@ -33,12 +32,12 @@
 <script lang="ts" setup>
   import { computed } from 'vue'
   import ScheduleFormModal from '@/components/ScheduleFormModal.vue'
-  import { useCan } from '@/compositions/useCan'
   import { Schedule } from '@/models'
 
   const props = defineProps<{
     modelValue: Schedule | null,
     loading?: boolean,
+    readonly?: boolean,
```
pleek91, line 40: This seems a little risky since it defaults to `false`. When really we need the implementation to tell us this. This component could easily be implemented without this being set and show a user an action they don't have permission to complete. 

The simplest way might be to just make `readonly` a required prop.

collincchoy: 👍 - I don't see any outside consumption of this component in oss or cloud so making it required should be non-breaking update for consumers.

src/components/FlowRunMenu.vue
collincchoy: uhhh. yeah I think so.

collincchoy: oh actually the explicit check works as a type guard for the other half of this line where it expects a non-null `flowRun.deploymentId`

pleek91 (approved):

PR #1800, opened 2023-10-09
swap `estimatedRunTime` to be preferred for `duration`

jakekaplan:
closes: https://github.com/PrefectHQ/nebula/issues/5778

Swaps `estimated_run_time` to take preference over `total_run_time`.

`total_run_time`: Is the total amount of time a flow/task run has spent in `RUNNING`. This is updated after a run has _left_ the `RUNNING` state.

`estimated_run_time`: If a state is not in `RUNNING` this is just the `total_run_time`. If a state is in `RUNNING`, this is the the `total_run_time` + the delta between when the run entered `RUNNING` and now.

This will correctly update the duration field in the UI when a run is still in progress after a retry. Previously because `total_run_time` was preferred, an in process run on it's second attempt would have it's duration updated while it was still in progress.

Examples: Note the `total_run_time` and `estimated_run_time` in the api responses and the `duration` on the page.
Screenshot 2023-10-09 at 12 35 14 PM
Screenshot 2023-10-09 at 12 35 14 PM

### DIFF
--- a/src/models/TaskRun.ts
+++ b/src/models/TaskRun.ts
@@ -84,6 +84,6 @@ export class TaskRun implements ITaskRun {
   }
 
   public get duration(): number {
-    return (this.estimatedRunTime && this.estimatedRunTime > 0 ? this.estimatedRunTime : this.totalRunTime) ?? 0;
+    return (this.estimatedRunTime && this.estimatedRunTime > 0 ? this.estimatedRunTime : this.totalRunTime) ?? 0
   }
 }
--- a/src/models/TaskRun.ts
+++ b/src/models/TaskRun.ts
@@ -84,6 +84,6 @@ export class TaskRun implements ITaskRun {
   }
 
   public get duration(): number {
-    return (this.totalRunTime && this.totalRunTime > 0 ? this.totalRunTime : this.estimatedRunTime) ?? 0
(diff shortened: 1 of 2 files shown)


pleek91:
@jakekaplan do we even need to use total run time at all? It sounds like the estimate will always be set

src/models/TaskRun.ts
```diff
@@ -84,6 +84,6 @@ export class TaskRun implements ITaskRun {
   }
 
   public get duration(): number {
-    return (this.totalRunTime && this.totalRunTime > 0 ? this.totalRunTime : this.estimatedRunTime) ?? 0
+    return (this.estimatedRunTime && this.estimatedRunTime > 0 ? this.estimatedRunTime : this.totalRunTime) ?? 0;
```
znicholasbrown, line 87:
```suggestion
    return (this.estimatedRunTime && this.estimatedRunTime > 0 ? this.estimatedRunTime : this.totalRunTime) ?? 0
```

znicholasbrown (approved):

jakekaplan:
@pleek91 that is a good call! I think technically if it were an older server version it might not have it? I'm not sure what our stance is on server/ui compatibility of different versions?

pleek91:
> @pleek91 that is a good call! I think technically if it were an older server version it might not have it? I'm not sure what our stance is on server/ui compatibility of different versions?

@jakekaplan good point!

pleek91 (approved):

PR #1808, opened 2023-10-10
Enhancement: Show flow name in deployment combobox when an option is selected

znicholasbrown:
Before:
Screenshot 2023-10-10 at 4 31 55 PM

After:
Screenshot 2023-10-10 at 4 32 30 PM

### DIFF
--- a/src/components/DeploymentCombobox.vue
+++ b/src/components/DeploymentCombobox.vue
@@ -7,12 +7,7 @@
       <slot v-bind="scope">
         <UseDeploymentSlot v-if="isString(scope.value)" :deployment-id="scope.value">
           <template #default="{ deployment }">
-            <template v-if="deployment.flowId">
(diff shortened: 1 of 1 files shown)


src/components/DeploymentCombobox.vue
```diff
@@ -7,15 +7,20 @@
       <slot v-bind="scope">
         <UseDeploymentSlot v-if="isString(scope.value)" :deployment-id="scope.value">
           <template #default="{ deployment }">
-            {{ deployment.name }}
+            <template v-if="deployment.flowId">
```
pleek91, line 10: Is flowId not a required field? Surprised we need a check here or the name fallback.

pleek91 (approved):

src/components/DeploymentCombobox.vue
znicholasbrown: updated 👍🏻

PR #1809, opened 2023-10-10
Enhancement: Deployment combobox option alignment

znicholasbrown:
This PR:
- fixes an issue with the `useFlowRunsMock` composition where the deployments and flows created for downstream flow runs had no relation and so would break demos
- fixes alignment and visibility issues with deployment combobox options

I've simplified the `DeploymentComboboxOption` quite bit because it was overly complicated attempting to account for visibility and wasn't using any of our new compositions. Should be much more straight forward now

Before:
Screenshot 2023-10-10 at 7 15 53 PM

After:
Screenshot 2023-10-10 at 7 16 44 PM

### DIFF
--- a/src/components/DeploymentComboboxOption.vue
+++ b/src/components/DeploymentComboboxOption.vue
@@ -20,7 +20,7 @@
   }>()
 
   const element = ref<HTMLDivElement>()
-  const { flow } = useFlow(props.flowId)
+  const { flow } = useFlow(() => props.flowId)
 </script>
 
 <style>
--- a/src/components/DeploymentComboboxOption.vue
+++ b/src/components/DeploymentComboboxOption.vue
@@ -1,5 +1,5 @@
 <template>
-  <div ref="element" class="deployment-combobox-option">
+  <div class="deployment-combobox-option">
     <template v-if="flow">
       <span class="deployment-combobox-option__flow-name">
         {{ flow.name }}
@@ -11,7 +11,6 @@
 </template>
 
 <script lang="ts" setup>
-  import { ref } from 'vue'
   import { useFlow } from '@/compositions'
 
   const props = defineProps<{
@@ -19,7 +18,6 @@
     deploymentName?: string,
   }>()
 
-  const element = ref<HTMLDivElement>()
   const { flow } = useFlow(() => props.flowId)
 </script>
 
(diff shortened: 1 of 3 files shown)


znicholasbrown:
Note that this will mean we're calling the underlying subscription whenever this component is mounted, regardless of visibility. I think this is ok because I don't think a component at this level should have to worry about its visibility, that feels like the wrong level of abstraction. In addition, this component is only used as a child of `DeploymentCombobox` which uses virtual scrolling/pagination to control component mounting/unmounting which will mean only a small number of these are ever really mounted at once.

src/components/DeploymentComboboxOption.vue
```diff
@@ -1,48 +1,37 @@
...
-  const flowArgs = computed<Parameters<typeof api.flows.getFlow> | null>(() => {
-    if (!visible.value) {
-      return null
-    }
-    return [flowId.value]
-  })
-
-  const flowSubscription = useSubscriptionWithDependencies(api.flows.getFlow, flowArgs)
-  const flowResponse = computed(() => flowSubscription.response)
+  const { flow } = useFlow(props.flowId)
```
pleek91, line 23: Will need to use a getter here so it's reactive
```suggestion
  const { flow } = useFlow(() => props.flowId)
```

src/components/DeploymentComboboxOption.vue
```diff
@@ -1,48 +1,37 @@
...
 
   const props = defineProps<{
     flowId: string,
     deploymentName?: string,
   }>()
 
-  const api = useWorkspaceApi()
   const element = ref<HTMLDivElement>()
-  const { visible } = useVisibilityObserver(element, { disconnectWhenVisible: true })
-
```
pleek91, line 24: I don't think we need this element ref now.

src/components/DeploymentComboboxOption.vue
pleek91, line 1: So is the idea that we don't really need the visibility observer anymore because the flow requests are batched?

pleek91 (changes requested):
Just one request but otherwise looks good. Haven't seen mock code in a while!

src/components/DeploymentComboboxOption.vue
znicholasbrown: ah nice catch

src/components/DeploymentComboboxOption.vue
znicholasbrown: true! removed

src/components/DeploymentComboboxOption.vue
znicholasbrown: Between batching and the fact that this is only used in a virtual scroller I think the overhead of a bunch of visibility observers didn't make much sense

pleek91: Yeah I agree. We have better tools now 👍

pleek91 (approved):

PR #1810, opened 2023-10-11
Fix: object level can returns falsies in templates

collincchoy:
The object level can Proxy object which is supposed to by design return **true** for all property accesses to simulate a `can.run => true`, `can.read => true`, etc. object.

This PR fixes a bug where the object level can gets unwrapped _in vue templates_ resulting in `can.x` property accesses resolving to undefined which is **falsey**.

The fix here is to change the proxy's implementation to _only proxy known properties_ rather than all properties so that whatever key/symbol that Vue uses to identify the proxy as a ref needing unwrapping is left untouched.

### DIFF
--- a/src/models/ObjectLevelCan.ts
+++ b/src/models/ObjectLevelCan.ts
@@ -3,7 +3,7 @@ import { PermissionVerb, WorkspacePermission, workspacePermissions } from '@/ser
 type PermissionVerbs = PermissionVerb<WorkspacePermission>
 const permissionVerbs = workspacePermissions.map(permission => permission.split(':')[0]) as readonly PermissionVerbs[]
 
-type ObjectTypesWithPermissions = WorkspacePermission extends `${string}:${infer TObject}` ? TObject : never
+export type ObjectTypesWithPermissions = WorkspacePermission extends `${string}:${infer TObject}` ? TObject : never
 
 type ActionsForObjectsHelper<O extends string, T extends string> = T extends `${infer Action}:${infer TObject}` ? TObject extends O ? Action : never : never
 type PermissionsForObjectType<TObjectType extends ObjectTypesWithPermissions> = ActionsForObjectsHelper<TObjectType, WorkspacePermission>
--- a/src/models/ObjectLevelCan.ts
+++ b/src/models/ObjectLevelCan.ts
@@ -1,19 +1,23 @@
-type ObjectLevelCan<TPermissions extends string> = {
-  [key in TPermissions]: boolean
+import { PermissionVerb, WorkspacePermission, workspacePermissions } from '@/services/can'
+
+type PermissionVerbs = PermissionVerb<WorkspacePermission>
+const permissionVerbs = workspacePermissions.map(permission => permission.split(':')[0]) as readonly PermissionVerbs[]
+
+type ObjectTypesWithPermissions = WorkspacePermission extends `${string}:${infer TObject}` ? TObject : never
+
+type ActionsForObjectsHelper<O extends string, T extends string> = T extends `${infer Action}:${infer TObject}` ? TObject extends O ? Action : never : never
+type PermissionsForObjectType<TObjectType extends ObjectTypesWithPermissions> = ActionsForObjectsHelper<TObjectType, WorkspacePermission>
+
+export type ObjectLevelCan<TObjectType extends ObjectTypesWithPermissions> = {
+  [key in PermissionsForObjectType<TObjectType>]: boolean
 }
 
-export const BasicObjectLevelPermissions = ['create', 'read', 'update', 'delete'] as const
-export const RunnableObjectLevelPermissions = [...BasicObjectLevelPermissions, 'run'] as const
-
-export type BasicPermissionsObjectLevelCan = ObjectLevelCan<typeof BasicObjectLevelPermissions[number]>
-export type BasicRunnablePermissionsObjectLevelCan = ObjectLevelCan<typeof RunnableObjectLevelPermissions[number]>
-
-export function createObjectLevelCan<T extends Record<string, boolean>>(knownProperties: typeof BasicObjectLevelPermissions | typeof RunnableObjectLevelPermissions): T {
-  const set = new Set(knownProperties)
-  return new Proxy({} as T, {
-    get(___, property) {
+export function createObjectLevelCan<T extends ObjectTypesWithPermissions>(): ObjectLevelCan<T> {
+  const knownProperties = permissionVerbs
+  return new Proxy({} as ObjectLevelCan<T>, {
+    get(_target, property) {
       // only proxy known properties so that vue doesn't think it's a ref in templates
-      if (set.has(property as unknown as typeof RunnableObjectLevelPermissions[number])) {
+      if (knownProperties.includes(property as PermissionVerbs)) {
         return true
       }
     },
--- a/src/models/ObjectLevelCan.ts
+++ b/src/models/ObjectLevelCan.ts
@@ -15,6 +15,7 @@ export function createObjectLevelCan<T extends Record<string, boolean>>(knownPro
   const set = new Set(knownProperties)
   return new Proxy({} as T, {
     get(___, property) {
+      // only proxy known properties so that vue doesn't think it's a ref in templates
       if (set.has(property as unknown as typeof RunnablePermissions[number])) {
         return true
       }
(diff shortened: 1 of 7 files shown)


src/models/ObjectLevelCan.ts
```diff
@@ -5,13 +5,20 @@ type ObjectLevelCan<TPermissions extends string> = {
...
-export function createObjectLevelCan<T extends Record<string, boolean>>(): T {
+export type BasicPermissionsObjectLevelCan = ObjectLevelCan<typeof BasicPermissions[number]>
+export type BasicRunnablePermissionsObjectLevelCan = ObjectLevelCan<typeof RunnablePermissions[number]>
+
+export function createObjectLevelCan<T extends Record<string, boolean>>(knownProperties: typeof BasicPermissions | typeof RunnablePermissions): T {
+  const set = new Set(knownProperties)
   return new Proxy({} as T, {
-    get() {
-      return true
+    get(___, property) {
```
pleek91, line 14: super nit: just because we haven't used the underscore skip pattern anywhere else

```suggestion
    get(target, property) {
```

src/models/ObjectLevelCan.ts
```diff
@@ -1,17 +1,21 @@
...
-export type BasicRunnablePermissionsObjectLevelCan = ObjectLevelCan<BasicPermissions | 'run'>
+export const BasicObjectLevelPermissions = ['create', 'read', 'update', 'delete'] as const
+export const RunnableObjectLevelPermissions = [...BasicObjectLevelPermissions, 'run'] as const
+
+export type BasicPermissionsObjectLevelCan = ObjectLevelCan<typeof BasicObjectLevelPermissions[number]>
+export type BasicRunnablePermissionsObjectLevelCan = ObjectLevelCan<typeof RunnableObjectLevelPermissions[number]>
 
-export function createObjectLevelCan<T extends Record<string, boolean>>(): T {
+export function createObjectLevelCan<T extends Record<string, boolean>>(knownProperties: typeof BasicObjectLevelPermissions | typeof RunnableObjectLevelPermissions): T {
+  const set = new Set(knownProperties)
```
pleek91, line 12: I don't have a perf to point to ATM but I've had issues in the past where creating lots of sets was actually a lot slower than using an array. Since this happens for every single deployment/block and considering how small these array are IMO we should just check the array directly using `Array.includes` which I actually think is faster on super small arrays like this.

src/models/ObjectLevelCan.ts
```diff
@@ -1,17 +1,21 @@
...
-export type BasicPermissionsObjectLevelCan = ObjectLevelCan<BasicPermissions>
-export type BasicRunnablePermissionsObjectLevelCan = ObjectLevelCan<BasicPermissions | 'run'>
+export const BasicObjectLevelPermissions = ['create', 'read', 'update', 'delete'] as const
+export const RunnableObjectLevelPermissions = [...BasicObjectLevelPermissions, 'run'] as const
+
+export type BasicPermissionsObjectLevelCan = ObjectLevelCan<typeof BasicObjectLevelPermissions[number]>
+export type BasicRunnablePermissionsObjectLevelCan = ObjectLevelCan<typeof RunnableObjectLevelPermissions[number]>
 
-export function createObjectLevelCan<T extends Record<string, boolean>>(): T {
+export function createObjectLevelCan<T extends Record<string, boolean>>(knownProperties: typeof BasicObjectLevelPermissions | typeof RunnableObjectLevelPermissions): T {
```
pleek91, line 11: Thought:

Do we really care if only the exact permissions that object has return true as long as the types are correct? I'd vote not having this accept any arguments and just checking the array of known values directly. Especially since you're defining the known values here and exporting them just so the implementations can pass them back. Probably not worth the added complexity.

collincchoy, line 11: this could be simpler if I just default to the superset of expected permissions (which would atm be `RunnableObjectLevelPermissions`) but this matches `createCan` more closely and leaning toward explicit over implicit here.

src/models/ObjectLevelCan.ts
```diff
@@ -1,17 +1,21 @@
...
-
-
 type ObjectLevelCan<TPermissions extends string> = {
   [key in TPermissions]: boolean
 }
 
-export type BasicPermissionsObjectLevelCan = ObjectLevelCan<BasicPermissions>
-export type BasicRunnablePermissionsObjectLevelCan = ObjectLevelCan<BasicPermissions | 'run'>
+export const BasicObjectLevelPermissions = ['create', 'read', 'update', 'delete'] as const
+export const RunnableObjectLevelPermissions = [...BasicObjectLevelPermissions, 'run'] as const
```
pleek91, line 6: I want to tie these to the actual permissions so we get type errors and such when permissions are being changed. Tried a few things but the best idea I came up with was to update `can.ts` like this.

```typescript
export const permissionVerbs = ['create', 'delete', 'read', 'run', 'update'] as const
export type PermissionVerbs = typeof permissionVerbs[number]

export const workspacePermissions = [
  'create:artifact',
  'create:automation',
  'create:block',
  'create:concurrency_limit',
  'create:deployment',
  'create:flow_run',
  'create:flow',
  'create:log',
  ...etc
] as const satisfies Readonly<`${PermissionVerbs}:${string}`[]>
```

Then use `permissionVerbs` and `PermissionVerbs` in `createObjectLevelCan` as needed.

src/models/ObjectLevelCan.ts
collincchoy: Yeah i def. thought this - https://github.com/PrefectHQ/prefect-ui-library/pull/1810/files#r1355341844

figured i'd just lean toward explicit over implicit 🤷

pleek91: I also commented on this. I don't feel strongly, but IMO the extra complexity to distinguish between runnable and non runnable objects isn't worth it here. The types are enough protection for me. But I don't feel strongly.

src/models/ObjectLevelCan.ts
collincchoy: oh i forgot that each object invokes this on creation as opposed to a single can for deployments.

src/models/ObjectLevelCan.ts
collincchoy: I feel like... I do very much prefer not seeing this at a glance 
image


vs 

image

collincchoy: like vs code seems to know what i'm telling it so it feels like a common enough convention to me

pleek91: weird, never noticed vscode did magic with underscores for skips. The greyed out argument doesn't bother me. In fact I kinda prefer that because I can see its not used. And eslint is smart enough to not flag that its not used because it was needed to access the second property

pleek91: idk, we could start using underscores but we should updated our linting rules to allow a single underscore. I've just never liked this pattern because it ends up with thinks like
```typescript
fn(_, __, ___, thingIWant)
```

collincchoy: oh the grey'ed out to me makes me immediately want to go in and delete it. especially with the popover which isn't literally a warning but I associate it w/ linting warns

pleek91: Well its a super nit so not going to block this PR lol. If you wanna start introducing that pattern lets update eslint to allow single and double underscores 👍

collincchoy: image

vs code says the "quick fix" is to _prefix it with an underscore_

pleek91: Lol 🤷

src/models/ObjectLevelCan.ts
```diff
@@ -1,17 +1,25 @@
...
-    get() {
-      return true
+export function createObjectLevelCan<T extends ObjectTypesWithPermissions>(): ObjectLevelCan<T> {
+  const knownProperties = permissionVerbs
+  return new Proxy({} as ObjectLevelCan<T>, {
+    get(_target, property) {
+      // only proxy known properties so that vue doesn't think it's a ref in templates
+      if (knownProperties.includes(property as PermissionVerbs)) {
+        return true
+      }
```
collincchoy, line 22: This is the primary fix - only proxying known properties which is being computed from workspace permissions such that it should constrain the property access proxying to only affect properties like `.create`, `.read`, `.run`, `.update`, and `.delete` along with other verbs if they show up as valid workspace permissions.

The rest of changes are TS-related stuff to just make it a little easier to use this thing typing-wise. Instead of having types for each can type depending on what permission verbs are relevant to that object, TS magic lets you now use `ObjectLevelCan<the name of some object type constrained by real workspace permission values>` and it should _just_ work for knowing that the can should have `.run` for example.

znicholasbrown (approved):

PR #1827, opened 2023-10-26
Fix: Reduce sensitivity of now to remove duplicate queries

zhen0:
Along with https://github.com/PrefectHQ/vue-compositions/pull/332 this should resolve https://github.com/PrefectHQ/nebula-ui/issues/3577

### DIFF
--- a/src/maps/dashboard.ts
+++ b/src/maps/dashboard.ts
@@ -3,8 +3,16 @@ import { FlowRunsFilter, TaskRunsFilter, TaskRunsHistoryFilter, WorkPoolWorkersF
 import { MapFunction } from '@/services/Mapper'
 import { WorkspaceDashboardFilter } from '@/types/dashboard'
 
-export const mapWorkspaceDashboardFilterToTaskRunsFilter: MapFunction<WorkspaceDashboardFilter, TaskRunsFilter> = function(source) {
+function nowWithoutMilliseconds(): Date {
   const now = new Date()
+
+  now.setMilliseconds(0)
+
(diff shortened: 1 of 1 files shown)


pleek91:
This just makes sure we get cache hits on subscriptions. millisecond precision meant if there were multiple subscriptions to the "same" filter when the subscription watcher evaluated the change parameters each subscription would get a slightly different filter (difference of a millisecond usually). So truncating the milliseconds ensures we get cache hits and don't make duplicate requests.

collincchoy (commented):
~~Hmm can we add something explaining what/how this fixes something? I like PRs as anchors so just a little more context here or in the description will be helpful for future (and present!) collin. or code comments too.~~

~~I see _what_ this is doing but I don't get _why_/how/what-it-fixes~~

🤦 nvm i see the comment now.

collincchoy (approved):
This. makes sense to me. seems relatively harmless in a worst case.

I wonder if cutting to the minute even could be better? Idk what the polling interval is here already - is it <1min?

zhen0:
Ah!  I thought you wanted more than Craig's comment.  Oh well, a few videos to help. But yeah basically what Craig said!

Before:
https://github.com/PrefectHQ/prefect-ui-library/assets/40272060/10311258-a8eb-4279-b97d-1f1a8c2e79ee



After:
https://github.com/PrefectHQ/prefect-ui-library/assets/40272060/d4c10297-b2c8-4dbf-9c8d-5a8383d3e2ef

PR #1830, opened 2023-10-27
Fix: Remove empty state flicker on the dashboard

zhen0:
Adds a check to make the subscription is fully executed before we show the empty state

Current:

https://github.com/PrefectHQ/prefect-ui-library/assets/40272060/320c28c5-4101-49fb-8dd4-bad03ff816d4

Fix:

https://github.com/PrefectHQ/prefect-ui-library/assets/40272060/a1f358b8-b356-42f7-9c9a-eea5b061cb3c


pleek91 (approved):
Nice catch!

PR #1842, opened 2023-11-01
Fix: Hide flow run concurrency on editing push work pools

zhen0:
Push work pools do not allow concurrency limits. We hide the concurrency limit on creation of push work pools but it is still visible on edit.  This PR checks if a work pool is push type and hides the concurrency limit input.

### DIFF
--- a/src/components/WorkPoolEditForm.vue
+++ b/src/components/WorkPoolEditForm.vue
@@ -13,7 +13,7 @@
         </template>
       </p-label>
 
-      <p-label label="Flow Run Concurrency (Optional)">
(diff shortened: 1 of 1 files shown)


src/components/WorkPoolEditForm.vue
pleek91, line 1: I think `workPool.isPushPool` is `boolean` so I don't think we need the computed. `v-if="!workPool.isPushWorkPool"` should do the trick.

zhen0: Oh nice yeah!

pleek91 (approved):

PR #1850, opened 2023-11-06
Run graph v2 implementation

pleek91:
# Description
Swapping out all v1 graph code for v2 updates. Renames some things from "FlowRunTimeline" to "FlowRunGraph" to be consistent in our terminology. 

Cloud PR - https://github.com/PrefectHQ/nebula-ui/pull/3682
OSS - https://github.com/PrefectHQ/prefect/pull/11112


src/components/FlowRunGraph.vue
```diff
@@ -0,0 +1,141 @@
...
+  import { NodeSelection, RunGraph, RunGraphConfig, ViewportDateRange } from '@prefecthq/graphs'
+  import { useColorTheme } from '@prefecthq/prefect-design'
+  import { computed, ref } from 'vue'
+  import FlowRunGraphConfirmation from '@/components/FlowRunGraphConfirmation.vue'
+  import { useTaskRunsCount } from '@/compositions/useTaskRunsCount'
+  import { useWorkspaceApi } from '@/compositions/useWorkspaceApi'
+  import { FlowRun } from '@/models/FlowRun'
+  import { ServerStateType } from '@/models/StateType'
+
+  const OPT_IN_THRESHOLD = 2000
```
znicholasbrown, line 28: So this const is used to determine whether the graph loads automatically? And if the number of task runs exceeds the threshold we provide a disclaimer?

src/components/FlowRunGraph.vue
```diff
@@ -0,0 +1,141 @@
...
+  const selected = computed({
+    get() {
+      return props.selected
+    },
+    set(value) {
+      emit('update:selected', value)
+    },
+  })
+
+  // these will be replaced with brandon's styles
```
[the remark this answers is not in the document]
pleek91: These are the correct values though I believe. Safe to clean these up later? The actual implementation is kinda messy IMO because I think it uses the bg from dark mode and the text color from light mode or something like that

src/services/WorkspaceFlowRunsApi.ts
```diff
@@ -72,16 +70,40 @@ export class WorkspaceFlowRunsApi extends WorkspaceApi implements IWorkspaceFlow
...
-    return mapper.map('FlowRunGraphResponse', data, 'GraphNode')
-  }
-
-  public async getFlowRunsTimeline(id: string): Promise<GraphTimelineNode[]> {
-    const { data } = await this.get<FlowRunGraphResponse[]>(`/${id}/graph`)
+  public async getFlowRunsGraph(id: string): Promise<RunGraphData> {
+    // todo: should we just send this through the mapper?
+    // context: the idea is that this would be more efficient since the response can be very large.
+    // But that is unproven from a perf perspective. However this is much simpler than a dedicated mapper.
+    // But generally consistency is better so should test if this even makes a difference perf wise.
```
brandonreid, line 77: @pleek91 idk if we want to resolve this comment or not

pleek91: Its more a plead for others opinions lol. Unless someone thinks the mapper would be better I'd fine with shipping this.

pleek91: It should be faster because it maps while converting from json rather than going through the whole response twice (once to convert from json and once to map)

src/components/FlowRunGraph.vue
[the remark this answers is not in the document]
brandonreid: Yeah we can do that. FYI on the text colors, I believe all state colors work well with white text on them (well enough for the state tags at least, prob not always accessible but for another day).

brandonreid (approved):

src/components/FlowRunGraph.vue
znicholasbrown, line 74: Still happening?

src/services/WorkspaceFlowRunsApi.ts
```diff
@@ -72,16 +70,40 @@ export class WorkspaceFlowRunsApi extends WorkspaceApi implements IWorkspaceFlow
...
+
+        return new Map(value)
+      }
+
+      return value
+    }
+
+    const { data } = await this.get<RunGraphData>(`/${id}/graph-v2`, {
+      transformResponse: response => JSON.parse(response, reviver),
+    })
```
znicholasbrown, line 104: For consistency I think this should end up in the mapper - it's a little different from the way we map other payloads but I think the idea is the same - take an API-readable response and transform it into a UI-readable response.

znicholasbrown (approved):
A couple of minor comments, this largely matches what I expected - I've approved but would like to see the reviver functionality in the mapper and not the service layer

src/components/FlowRunGraph.vue
znicholasbrown: oh sorry i submitted without seeing the convo above on this, ignore this

src/components/FlowRunGraph.vue
znicholasbrown: This might be worth a note or a more descriptive variable name

src/services/WorkspaceFlowRunsApi.ts
pleek91: Yeah this approach was strictly for performance because this payload can have up to 10k nodes. But I agree the performance gains are likely minimal and not worth the inconsistency. 

I'll update this in a follow up PR.

src/components/FlowRunGraph.vue
pleek91: Updated

PR #1857, opened 2023-11-08
Map the run graph data rather than using a json reviver

pleek91:
# Description
Resolves this [comment](https://github.com/PrefectHQ/prefect-ui-library/pull/1850#discussion_r1387002088)

### DIFF
--- a/src/maps/runGraphData.ts
+++ b/src/maps/runGraphData.ts
@@ -1,0 +1,31 @@
+import { RunGraphData, RunGraphNode } from '@prefecthq/graphs'
+import { RunGraphDataResponse, RunGraphNodeResponse } from '@/models/api/RunGraphDataResponse'
+import { MapFunction } from '@/services/Mapper'
+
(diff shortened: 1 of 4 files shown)


src/maps/runGraphData.ts
```diff
@@ -0,0 +1,31 @@
...
+  return {
+    kind: source.kind,
+    id: source.id,
+    label: source.label,
+    state_type: source.state_type,
+    state_name: source.state_name,
+    start_time: this.map('string', source.start_time, 'Date'),
+    end_time: this.map('string', source.end_time, 'Date'),
+    parents: source.parents,
+    children: source.children,
```
znicholasbrown, line 15: Why retain snake casing here and in the data response?

znicholasbrown (approved):

src/maps/runGraphData.ts
pleek91: talked about this in slack. Might as well, but not going to address in this PR

PR #1875, opened 2023-11-14
Fix interval in FlowRunsFilteredList

pleek91:
# Description
The interval for flow runs was set to 3 seconds when it should be 30 seconds. Which was accidentally changed [here](https://github.com/PrefectHQ/prefect-ui-library/pull/1777/files#diff-b1aa283649729074ab8c72db7cf1838347d6f820d19c0be1b384c7fd24c62893R74).


zhen0 (approved):
Thank you!!

PR #1876, opened 2023-11-14
Add pagination to the block documents table component

pleek91:
# Description
Because the block documents table has always done search and filtering in the client we've been unable to do pagination. 

Related API PRs
- https://github.com/PrefectHQ/nebula/pull/6059 
- https://github.com/PrefectHQ/nebula/pull/5994 
- https://github.com/PrefectHQ/nebula/pull/6233

This PR includes all the front end changed necessary for searching, filtering, and pagination via the api. It also stores the current state of the page in the url so a refresh preserves the page, filters, and search. 

- [OSS PR](https://github.com/PrefectHQ/prefect/pull/11214)
- [Cloud PR](https://github.com/PrefectHQ/nebula-ui/pull/3767)


src/components/BlockDocumentsTable.vue
```diff
@@ -93,31 +95,34 @@
...
+  const searchTermDebounced = useDebouncedRef(searchTerm, 500)
 
-    return `${name} ${blockType.name} ${blockSchemaCapabilities.join(' ')}`.toLowerCase().includes(searchTerm.value.toLowerCase())
-  }
+  const { filter } = useBlockDocumentsFilterFromRoute({
+    blockSchemas: {
+      blockCapabilities: capabilities,
+    },
+    blockDocuments: {
+      nameLike: searchTermDebounced,
```
znicholasbrown, line 109: Does this filters get ignored if the key is included but the value is `null` or `''`?

znicholasbrown (approved):

src/components/BlockDocumentsTable.vue
pleek91: Yes, any falsey value is ignored https://github.com/PrefectHQ/prefect-ui-library/blob/main/src/maps/filters.ts#L60

PR #1878, opened 2023-11-15
combine flows pages into one new one

brandonreid:
Currently, OSS and Nebula have differing implementations for the Flows list. Each existing one has a couple advantages to it. This brings the best of both into one updated table.

A couple things I left off for now:
- Ability to sort by deployment counts (couldn't figure it out)
- A time toggle in the Activity header for selecting 8h, 24h, 1w - used by the mini charts. It felt a little unnecessary but open to opinions there:
image

### How it looks currently:

image


src/components/DeploymentsCount.vue
```diff
@@ -1,29 +1,26 @@
...
-    <template v-if="visible && deploymentsCount">
-      {{ deploymentsCount }}
-    </template>
-  </div>
+  <template v-if="deploymentsCount">
+    <span class="deployments-count">{{ deploymentsCount }} {{ toPluralString(localization.info.deployment, deploymentsCount) }}</span>
+  </template>
+  <template v-else>
+    <span class="deployments-count--none">{{ localization.info.none }}</span>
+  </template>
```
pleek91, line 7: I'm not 100% sure since this is a v-if v-else but I think you'll need to bind attrs manually since the root element can change.

src/styles/states.css
```diff
@@ -4,56 +4,91 @@
   dark:bg-state-completed-600
   dark:text-state-completed-50
 }
+.state-text--completed { @apply
+  text-state-completed-50
+  dark:text-state-completed-600
+}
```
brandonreid, line 10: I don't super love this CSS file. But I have intent to create a raft around updating state colors more broadly, so I know it's something that will be revisited.

src/components/DeploymentsCount.vue
```diff
@@ -35,4 +32,10 @@
 
   const deploymentsCountSubscription = useSubscriptionWithDependencies(api.deployments.getDeploymentsCount, deploymentsCountFilter)
   const deploymentsCount = computed(() => deploymentsCountSubscription.response ?? null)
```
pleek91, line 34: While you're here, there's a `useDeploymentsCount` composition that would simplify this quite a bit. 
```typescript
const { count } = useDeploymentsCount(() => ({
  flows: {
    id: [props.flowId]
  }
})
```

billpalombi:
I'm so glad to see that you are working on this. I don't think we need the ability to sort by deployment counts at all. A time toggle is a nice to have but not a must have.

The most important user story to solve for next is "show me all of the deployments of Flow foo". In an earlier deign we had an accordion style expansion, but linking the deployment count to the Deployments page with a filter applied for that Flow is another totally viable solution. Regardless, we don't have to wait to solve that user story to ship this.

src/components/FlowList.vue
```diff
@@ -1,236 +1,222 @@
...
-        </slot>
+        <p-pager v-if="pages > 1" v-model:page="page" :pages="pages" />
       </template>
     </p-layout-table>
   </div>
 </template>
 
 <script lang="ts" setup>
-  import { media } from '@prefecthq/prefect-design'
+  import { PTable, PEmptyResults, PLink, CheckboxModel, TableColumn } from '@prefecthq/prefect-design'
```
pleek91, line 90: Don't import these from prefect design. If you import them it'll always use the prefect-design version set in ui-library rather than using the peer installed in oss/cloud
```suggestion
```

src/components/FlowList.vue
```diff
@@ -1,236 +1,222 @@
...
 
   const props = defineProps<{
     filter?: FlowsFilter,
-    selectable?: boolean,
   }>()
 
-  const emit = defineEmits<{
-    (event: 'delete' | 'update', value?: string): void,
-  }>()
+  const activityTimeSpanInSeconds = 24 * 60 * 60
```
pleek91, line 116: date-fns has a const for this. `secondsInDay`.

src/components/FlowList.vue
```diff
@@ -1,236 +1,222 @@
...
-  })
+    limit: 50,
+  }))
 
-  const deploymentNameSearch = ref<string>()
-  const deploymentNameLikeDebounced = useDebouncedRef(deploymentNameSearch, 800)
+  const page = useRouteQueryParam('page', NumberRouteParam, 1)
+  const { flows, subscriptions, total, pages } = useFlows(filter, {
+    page,
+  })
```
pleek91, line 136: Should we add an interval to this? 
```suggestion
  const { flows, subscriptions, total, pages } = useFlows(filter, {
    page,
    interval: 30000
  })
```

brandonreid:
> In an earlier deign we had an accordion style expansion, but linking the deployment count to the Deployments page with a filter applied for that Flow is another totally viable solution.

Yeah I was starting to do this, but realized if you go to the flow page, there's a tab for deployments. If someone could help me figure out how to create a `p-link` that routes to the flow page with that tab selected, I think it would be nice to be able to jump to that.

src/components/LastFlowRun.vue
```diff
@@ -0,0 +1,22 @@
...
+    flowId?: string,
+  }>()
+
+  const filter = computed(() => ({
+    flows: {
+      id: [props.flowId],
+    },
+  })) as UnionFilter
+
+  const { flowRun } = useLastFlowRun(filter)
```
pleek91, line 21: Its preferred to just use the getter directly for this sort of thing rather than using a computed. 

```suggestion
  const { flowRun } = useLastFlowRun(() => ({
    flows: {
      id: [props.flowId],
    },
  }))
```

src/components/LastFlowRun.vue
```diff
@@ -0,0 +1,22 @@
...
+</template>
+
+<script lang="ts" setup>
+  import { computed } from 'vue'
+  import { FlowRunName } from '@/components'
+  import { useLastFlowRun } from '@/compositions'
+  import { UnionFilter } from '@/models'
+
+  const props = defineProps<{
+    flowId?: string,
```
pleek91, line 12: Why is this optional?

src/components/NextFlowRun.vue
```diff
@@ -0,0 +1,22 @@
...
+    flowId?: string,
+  }>()
+
+  const filter = computed(() => ({
+    flows: {
+      id: [props.flowId],
+    },
+  })) as UnionFilter
+
+  const { flowRun } = useNextFlowRun(filter)
```
pleek91, line 21:
```suggestion
  const { flowRun } = useNextFlowRun(() => ({
    flows: {
      id: [props.flowId],
    },
  }))
```

src/components/NextFlowRun.vue
```diff
@@ -0,0 +1,22 @@
...
+</template>
+
+<script lang="ts" setup>
+  import { computed } from 'vue'
+  import { FlowRunName } from '@/components'
+  import { useNextFlowRun } from '@/compositions'
+  import { UnionFilter } from '@/models'
+
+  const props = defineProps<{
+    flowId?: string,
```
pleek91, line 12: Same question here about this being optional

pleek91 (changes requested):
Overall this looks pretty good code wise. Few suggestions and questions but I think the only blocker is the prefect-design imports.

src/components/FlowList.vue
brandonreid: Ah ok, copy paste err. I'll still need to import the types though, unless I'm wrong there.

pleek91: Oh right the types are fine. Just not the components

src/components/DeploymentsCount.vue
brandonreid: Tested it out, the attrs do actually bind to either element here 👍

pleek91: Interesting, TIL. Thanks for double checking!

pleek91 (approved):
LGTM from a code perspective

billpalombi (approved):
Symbolically approving from UX perspective

zhen0:
Very excited to get this into the world.  And give the deployments page some similar ❤️

PR #1880, opened 2023-11-16
Make requests to GET endpoints when only a single value is batched

pleek91:
# Description
We use request batchers to reduce the number of requests being made. So if a single page requests 5 different flow runs only one `flow_runs/filter` request is made. 

Previously we would also hit the `flow_runs/filter` endpoint even if only a single flow run was batched. This applies to everywhere we use batchers. Updating the logic in each batcher to check for single id requests and hit the more efficient GET endpoints.

### DIFF
--- a/src/services/WorkspaceArtifactsApi.ts
+++ b/src/services/WorkspaceArtifactsApi.ts
@@ -22,12 +22,28 @@ export class WorkspaceArtifactsApi extends WorkspaceApi implements IWorkspaceArt
   protected override routePrefix = '/artifacts'
 
   private readonly batcher = new BatchProcessor<string, Artifact>(async ids => {
+    if (ids.length === 1) {
+      const [id] = ids
+      const { data } = await this.get<ArtifactResponse>(`/${id}`)
+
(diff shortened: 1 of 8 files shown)


src/services/WorkspaceArtifactsApi.ts
```diff
@@ -22,12 +22,28 @@ export class WorkspaceArtifactsApi extends WorkspaceApi implements IWorkspaceArt
...
+
     const artifacts = await this.getArtifacts({ artifacts: { id: ids } })
+
     return toMap(artifacts, 'id')
   }, { maxBatchSize: 200 })
 
   private readonly keyBatcher = new BatchProcessor<string, ArtifactCollection>(async keys => {
+    if (keys.length === 1) {
+      const [key] = keys
+      const { data } = await this.get<ArtifactCollectionResponse>(`/${key}/latest`)
```
pleek91, line 40: @nicolascribbles can you confirm this is correct

znicholasbrown: This and the other batcher both look correct to me

znicholasbrown (approved):

PR #1889, opened 2023-11-21
Fix schema object properties with empty strings

### LINKED ISSUE #11221: Unable to save work pool config after removing json from any `object` field with an existing value
### First check

### Bug summary

When editing a work pool with an existing value in an `object` type field, removing all the text from that value causes a 422 error. The `default` value in that field in the json view of the work pool is changed to an empty string.

```json
"labels": {
  "type": "object",
  "title": "Labels",
  "default": "",
  "description": "Labels applied to infrastructure created by a worker.",
  "additionalProperties": {
      "type": "string"
  }
}
```

### Reproduction

1. Go to the edit page for any work pool
2. Add some values for a field that's of type object
3. Save the work pool config
4. Edit the work pool again
5. Delete the values you added in the field
6. Save again

### Error

"Validation failed for field 'labels'. Failure reason: '' is not of type 'object'"

### Browsers

- [X] Chrome
- [ ] Firefox
- [X] Safari
- [ ] Edge

### Prefect version

_No response_

### Additional context

_No response_

pleek91:
# Description
When a json input is used for editing a property of type `object` a JsonInput is used which produces an empty string if a value is cleared by the user. Which isn't a valid value to send to the api and produces validation errors. Removing empty strings in the request mapper fixes this and is also how we're handling this same problem for [arrays](https://github.com/PrefectHQ/prefect-ui-library/blob/main/src/services/schemas/properties/SchemaPropertyArray.ts#L34). 

Closes https://github.com/PrefectHQ/prefect/issues/11221

### DIFF
--- a/src/services/schemas/properties/SchemaPropertyObject.ts
+++ b/src/services/schemas/properties/SchemaPropertyObject.ts
@@ -2,7 +2,7 @@ import { JsonInput } from '@/components'
 import { SchemaPropertyService } from '@/services/schemas/properties/SchemaPropertyService'
 import { SchemaPropertyComponentWithProps, getSchemaPropertyRequestValue, getSchemaPropertyResponseValue } from '@/services/schemas/utilities'
 import { SchemaValue, isSchemaValues, SchemaValues } from '@/types/schemas'
-import { isEmptyObject, isNullish, mapValues, parseUnknownJson, stringifyUnknownJson } from '@/utilities'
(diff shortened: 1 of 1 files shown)


znicholasbrown (approved):
`JSON.parse('')` should just return `undefined` 🙄

pleek91:
> `JSON.parse('')` should just return `undefined` 🙄

That would be nice sometimes. Though I understand why it just errors as invalid JSON.

PR #1890, opened 2023-11-22
Fix:  Use scheduled state name in nextRun composition

zhen0:
The nextRun composition is being used on the deployment page to get the next upcoming run.  However, it isn't actually picking up the next run/updating when the subscription is refreshed after a run is kicked off.  This PR updates the subscription to use the "scheduled" state name instead of filtering by time. 

The composition is also used in the flowListItem which I believe is currently in OSS but not in Cloud.  I can't see a strong reason for that to use time instead of state name - the idea of next run should be consistent across components.  Happy to discuss if I'm missing something there though.  

Current:

https://github.com/PrefectHQ/prefect-ui-library/assets/40272060/c1c651eb-8984-4d92-9be7-1f9368b9eddf

With fix:

https://github.com/PrefectHQ/prefect-ui-library/assets/40272060/bb42ced1-ab37-445d-a1bd-a1c50a277383

### DIFF
--- a/src/components/DeploymentQuickRunOverflowMenuItem.vue
+++ b/src/components/DeploymentQuickRunOverflowMenuItem.vue
@@ -11,7 +11,7 @@
   import { h } from 'vue'
   import { useRouter } from 'vue-router'
   import ToastFlowRunCreate from '@/components/ToastFlowRunCreate.vue'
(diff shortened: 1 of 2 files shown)


pleek91 (approved):
Ah this makes sense. Because the run button tries to refresh the next run subscription but its not getting the same subscription. Nice fix!

There's a few unused variables that could be removed that eslint picked up. But approving the functionality!

PR #1892, opened 2023-11-22
Enhancement: Make logs view more mobile friendly

zhen0:
Currently our logs view gets pretty squished on a small screen.  This PR adds a media query with more friendly handling for smaller screens. 

Current:
image

With update:
image

### DIFF
--- a/src/components/LogRow.vue
+++ b/src/components/LogRow.vue
@@ -45,6 +45,15 @@
   grid-template-columns: 84px minmax(0, 1fr) 150px;
 }
 
+@media (max-width: 768px) {
+  .log-row { @apply
+  grid
+  py-2;
+  grid-template-columns: none;
+  grid-template-rows: 20px minmax(0, 1fr) 50px;
+}
+}
+
 .log-row__leading { @apply
   select-none;
 }


src/components/LogRow.vue
```diff
@@ -45,6 +45,15 @@
...
 
+@media (max-width: 768px) {
+  .log-row { @apply
+  grid
+  py-2;
+  grid-template-columns: none;
+  grid-template-rows: 20px minmax(0, 1fr) 50px;
+}
+}
+
```
znicholasbrown, line 56: Thoughts on this instead?
```css
@media (max-width: 768px) {
  .log-row {
    grid-template-areas: 
      "leading trailing"
      "content content";
    grid-template-columns: 1fr 1fr;
    grid-template-rows: auto auto;
  }

  .log-row__leading { 
    grid-area: leading;
  }

  .log-row__trailing { 
    grid-area: trailing;
  }

  .log-row__content { 
    grid-area: content;
  }
}

```

which if I free-handed this correctly will mean we'd have:
[ level ] [ timestamp ]
[           content         ]

instead of:
[ level ]
[ content ]
_______ [ timestamp ]

znicholasbrown: and for even smaller screens:

```css
@media (max-width: 425px) {
  .log-row {
    grid-template-areas: 
      "leading"
      "trailing"
      "content";
    grid-template-columns: 1fr;
    grid-template-rows: auto auto auto;
  }

  .log-row__trailing { @apply
    text-left
  }
}
```

znicholasbrown (approved):
Approving since this is decidedly better than now but with a suggestion for how to improve it further

src/components/LogRow.vue
zhen0: I think it needs a bit of tweaking to get the log level button at the same height as the time and trailing info:

image

Going to merge as is right now and then work on getting your suggestion to fit.

PR #1899, opened 2023-11-28
Enhancement: Remove PLayoutTable in favor of PListHeader

brandonreid:
This PR cleans up instances of `PLayoutTable` for the new strategy of using `PContent` with `PListHeader`.

It also removes the `useOffsetStickyRootMargin` composition since that has been hoisted up to prefect-design, any instances of it's use in this package has been updated to import from it's new home.


pleek91 (approved):
Approving with the assumption that nothing really changed and this is just cleaning things up

PR #1905, opened 2023-12-01
Enhancement: Hide flow run concurrency controls for managed work pools

desertaxle:
Managed work pools don't currently support flow run concurrency, so we'll hide the control until support is added.


zhen0 (approved):
LGTM

PR #1906, opened 2023-12-01
Add support for `Suspended` state

### LINKED ISSUE #1903: Make UI library aware of `Suspended` state
In [prefect](https://github.com/PrefectHQ/prefect/issues/11271) we created a sub-state of `Paused` called `Suspended`. The UI mostly handles this fine, but we should make sure that `Suspended` is a first-class citizen along side `Paused`.

### LINKED ISSUE #1904: Update UI to say `Suspend` instead of `Pause`
When a flow run is executing we give the option to 'pause' the flow run. With the new nomenclature this is no longer a 'pause' but instead a 'suspend', we should update the UI accordingly.

bunchesofdonald:
We're splitting the concept of `Paused` into two separate ideas:

1. Paused, where execution of the flow run is paused within the process. This type of pause is only accomplishable from within the flow-run itself.
2. Suspended, where execution of the flow run is stopped and the process is exited. This type of pause is accomplishable from the UI as well as the flow run.

To that end this adds support for the `Suspended` state and changes the UI to use the `Suspended` state where it normally would have used the `Paused` state as the UI is unable to actually put a flow run into a `Paused` state as that can only occur within the flow run itself.

Closes #1903 
Closes #1904


pleek91 (approved):
Looks good to me! Thanks for taking the time to update all the paused references to suspended

PR #1918, opened 2023-12-06
Chore: Add an extra info tooltip component

zhen0:
Adds a component for giving help or extra info in a tooltip - as used recently in both work pools and deployment metrics. The aim is to keep consistency across uses. 

image


pleek91 (approved):
❤️

PR #1919, opened 2023-12-06
Enhancement: Give timestamp and timezone more space on the deployment schedule form

zhen0:
New:

image

Old:

image


src/components/IntervalScheduleForm.vue
```diff
@@ -154,5 +155,10 @@
     @apply
     col-span-3
   }
+
+  .interval-schedule-form__column--span-4 {
+    @apply
+    col-span-4
+  }
```
pleek91, line 162: I think you could just move these two out of the columns (leaving the cols just for the two things that need cols)

The p-content should take care of the row spacing.

pleek91 (approved):

PR #1920, opened 2023-12-06
Fix: Make the info on the retry modal more accurate

zhen0:
The retry modal currently states that we'll retry failed, crashed or cancelled state.  This is technically true but misses that we will also retry any run that does not have an output.  This PR updates the modal and includes a link to the results docs. 

image


pleek91 (approved):

zhen0:
@WillRaphaelson or @billpalombi - to check I've got the copy right here. Very small change but feels important to get it right.

PR #1922, opened 2023-12-08
Enhancement: Add a deployment list component 

zhen0:
Adds a new deployment list component to match the flow list component.  This would replace the DeploymentsTable component and be shown on the deployments page. 

image

I know we possibly have more great work coming for the deployments page but this felt like a quick upgrade to match flows. 

@brandonreid - for design feedback 

@WillRaphaelson and @billpalombi - I already showed Will but flagging for info


pleek91 (approved):
LGTM from a code perspective.

brandonreid:
@zhen0 This is 🔥!

I think this is definitely an improvement as is, but here are a few design decisions for your consideration:

image

- I believe we generally place status indicators after the text (in the case of the deployment name)
- In this mock, I’ve put the flow name below the deployment so there’s some proximity and tighter association there.
- This mock has a “Work Pool” column that also shows the work queue, this is from a previous design exploration, so feel free to include if it seems valuable, or totally dismiss.
- The “Applied by” name and the date have been combined into one “Created” column.
- The columns have been sorted a bit.

This makes me wonder if the Flows page is currently showing tags or not.. We probably should be. 🤔

billpalombi:
I agree with nearly all of @brandonreid's proposals. That said, I don't want the perfect to be the enemy of the good. @zhen0, I think the only blocking changes are:
- The status indicator does need to be to the right of the deployment name as it is elsewhere
- We should remove the applied by column entirely. I know that came from user feedback, but I think we over-indexed on it. That information does not belong on this page.

zhen0:
With Bill's updates:

image

zhen0:
I'm going to merge this as is and we can continue discussion about icon placement elsewhere if we think it important.

billpalombi:
I agree both with moving the status indicator and with treating that as a seperate PR. Thanks @zhen0 and @aaazzam!

PR #1932, opened 2023-12-12
BugFix: Flow runs card overflowing container and breaking layout

pleek91:
# Description
Fixes an issue on the flows page with the flow runs card where the flow runs could overflow the card and also have extremely large heights. Two part fix

1. Add a fixed height to the chart. This mirrors the dashboard implementation and fixes the bars having weird heights.
2. Add a check in the chart itself to prevent runs from being positioned outside of the max number of bars that should be displayed. 

Before
image

With fixed height
image

With fixed index
image

### DIFF
--- a/src/components/FlowRunHistoryCard.vue
+++ b/src/components/FlowRunHistoryCard.vue
@@ -43,7 +43,7 @@
 }
 
 .flow-run-history-card__chart { @apply
-  min-h-[20px]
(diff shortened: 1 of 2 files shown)


brandonreid (approved):
Ty!

znicholasbrown (approved):
nice fix

PR #1937, opened 2023-12-13
Update workspace dashboard filter

pleek91:
# Description
Updates the `WorkspaceDashboardFilter` to support the new date range select. Also adds a DateRangeSelect component for timezone support and a composition for interacting with the dashboard filter.


src/maps/dateRangeSelectValue.ts
```diff
@@ -0,0 +1,44 @@
...
+  timeSpanInSeconds: number,
+}
+
+function mapDateRangeSelectRangeValue({ startDate, endDate }: DateRangeSelectRangeValue): DateRange {
+  const timeSpanInSeconds = differenceInSeconds(endDate, startDate)
+
+  return { startDate, endDate, timeSpanInSeconds }
+}
+
+function mapDateRangeSelectSpanValue({ seconds }: DateRangeSelectSpanValue): DateRange {
```
znicholasbrown, line 26: Can we change the name of these to match the format other mappers are using?
`mapDateRangeSelectRangeValue` => `mapDateRangeSelectRangeValueToDateRange`
and
`mapDateRangeSelectSpanValue` => `mapDateRangeSelectSpanValueToDateRange`
? 

Also might not hurt to make these available to the global mapper 🤷🏻

znicholasbrown (approved):
lgtm, one non-blocking request

src/maps/dateRangeSelectValue.ts
pleek91: Updated the names. Didn't seem like adding them to the global mapper was worth it since there isn't currently a need for mapping these individually. The functions are just organization for the primary mapping function.

znicholasbrown: i thought you might feel that way, nbd imo

PR #1944, opened 2023-12-15
Fix: Improve handling for long deployment names on the deployments list

### LINKED ISSUE #11418: Unnecessary scrollbar in deployment overview
### First check

### Bug summary

The new deployment overview page layout reduces usability compared to previous releases.
This is because the deployment name column uses a div with fixed with that is just wide enough to display the `Created ...` info but not wide enough for the deployment name including the red or green dot to the right of the name that indicates if the deployment is actively polled.

image

Having to scroll on each deployment to see the full name and status is very inconvenient, especially if the total width of the table would allow to display of the name column without any scrollbar.

### Reproduction

Create a deployment with a name with more than 30 characers.

### Error

see screenshot

### Browsers

- [X] Chrome
- [X] Firefox
- [ ] Safari
- [ ] Edge

### Prefect version

```Text
Version:             2.14.10
API version:         0.8.4
Python version:      3.8.15
Git commit:          70888afe
Built:               Thu, Dec 7, 2023 7:27 PM
Server type:         cloud
```

### Additional context

Using Prefect Cloud

zhen0:
Long deployment names were cut off in the current display which means that the deployment status is also not shown. This PR switches deployment status back to the left and removes width restrictions to make sure full names are shown.  (Now that the 'Applied By' column has been removed there's more space on md plus screens). 

Will resolve [Unnecessary scrollbar in deployment overview#11418](https://github.com/PrefectHQ/prefect/issues/11418)
New:
image

Old:
image

@brandonreid


zhen0:
Should also now wrap names instead of scroll: 
image

znicholasbrown (approved):
this lgtm - i'll leave it up to your discretion when you want to merge or if you'd prefer to get design approval

zhen0:
I've got Brandon's changes in here so I tin we're good to merge.

PR #1945, opened 2023-12-15
Show input form if pause state requires it

abrookins:
Paused or suspended flow runs can now require that a user enter input to resume the run. If that's true, we now render a schema-based form on the resume modal.


src/components/FlowRunResumeModal.vue
```diff
@@ -3,10 +3,18 @@
     <p-label label="Current Flow Run State">
       <StateBadge :state="flowRun.state" />
     </p-label>
-    <div>
+
+    <div v-if="inputSchema">
+      <strong>Flow requires input.</strong> Please fill out the form below to resume.
```
pleek91, line 8: Should this say `Flow run requires input`?

pleek91 (approved):
LGTM!

src/components/FlowRunResumeModal.vue
abrookins: Flow run seems better, yeah!

PR #1952, opened 2023-12-19
Chore: Remove default value overrides

### LINKED ISSUE #10781: Flow submission with JSON parameters fails when using default values of parameter model
### First check

### Bug summary

When submitting flows with required parameters via the JSON parameter input field that match exactly the default values of the respective parameter model, the parameters are stripped from the input and validation consequently fails.

This is problematic for workflows where large JSON configurations dicts are used to configure flows. In this case, it is not trivial to spot that a provided parameter is equal to the default value.

### Reproduction

deploy

```
from prefect import flow
from pydantic import BaseModel


class FooModel(BaseModel):
    foo: str = "bar"


@flow
def main(
   foo: FooModel
) -> None:
    pass


if __name__ == "__main__":
    main()

```

run flow with JSON parameters that are equal to the model default parameters

image

### Error

```
Agent started! Looking for work from queue(s): default...
18:10:56.724 | INFO    | prefect.agent - Submitting flow run '68c64541-8642-4738-a8b5-a218188e0afa'
18:10:56.834 | INFO    | prefect.infrastructure.process - Opening process 'zeta6-sothis-y'...
18:10:56.860 | INFO    | prefect.agent - Completed submission of flow run '68c64541-8642-4738-a8b5-a218188e0afa'
...
18:10:58.798 | ERROR   | Flow run 'zeta6-sothis-y' - Validation of flow parameters failed with error:
Traceback (most recent call last):
  File ".../prefect/engine.py", line 422, in retrieve_flow_then_begin_flow_run
    parameters = flow.validate_parameters(flow_run.parameters)
  File ".../prefect/flows.py", line 456, in validate_parameters
    raise ParameterTypeError.from_validation_error(exc) from None
prefect.exceptions.ParameterTypeError: Flow run received invalid parameters:
 - foo: field required
18:10:59.082 | INFO    | prefect.infrastructure.process - Process 'zeta6-sothis-y' exited cleanly.

```

image

### Browers

- [X] Chrome
- [X] Firefox
- [ ] Safari
- [ ] Edge

### Prefect version

```Text
Version:             2.13.1
API version:         0.8.4
Python version:      3.8.15
Git commit:          d147b180
Built:               Thu, Sep 14, 2023 2:54 PM
Profile:             default
Server type:         ephemeral
Server:
  Database:          sqlite
  SQLite version:    3.37.2
```

### Additional context

_No response_

### LINKED ISSUE #11462: Custom run will always use override param value when flow default param value is supplied at custom run page.
### First check

### Bug summary

Custom runtime parameter values supplied in a custom run page (in the UI) revert to deployment-specific defaults instead of retaining the supplied values, a problem that seems to occur only when the supplied value matches the flow's default value. This does not affect deployment run via CLI that supply the same parameter values as what was passed to the UI page.

### Reproduction

```python
from prefect import flow


@flow(log_prints=True)
def my_f_repro_flow(
    param_a: str = "abc_def",
    param_b: str = "hij_klm",
    source_2_db_schema: str = "mva_output",
):
    print("param_a: ", param_a)
    print("param_b: ", param_b)
    print("source_2_db_schema: ", source_2_db_schema)
    return param_a + param_b


if __name__ == "__main__":
    my_f_repro_flow.deploy(
        name="numbers-deployment",
        work_pool_name="my-k8s-pool",
        image="taycurran/params-bug:repro",
        parameters={
            "param_a": "abc_def_ghi",
            "param_b": "789_012",
            "source_2_db_schema": "mva_output_prefect",
        },
    )
```

### Error

Your deployment should look like:
image

Now click on custom run:
image
^In the custom run page change a few of the values to the flow's default parameters (not the deployment specific parameters) then run the flow.

image
^You'll see that the values have gone back to the deployment specific values instead of the values I supplied in the custom run page.

**I only see this happening when the value supplied in the flow exactly matches the flow's default value.**

This also appears to not be an issue when I run the deployment via cli, leading me to believe this is a UI issue only.

### Browsers

- [X] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### Prefect version

```Text
Version:             2.14.11
API version:         0.8.4
Python version:      3.11.5
Git commit:          e6d7d76d
Built:               Thu, Dec 14, 2023 5:45 PM
OS/Arch:             darwin/arm64
Profile:             cm-demo
Server type:         cloud
```

### Additional context

```bash
prefect deployment run my-f-repro-flow/numbers-deployment --param param_a=abc_def --param param_b=789_012222 --param source_2_db_schema=mva_output
```

image
^this is correct.

znicholasbrown:
Removes request/response mapping logic that remove values that match schema defaults.
I'm not sure what relies on this behavior yet - opening this so we can test.

Potentially resolves: https://github.com/PrefectHQ/prefect/issues/11462
Potentially resolves: https://github.com/PrefectHQ/prefect/issues/10781

### DIFF
--- a/src/services/schemas/properties/SchemaPropertyAny.ts
+++ b/src/services/schemas/properties/SchemaPropertyAny.ts
@@ -3,7 +3,7 @@ import { getSchemaValueDefinition, schemaPropertyServiceFactory } from '@/servic
 import { SchemaPropertyService } from '@/services/schemas/properties/SchemaPropertyService'
 import { getSchemaPropertyDefaultValue, SchemaPropertyComponentWithProps } from '@/services/schemas/utilities'
 import { SchemaValue } from '@/types/schemas'
-import { isEmptyObject, sameValue } from '@/utilities'
+import { isEmptyObject } from '@/utilities'
 import { jsonSafeParse } from '@/utilities/jsonSafeParse'
 import { jsonSafeStringify } from '@/utilities/jsonSafeStringify'
 
@@ -72,10 +72,6 @@ export class SchemaPropertyAny extends SchemaPropertyService {
   }
 
   private referenceRequest(value: SchemaValue): SchemaValue {
-    if (this.isDefaultValueForReference(value)) {
-      return undefined
-    }
-
     const definition = getSchemaValueDefinition(this.property, value)
 
     if (definition === null) {
@@ -103,11 +99,4 @@ export class SchemaPropertyAny extends SchemaPropertyService {
 
     throw new Error('Could not find first definition for schema property')
   }
-
-  private isDefaultValueForReference(value: SchemaValue): boolean {
-    const definitions = this.property.anyOf ?? this.property.allOf ?? []
-
-    return definitions.some(definition => sameValue(value, getSchemaPropertyDefaultValue(definition)))
-  }
-
 }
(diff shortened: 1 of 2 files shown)


pleek91 (approved):
Yeah this makes sense to me. Curious if someone will have an issue with a runtime default. But this will probably solve a couple issues.

PR #1955, opened 2023-12-19
Update usePagination to correctly work with ref wrapped getters for parameters

pleek91:
# Description
`usePagination` wasn't correctly re-evaluating ref wrapped getters. This meant for any page that has a relative date filter wouldn't get the latest interval when the subscription was refreshed.

### DIFF
--- a/src/compositions/usePagination.ts
+++ b/src/compositions/usePagination.ts
@@ -1,5 +1,5 @@
 import { Getter } from '@prefecthq/prefect-design'
-import { SubscribeArguments, SubscriptionOptions, UseSubscription, useSubscription, useSubscriptionWithDependencies } from '@prefecthq/vue-compositions'
+import { SubscriptionOptions, UseSubscription, useSubscriptionWithDependencies } from '@prefecthq/vue-compositions'
 import { ComputedRef, MaybeRef, Ref, computed, onScopeDispose, reactive, ref, toRef, watch } from 'vue'
 import { GLOBAL_API_LIMIT } from '@/compositions/useFilterPagination'
 import { UseSubscriptions, useSubscriptions } from '@/compositions/useSubscriptions'
--- a/src/compositions/usePagination.ts
+++ b/src/compositions/usePagination.ts
@@ -1,6 +1,6 @@
 import { Getter } from '@prefecthq/prefect-design'
-import { SubscriptionOptions, UseSubscription, useSubscription, useSubscriptionWithDependencies } from '@prefecthq/vue-compositions'
-import { ComputedRef, MaybeRef, Ref, computed, onScopeDispose, reactive, ref, watch } from 'vue'
+import { SubscribeArguments, SubscriptionOptions, UseSubscription, useSubscription, useSubscriptionWithDependencies } from '@prefecthq/vue-compositions'
+import { ComputedRef, MaybeRef, Ref, computed, onScopeDispose, reactive, ref, toRef, watch } from 'vue'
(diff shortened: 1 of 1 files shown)


src/compositions/usePagination.ts
```diff
@@ -74,7 +74,7 @@ export function usePagination<
   const page = getPageRef()
   const pages = computed(() => Math.ceil(total.value / getLimit()))
 
-  const countSubscriptionParameters = computed(() => {
+  const countSubscriptionParameters = toRef(() => {
```
pleek91, line 77: using `toRef` rather than a computed ensures this callback gets re-evaluated when the value is accessed. This is important for how we do relative dates to work.

src/compositions/usePagination.ts
```diff
@@ -123,29 +127,27 @@ export function usePagination<
...
-    return repeat(page, index => getFetchFilterForPage(index + 1, filter))
+      if (parameters === null) {
+        return null
+      }
+
+      const [filter, ...rest] = parameters
+      const pageFilter = getFetchFilterForPage(page, filter)
+
+      return [pageFilter, ...rest]
+    }) as Ref<Parameters<TFetch> | null>
```
pleek91, line 150: Rather than generating every pages parameters statically via watchers this gets the parameters for a single page in a ref wrapped getter. This means every page has the ability to re-evaluate the parameters when the subscription is refreshed. Again, important for how relative time works.

src/compositions/usePagination.ts
```diff
@@ -84,18 +84,22 @@ export function usePagination<
...
 
-  const fetchSubscriptionParameters = ref(getFetchSubscriptionParameters()) as Ref<Parameters<TFetch>[] | null>
+  const fetchParametersRef = toRef(fetchParametersGetter)
   const fetchSubscriptions: UseSubscription<TFetch>[] = reactive([])
   const results = computed(() => fetchSubscriptions.flatMap(subscription => subscription.response ?? []) as unknown as Awaited<ReturnType<TFetch>>)
 
-  watch([total, fetchSubscriptionParameters], ([total, parameters]) => {
-    if (parameters === null || total === 0) {
+  watch([total, page, fetchParametersRef], ([total, page, parameters]) => {
+    if (total === 0 || page === 0 || parameters === null) {
```
znicholasbrown, line 92: I'm following the other two, but why `page === 0`? when would that occur?

src/compositions/usePagination.ts
znicholasbrown, line 150: I'm surprised this needs to be cast, what's up there?

znicholasbrown (approved):
a couple of smaller questions otherwise the implementation makes sense

src/compositions/usePagination.ts
pleek91: When in infinite mode the page starts at 0 and waits for the implementation to call `next`.

src/compositions/usePagination.ts
pleek91: Common mistake with our subscriptions and compositions. They're expecting `Ref<T>` but a wrapped getter produces `Readonly<Ref<T>>` so typescript complains about the ref being readonly. Something that would be good to fix at some point but it rarely comes up.

src/compositions/usePagination.ts
znicholasbrown: ahhhh ok that makes sense

src/compositions/usePagination.ts
znicholasbrown: got it, thanks

PR #1956, opened 2023-12-19
Updates to saved searches and flow run filters for the flow runs page date filter update

pleek91:
# Description
Necessary updates to ui-library components to support the new DateRangeSelect component on the flow runs page. Previously the flow runs page dealt strictly in static date ranges. And every saved search defaulted to the 7 days in the past 1 day in the future view. Now saved searches can also save the selected date range whether its a static date or a date interval. 

Implementation
- [Nebula UI](https://github.com/PrefectHQ/nebula-ui/pull/3944)
- [OSS](https://github.com/PrefectHQ/prefect/pull/11473)


src/components/FlowRunsFilterGroup.vue
pleek91, line 1: Updated this component to v-model the filter rather than accessing it from the url directly. Seems like a better pattern considering this component shouldn't really care where the filter is coming from.

src/components/PageHeadingFlowRuns.vue
pleek91, line 1: Again, exposing the v-model for `SavedFilters`

src/components/SaveFilterModal.vue
pleek91, line 1: Passing the filter in rather than assuming its coming from the url.

src/components/SavedFilters.vue
pleek91, line 1: Passing the current filter in as a prop. And simplifying the logic for the options and selected value.

demo/sections/flowRuns/FlowRunsFilterGroup.vue
```diff
@@ -1,7 +1,7 @@
 <template>
   <ComponentPage title="FlowRunsFilterGroup">
     <div class="flow-runs-filter-group__demo">
-      <FlowRunsFilterGroup />
+      <!-- <FlowRunsFilterGroup /> -->
```
brandonreid, line 4: Probably didn't want this commented out?

brandonreid (approved):
One comment, bit of an on the run review but looks good to me

src/utilities/savedFilters.ts
```diff
@@ -1,64 +1,34 @@
...
-  if (!isSame(asArray(filterA.deployment), asArray(filterB.deployment))) {
-    return false
-  }
-
-  if (!isSame(asArray(filterA.workPool), asArray(filterB.workPool))) {
-    return false
-  }
-
-  return true
+  return isEqual(filterA, filterB)
```
znicholasbrown, line 7: 🥇

src/maps/savedSearchCreate.ts
```diff
@@ -14,8 +16,49 @@ function mapSavedSearchFiltersToSavedSearchCreate(filters: SavedSearchFilter | u
...
+
+  return values
+}
+
+type SavedSearchProperty = 'flow' | 'tag' | 'deployment' | 'workQueue' | 'workPool' | 'state'
+
+function getObjectFilter(filters: SavedSearchFilter, property: SavedSearchProperty): SavedSearchFilterResponse | null {
+  const value = filters[property]
+
+  if (!value || isEmptyArray(value)) {
```
znicholasbrown, line 37: eslint is complaining about this one, does this need to be
```ts
if (isNullish(value) || isEmptyArray(value))
```
? (i forget if `isNullish` does list comparisons)

src/components/PageHeadingFlowRuns.vue
```diff
@@ -1,19 +1,34 @@
...
 <script lang="ts" setup>
+  import { computed } from 'vue'
   import PageHeading from '@/components/PageHeading.vue'
   import SavedFilters from '@/components/SavedFilters.vue'
+  import { SavedSearchFilter } from '@/models/SavedSearch'
 
-
-  defineProps<{
+  const props = defineProps<{
+    filter: SavedSearchFilter,
```
znicholasbrown, line 16: I'm surprised to see this as a prop here - feels like a not-ideal drill/emit situation. If the heading doesn't do anything but re-emit events from the `SavedFilters` component it might save some headache to expose the `#actions` slot for consumers to handle themselves.

src/components/FlowRunsFilterGroup.vue
```diff
@@ -1,64 +1,80 @@
...
   import WorkPoolCombobox from '@/components/WorkPoolCombobox.vue'
-  import { useRecentFlowRunsFilterFromRoute } from '@/compositions/filters'
+  import { SavedSearchFilter } from '@/models/SavedSearch'
 
-  const flowRunNameLike = ref<string>()
-  const flowRunNameLikeDebounced = useDebouncedRef(flowRunNameLike, 1200)
+  const props = defineProps<{
+    filter: SavedSearchFilter,
+    name: string,
+  }>()
```
znicholasbrown, line 47: This took me a second to understand - `name` is actually the prop used to filter the list based on flow run name, right? Maybe a more apt prop name would be `nameSearch` or `flowRunNameSearch`. As is, I couldn't figure out why a `FlowRunsFilterGroup` component needed a name

src/maps/savedSearchCreate.ts
znicholasbrown: If eslint is wrong, can you add an ignore?

src/components/FlowRunsFilterGroup.vue
znicholasbrown: Also I'm a little surprised that this isn't captured in the filter as an `ilike` property

pleek91: updated the name. The name search (along with the subflow runs toggle) are not included in saved searches. Which is why the name is a separate v-model here.

src/components/PageHeadingFlowRuns.vue
pleek91: Generally speaking I agree prop drilling and event bubbling isn't a great pattern. But this felt better to me so that the heading component could remain a single component for implementation. Otherwise each project would need to add the heading and add the correct actions. This way its all included and there's no chance for oss and cloud to be different on accident.

PR #1957, opened 2023-12-19
Enhancement: add 'All except scheduled' option to tags select

brandonreid:
Adds an option for selecting all but "scheduled". This allows the default filter set to look nicer on the deployments page and makes selecting all but scheduled much easier on other pages.

https://github.com/PrefectHQ/prefect-ui-library/assets/6776415/632173b9-abd9-49c3-a321-cbf95954c24e

Note – On the deployment page, since "all except scheduled" is the default, it sort of forces it back on if you uncheck it. Open to feedback there, but I feel like most folks won't do that.

### DIFF
--- a/src/components/FlowRunFilteredList.vue
+++ b/src/components/FlowRunFilteredList.vue
@@ -6,8 +6,8 @@
       <FlowRunsDeleteButton v-if="can.delete.flow_run" :selected="selected" @delete="deleteFlowRuns" />
 
       <template #controls>
-        <SearchInput v-model="searchTerm" placeholder="Search by run name" label="Search by run name" />
-        <StateNameSelect v-model:selected="filter.flowRuns.state.name" multiple empty-message="All run states" />
+        <SearchInput v-model="searchTerm" placeholder="Search by run name" label="Search by run name" class="flow-run-filtered-list__search" />
+        <StateNameSelect v-model:selected="filter.flowRuns.state.name" multiple empty-message="All run states" exclude-scheduled />
       </template>
 
       <template #sort>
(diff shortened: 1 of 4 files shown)


src/components/StateNameSelect.vue
pleek91, line 1: Clever to add this functionality just within this select without having to account for it everywhere. Good idea 👍

src/components/FlowRunFilteredList.vue
```diff
@@ -6,8 +6,8 @@
       <FlowRunsDeleteButton v-if="can.delete.flow_run" :selected="selected" @delete="deleteFlowRuns" />
 
       <template #controls>
-        <SearchInput v-model="searchTerm" placeholder="Search by run name" label="Search by run name" />
-        <StateNameSelect v-model:selected="filter.flowRuns.state.name" multiple empty-message="All run states" />
+        <SearchInput v-model="searchTerm" placeholder="Search by run name" label="Search by run name" class="flow-run-filtered-list__search" />
+        <StateNameSelect v-model:selected="filter.flowRuns.state.name" multiple empty-message="All run states" exclude-scheduled />
```
pleek91, line 10: This isn't actually a prop is it? 
```suggestion
        <StateNameSelect v-model:selected="filter.flowRuns.state.name" multiple empty-message="All run states" />
```

brandonreid: Ah nope, past attempt, will clean up.

src/components/FlowRunFilteredList.vue
```diff
@@ -77,4 +77,10 @@
...
     subscriptions.refresh()
   }
-</script>
\ No newline at end of file
+</script>
+
+<style>
+.flow-run-filtered-list__search { @apply
+  min-w-[224px]
+}
```
brandonreid, line 85: These are only semi related, but helps search maintain some width when the state input tries to take up a lot of horizontal space in the list headers.

pleek91 (approved):

PR #1960, opened 2023-12-20
Show server-side validation error if resume orchestration fails

abrookins:
There are times when server-side validation of a Pydantic model won't match the UI's client-side validation -- in particular, around handling optional fields, though there are probably other cases as well. That would be fine if we showed the errors, but we don't! Resuming a flow run is a form of orchestration, so the `/resume` endpoint returns an orchestration result response. The problem with this response type is that rejected state transitions come back as 200 responses with an error inside, and the error is "human-readable," not a JSON object.

In this PR, we check for this type of response and display the error on the form, with an 'error' toast message. Previously, we reported that the flow run resumed when in fact it didn't.

**NOTE:** All calls to `/set_state` also return orchestration result responses, but we don't appear to be checking for rejected transitions. In the future, we should audit calls to `set_state` and make sure we show appropriate errors.


pleek91 (approved):
LGTM. Though I am curious why the resume endpoint returns a 200 response when it didn't actually resume. I would have expected some sort of 400 response.

PR #1965, opened 2023-12-20
Enhancement: use count limit instead of time on flow and deployment histograms

brandonreid:
Recently we added the flow run history mini histogram chart for on the Flow and Deployment list pages. We opted to use a 24 hour period, but some users found this confusing. For flows that have a run cadence in the hours or days, they'd see no runs and wonder what went wrong, the 24 hour period was not communicated to them.

To fix this, since the notion of time on the x-axis of this history graph is fluid anyway, this PR ditches the fixed time range for these instances. Instead of runs in the past 24 hours, it'll simply show last `n` runs. Sometimes there aren't enough runs to fill the graph, so it'll still go through the process of inserting empty bars to communicate some rough notion of time. A minimum of 24 hours is used so that, if you run a flow twice in a row, you don't end up with something like a bar, 6 gaps, then the last bar.

Before:
image

After:
image

### DIFF
--- a/src/components/FlowRunsBarChart.vue
+++ b/src/components/FlowRunsBarChart.vue
@@ -119,9 +119,9 @@
   }
 
   function organizeFlowRunsWithGaps(flowRuns: FlowRun[]): (FlowRun | null)[] {
-    const { expectedStartTimeAfter, expectedStartTimeBefore } = toValue(props.filter).flowRuns ?? {}
+    const { expectedStartTimeAfter, expectedStartTimeBefore } = toValue(props.filter).flowRuns ?? getTimeRangeFromFlows(flowRuns)
 
-    if (!expectedStartTimeBefore || !expectedStartTimeAfter) {
+    if (!expectedStartTimeAfter || !expectedStartTimeBefore) {
       return []
     }
 
@@ -160,6 +160,27 @@
     })
 
     return buckets
+  }
+
+  function getTimeRangeFromFlows(flowRuns: FlowRun[]): { expectedStartTimeAfter: Date, expectedStartTimeBefore: Date } {
+    return flowRuns.reduce((acc, flowRun) => {
+      if (!flowRun.startTime) {
+        return acc
+      }
+
+      if (flowRun.startTime.getTime() < acc.expectedStartTimeAfter.getTime()) {
+        acc.expectedStartTimeAfter = flowRun.startTime
+      }
+
+      if (flowRun.startTime.getTime() > acc.expectedStartTimeBefore.getTime()) {
+        acc.expectedStartTimeBefore = flowRun.startTime
+      }
+
+      return acc
+    }, {
+      expectedStartTimeAfter: new Date(),
+      expectedStartTimeBefore: new Date(),
+    })
   }
 </script>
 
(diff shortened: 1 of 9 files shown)


src/components/FlowRunsBarChart.vue
```diff
@@ -161,6 +166,34 @@
 
     return buckets
   }
+
+  function getTimeRangeFromFlows(flowRuns: FlowRun[]): { expectedStartTimeAfter: Date, expectedStartTimeBefore: Date } {
```
pleek91, line 170: Is this just getting the min and max start times? This would be a lot simpler to use the `min` and `max`utilities from date-fns. 
```javascript
const maxDate = new Date()
const minDate = subSeconds(now, secondsInDay)
const startTimes = flowRuns.map(flowRun => flowRun.startTime)

return {
  expectedStartTimeAfter: min(minDate, ...startTimes),
  expectedStartTimeBefore: max(maxDate, ...startTimes)
}
```

src/components/FlowRunsBarChart.vue
```diff
@@ -119,9 +121,12 @@
   }
 
   function organizeFlowRunsWithGaps(flowRuns: FlowRun[]): (FlowRun | null)[] {
-    const { expectedStartTimeAfter, expectedStartTimeBefore } = toValue(props.filter).flowRuns ?? {}
+    const flowRunsFilter = toValue(props.filter).flowRuns
+    const { expectedStartTimeAfter, expectedStartTimeBefore } = flowRunsFilter?.expectedStartTimeAfter && flowRunsFilter.expectedStartTimeBefore
+      ? flowRunsFilter
+      : getTimeRangeFromFlows(flowRuns)
```
pleek91, line 127: This is pretty hard to grok. I'd extract this into its own method called `getTimeRange` and do something like this

```javascript
const { expectedStartTimeAfter, expectedStartTimeBefore } = toValue(props.filter).flowRuns ?? {}

if(expectedStartTimeAfter && expectedStartTimeBefore) {
  return { expectedStartTimeAfter, expectedStartTimeBefore }
}

return getTimeRangeFromFlows(flowRuns)
```

src/maps/deploymentStatsFilter.ts
```diff
@@ -11,25 +10,9 @@ export const mapDeploymentStatsFilterToFlowRunsFilter: MapFunction<DeploymentSta
...
-      expectedStartTimeAfter: subSeconds(now, source.timeSpanInSeconds),
-      expectedStartTimeBefore: now,
+      startTimeBefore: now,
     },
   }
 
   return filter
 }
-
-export const mapDeploymentStatsFilterToTaskRunsFilter: MapFunction<DeploymentStatsFilter, TaskRunsFilter> = function(source) {
```
brandonreid, line 22: Checked in this repo and in nebula / oss and found no use. Plus it was named to map to Deployments but was mapping to Flows.. I probably wrote it 😭

src/components/FlowRunsBarChart.vue
brandonreid: Lol, awe man, ty

pleek91 (approved):

PR #1978, opened 2023-12-30
Remove demo app

pleek91:
# Description
The demo app is a very large portion of this codebase. It was a very useful tool early on when we did a lot of development isolated in this project working on apis and components for features that didn't yet exist in the application. But we hardly use it anymore and it is becoming unnecessary. Updating mock apis and demo's whenever we update api services or change components feels like extra work for little to no value. So I put this PR together to remove it altogether. If anyone feels strongly we should keep this, please comment on this PR. Otherwise an approval is a vote to remove.


znicholasbrown:
Go ahead and remove the `netlify.toml` file as well - I'll spin down the demo site netlify-side ahead of that change

pleek91:
> Go ahead and remove the `netlify.toml` file as well - I'll spin down the demo site netlify-side ahead of that change

@znicholasbrown I did remove that file in this PR 👍 I assume its still trying to make a preview because that file is still on main.

znicholasbrown (approved):
Ah ok that makes sense, thank you

collincchoy (approved):
> Otherwise an approval is a vote to remove.

I'm a bit indifferent on this - but in that way my vote leans toward less is more.

PR #1982, opened 2024-01-04
Add utility for creating tuples with type guards

pleek91:
# Description
I've been creating and working with a lot of tuple types lately and was getting annoyed with the redundancy and also with writing type guards that have an `as` in them all over the place. Introducing a `createTuple` utility to simplify this a bit. 

```typescript
const { values, isValue } = createTuple(['foo', 'bar'])
type Tuple = typeof values[number]
```

Use destructuring assignment to name your tuple appropriately

```typescript
const {
  values: myTuple,
  isValue: isMyTuple
} = createTuple(['foo', 'bar'])

type MyTuple = typeof myTuple[number]
```


src/utilities/tuples.ts
```diff
@@ -0,0 +1,16 @@
+export type TupleType<T extends string[]> = {
+  tuple: T,
+  isTuple: (value: unknown) => value is T,
+}
+
+export function createTuple<const T extends string[]>(tuple: T): TupleType<T> {
+
+  function isTuple(value: unknown): value is T {
+    return tuple.includes(value as T[number])
+  }
```
collincchoy, line 10: This function seems confusing - or atleast i'm confused about it's use. 

```ts
const { tuple, isTuple } = createTuple(['a', 'b', 'c'])

const someValue = 'b'
if (isTuple(someValue)) {
  console.log(someValue)  // on hover, TS shows someValue as `'b' & ['a', 'b', 'c']` so it thinks it has array-like methods on it now too
}
```

It reads like this tells me whether a value is valid _for_ the tuple like if you want to check existence against the tuple in an idiomatic way (but why not just use `tuple.includes`?). Or i'm vastly misunderstanding what the tuple itself is

src/utilities/tuples.ts
```diff
@@ -0,0 +1,16 @@
+export type TupleType<T extends string[]> = {
+  tuple: T,
+  isTuple: (value: unknown) => value is T,
+}
+
+export function createTuple<const T extends string[]>(tuple: T): TupleType<T> {
```
collincchoy, line 6: as-is this would only support tuples of strings? not other types nor variadic tuples?

collincchoy (commented):
I pulled locally to play around with it. Some thoughts. 

1) The resulting tuple object is mutable which is really unexpected to me as typically tuples are immutable. The type checker _does_ narrow the value types correctly(so TS warns on trying to mutate with an invalid literal value) but it isn't really a tuple imo if I can `createTuple(['a', 'b', 'c']).tuple.push('a')`

2) If the main goal is to reduce friction with things like `['a', 'b'] as const`, the return type here is actually a little _different_. Aside from the runtime mutability, the `readonly` narrows the typing down too to remove most mutations.

```ts
  const values = ['a', 'b', 'c'] as const   //   const values = ['a', 'b', 'c'] as const
  const { tuple, isTuple } = createTuple(['a', 'b', 'c'])  // const tuple: ["a", "b", "c"]
```

I'm personally a little confused on the utility and semantics, but if this is something that'd be immediately useful to you then I'd be cool w/ it as-is but renaming it to something that's not `tuple`. it's more of a string literal array?

src/utilities/tuples.ts
collincchoy, line 6: `const T` - neat! I had to go look this up 😅

pleek91:
@collincchoy replying here to several of your comments/questions

> The resulting tuple object is mutable which is really unexpected to me as typically tuples are immutable. The type checker does narrow the value types correctly(so TS warns on trying to mutate with an invalid literal value) but it isn't really a tuple imo if I can createTuple(['a', 'b', 'c']).tuple.push('a')
Ah good catch. The `tuple` that is returned should be `Readonly<T>` which reproduces what we do now with these using `as const`. 

> If the main goal is to reduce friction with things like ['a', 'b'] as const, the return type here is actually a little different. Aside from the runtime mutability, the readonly narrows the typing down too to remove most mutations.
Yep totally, making the returned `Readonly<T>` should fix this as well. I assumed the `const T` handled this but it does not. 

> It reads like this tells me whether a value is valid for the tuple like if you want to check existence against the tuple in an idiomatic way (but why not just use tuple.includes?). Or i'm vastly misunderstanding what the tuple itself is
So this is supposed to be a utility for this pattern
```typescript
export const pendingStateType = ['scheduled', 'pending']
export type PendingStateType = typeof pendingStateType[number]

export function isPendingStateType(value: unknown): value is PendingStateType {
  return  pendingStateType.includes(value as PendingStateType)
}
```

which could be replaced by 
```typescript
const {
  tuple: pendingStateType
  isTuple: isPendingStateType
} = createTuple(['scheduled', 'pending'])
type PendingStateType = typeof pendindStateType[number]
```

Attempting to abstract away to `as const` (which sometimes gets forgotten) and the creating of the type guard (which is always identical and requires an `as` for typescript to be happy). 

I see the confusion of the type guard that is returned. Open to better naming there, possibly `isValue`? I think I just replaced the name of tuples like `pendingStateType` with `tuple` so `pendingStateTupe` becomes `tuple` and `isPendingStateType` became `isTuple`. But I think `isValue` is probably better?

src/utilities/tuples.ts
pleek91: I hadn't thought about making this more general. Its design is to simplify the creating of string tuple types. Though it could work with any value. See https://github.com/PrefectHQ/prefect-ui-library/pull/1982/commits/b52f531bac1cf6570e5429155e6abf2178eeeeca

src/utilities/tuples.ts
pleek91: > but why not just use tuple.includes?

Because `tuple.includes` won't accept `unknown` values
image

collincchoy (approved):
Cool I can dig it 👍 . 

I think a jsdoc comment could be really helpful to better explain what this is useful for for any other folks, but definitely not a showstopper. For me, this really clicked for "oh yeah i see now" where otherwise, the pattern it's meant to replace reads pretty straight forward and idiomatically

image

znicholasbrown (approved):

PR #1990, opened 2024-01-09
Fix bug with converting a saved search into a date range value

pleek91:
# Description
Fixes missing conditions when mapping a `SavedSearch` into a `DateRangeSelectValue` that was causing errors to be thrown and filters to not be usable. 

[Community slack thread](https://prefect-community.slack.com/archives/C0192RWGJQH/p1704777648479909)

Error that was thrown
image

### DIFF
--- a/src/components/SavedFiltersMenu.vue
+++ b/src/components/SavedFiltersMenu.vue
@@ -54,7 +54,7 @@
   }>()
 
   const emit = defineEmits<{
-    (event: 'update:selectedSearchOption', value: SavedSearch | null): void,
+    (event: 'update:selectedSearchOption', value: SavedFlowRunsSearch | null): void,
   }>()
 
   const attrs = useAttrs()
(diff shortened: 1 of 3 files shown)


src/components/SavedFiltersMenu.vue
```diff
@@ -54,7 +54,7 @@
   }>()
 
   const emit = defineEmits<{
-    (event: 'update:selectedSearchOption', value: SavedSearch | null): void,
+    (event: 'update:selectedSearchOption', value: SavedFlowRunsSearch | null): void,
```
pleek91, line 57: This is unrelated but noticed this event type didn't match the prop type.

src/models/api/SavedSearchResponse.ts
```diff
@@ -73,7 +73,7 @@ export function isDateRangeAroundResponse(value: unknown): value is DateRangeAro
 
 export type DateRangeResponse = DateRangeSpanResponse | DateRangeRangeResponse | DateRangePeriodResponse | DateRangeAroundResponse
 
-export function isDateRangeResponse(value: unknown): value is DateRangeRangeResponse {
+export function isDateRangeResponse(value: unknown): value is DateRangeResponse {
```
pleek91, line 76: This was the primary issue that caused the types to not error when they should have.

src/maps/savedSearch.ts
pleek91, line 1: The changes to this file fix the bug but also ensure that the types should catch any issues like this in the future before changes are merged.

collincchoy (approved):

src/components/SavedFiltersMenu.vue
collincchoy: 😭

PR #1995, opened 2024-01-10
Add a try catch when mapping saved filters in case of malformed data

pleek91:
# Description
Working through some ideas for community reports that I'm unable to reproduce. Since the saved search endpoint is a json blob there's always a chance that the data returned won't match the expected type. Adding a try catch around that to filter out any malformed filters.


znicholasbrown (approved):

znicholasbrown:
Yeah this makes sense

PR #1996, opened 2024-01-12
Enhancement: Store the last used log sort in local storage

pleek91:
# Description
Adds a new composition called `useLogsSort` that automatically syncs the users sort to local storage. That way if they refresh the page or navigate to a different flow fun or task run their sort is preserved. 

In reference to https://github.com/PrefectHQ/prefect/issues/8242


znicholasbrown (approved):
logs gettin hella love today, nice

toro-berlin:
Nice one. Will greatly improve our everyday workflow with Perfect!

PR #2004, opened 2024-01-16
Enhancement: use `PCombobox` instead of `PSelect` for enum-backed schema properties

znicholasbrown:
`PCombobox` works as a drop-in replacement for `PSelect` (since it wraps the component under the hood) BUT also provides the ability to search/filter over any of the params in-memory. This PR replaces calls to `PSelect` in the schema property definitions with `PCombobox` to take advantage of this:

Before (multiselect):
Screenshot 2024-01-16 at 4 57 05 PM

After (multiselect):
Screenshot 2024-01-16 at 5 18 14 PM

Before:
Screenshot 2024-01-16 at 5 20 33 PM

After:
Screenshot 2024-01-16 at 5 20 09 PM

### DIFF
--- a/src/services/schemas/utilities.ts
+++ b/src/services/schemas/utilities.ts
@@ -2,6 +2,7 @@ import { JsonInput } from '@/components'
 import { isBlockDocumentReferenceValue, isBlockDocumentValue } from '@/models'
 import { schemaPropertyServiceFactory } from '@/services/schemas/properties'
 import { SchemaProperty, SchemaPropertyInputAttrs, Schema, SchemaValues, SchemaValue, schemaHas, SchemaPropertyAnyOf, SchemaPropertyAllOf } from '@/types/schemas'
+import { isArray } from '@/utilities'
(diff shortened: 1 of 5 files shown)


src/services/schemas/utilities.ts
```diff
@@ -113,6 +114,10 @@ export function getSchemaPropertyPlaceholder(property: SchemaProperty): string |
     return placeholder
   }
 
+  if (isArray(placeholder)) {
+    return placeholder.join(', ')
+  }
```
pleek91, line 119: What's this for? Surprised a placeholder could ever be an array

pleek91 (approved):

src/services/schemas/utilities.ts
znicholasbrown: It's cause we get the placeholder from the default value and stringify it below this; we've not noticed it cause pselect doesn't use the placeholder anywhere but if it did placeholder values would otherwise display as `["gratis"]`

PR #2008, opened 2024-01-17
Fix the font size for schema input descriptions

pleek91:
# Description
The markdown component used to render property descriptions defaults to font-md but label descriptions are supposed to be font-sm. Updating the input component to specific the font size fixes the descriptions so they're the correct size. 

Before
image

After
image


znicholasbrown (approved):

znicholasbrown:
Nice catch, thanks @pleek91

PR #2013, opened 2024-01-18
Chore: Check the component type before returning a joined placeholder

znicholasbrown:
#2004 slightly changed the way placeholders for list properties were displayed by joining list defaults. This didn't account for the fact that certain unstructured lists use only `JsonInput` components and so should in fact display the placeholder as JSON. This update checks the component type before returning the joined placeholder list and falls back to the stringified default as previous in cases where the JsonInput is in use.

### DIFF
--- a/src/services/schemas/utilities.ts
+++ b/src/services/schemas/utilities.ts
@@ -114,7 +114,7 @@ export function getSchemaPropertyPlaceholder(property: SchemaProperty): string |
     return placeholder
   }
 
-  if (isArray(placeholder)) {
+  if (isArray(placeholder) && property.meta?.component !== JsonInput) {
     return placeholder.join(', ')
   }
 


src/services/schemas/utilities.ts
```diff
@@ -114,7 +114,7 @@ export function getSchemaPropertyPlaceholder(property: SchemaProperty): string |
     return placeholder
   }
 
-  if (isArray(placeholder)) {
+  if (isArray(placeholder) && property.meta?.component !== JsonInput) {
```
pleek91, line 117: this is kinda counter intuitive. But it makes sense

pleek91 (approved):

src/services/schemas/utilities.ts
znicholasbrown: Agreed not the finest but seems ok for now

jakekaplan (approved):
🙏

PR #2019, opened 2024-01-22
BugFix: Make sure to preserve the value for `subFlowRunsExist` in `useTaskRunsCount` 

pleek91:
# Description
`undefined` is a valid value for the `taskRuns.subFlowRunsExist`. But `merge` will not copy an `undefined` value over an existing value. So the base filter in `useTaskRunsCount` would always be used. Specifically checking for the existence of the property prevents overriding that value. 

This fixes the flow run graph not being shown if the only nodes on the graph are subflow runs.


zhen0 (approved):
Thank you!

PR #1821, opened 2023-10-21
Enhancement: Add a block creation modal option

zhen0:
Gives the option to use a modal for block creation in the BlockDocumentInput and BlockCapabilityBlockDocumentInput components. 

Example use for automation notifications:
https://github.com/PrefectHQ/prefect-ui-library/assets/40272060/1763477b-6ad7-4ee2-a40e-750684b1cbc5

Example use for automation webhooks:

https://github.com/PrefectHQ/prefect-ui-library/assets/40272060/aa398c80-8469-4c3a-946e-47fc844e35b9

PR #2024, opened 2024-01-23
Enhancement: Always create blocks in modals from the block document input

pleek91:
# Description
When selecting a block in a schema form there's an "Add" button that allows you to create a new block to use. However this usually takes you away from the current page you're on and often loses the progress you've made on the current form. 

image

The ability to create the new block in a modal was added in https://github.com/PrefectHQ/prefect-ui-library/pull/1821 but it wasn't the default option because at the time we didn't have a way to handle nested blocks (where you cannot create a block without also creating a second block). https://github.com/PrefectHQ/prefect-design/pull/1093 gives us the ability to nicely open multiple modals at once. So now if a modal is used to create a block, and a second block needs to be created, that second block can also be created in a modal without creating an ugly UX.

This PR updates the "Add" button to always open in a modal. That way the user never loses their work and doesn't have to leave the current page. This will make creating work pools and block a much better user experience. 

image

### DIFF
--- a/src/components/BlockCreateModal.vue
+++ b/src/components/BlockCreateModal.vue
@@ -58,14 +58,6 @@
     'block-create-modal--schema-creation': blockType.value,
   }))
 
-  const title = computed(() => {
-    if (props.providedBlockType) {
-      return `Add a new ${props.providedBlockType.name} block`
-    }
-
-    return 'Add a new block'
-  })
-
   const handleAdd = (selectedBlockType: BlockType): void => {
     blockType.value = selectedBlockType
   }
--- a/src/components/BlockCreateModal.vue
+++ b/src/components/BlockCreateModal.vue
@@ -1,5 +1,5 @@
 <template>
-  <p-modal v-model:showModal="internalShowModal" class="block-create-modal" :class="modalClass" title="Add a new block">
+  <p-modal v-model:showModal="internalShowModal" class="block-create-modal" :class="classes" title="Add a new block">
     <BlockTypeList
       v-if="!blockType"
       :capability="capability"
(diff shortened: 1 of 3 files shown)


src/components/BlockCreateModal.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <p-modal v-model:showModal="internalShowModal" class="block-create-modal" :class="modalClass" title="Add a new block">
+  <p-modal v-model:showModal="internalShowModal" class="block-create-modal" :class="classes" title="Add a new block">
```
znicholasbrown, line 2:
```suggestion
  <p-modal v-model:showModal="internalShowModal" class="block-create-modal" :class="classes" :title="title">
```

src/components/BlockCreateModal.vue
```diff
@@ -44,13 +52,18 @@
 
   const blockType = ref(props.providedBlockType)
   const capability = ref(props.capability ?? '')
+  const formId = randomId()
```
znicholasbrown, line 55: What's this id on the block schema create form doing?

src/components/BlockCreateModal.vue
pleek91: Whoops, I actually means to keep the static title and use the heading in the form itself. I'll delete the unused title computed. 👍

src/components/BlockCreateModal.vue
pleek91: So by adding an id to the form itself the submit button (which is outside of the form in the modal's actions) can submit the form. Something I've seen collin do that makes the modal look much nicer. The submit button has `:form-id="formId"`.

znicholasbrown: ahhh ok that makes sense, i hadn't looked at that, thanks!

znicholasbrown (approved):

PR #2027, opened 2024-01-24
Add new types and type guards for schemas v2

pleek91:
# Description
Adds new `Schema` and `SchemaValues` types. This are paired down from our existing types to remove things that are not supported or no longer used by the ui.


src/schemas/types/schema.ts
```diff
@@ -0,0 +1,55 @@
...
+
+export const { values: schemaStringFormat, isValue: isSchemaStringFormat } = createTuple([
+  'date',
+  'date-time',
+  // 'time-delta',
+  'regex',
+  // 'email',
+  // 'json-string',
+  'password',
+])
```
pleek91, line 14: @znicholasbrown the commented out values are one's we had in the previous types that I don't know if we can actually produce from a pydantic type. Hoping you can help me validate that.

znicholasbrown: Hm those came from Pydantic docs somewhere but I can't find it quickly. Let's start without; we've got clear paths to extend functionality everywhere so it probably makes more sense to focus on real primitives for now instead of ones that are infrequently used/unclear

pleek91: Agreed, and worse case the user will get a text input.

znicholasbrown (approved):

src/schemas/types/schemaValues.ts
```diff
@@ -0,0 +1,66 @@
+import { isRecord, isString } from '@/utilities'
+import { createTuple } from '@/utilities/tuples'
+
+export type SchemaValues = Record<string, unknown>
+
+export const { values: prefectKinds, isValue: isPrefectKind } = createTuple([
+  null,
+  'json',
```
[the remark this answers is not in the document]
pleek91: Correct, it just allows us to have an object to stick other things in if we need to. But it has no hydration functionality, the value is used directly.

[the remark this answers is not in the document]
znicholasbrown: makes sense, thank you

PR #2028, opened 2024-01-25
Core components for Schemas V2

pleek91:
# Description
Adds the component structure for rendering dynamic forms based on a prefect schema. I stopped when I got to things like `anyOf` and `allOf` because this PR was already getting pretty large. But a lot of this is boilerplate props and v-models. 

Important distinction between this PR and the existing schema components is
- No client side validation (no vee-validate OR useValidation)
- v-models 🐐 
- Intentionally not supporting every possibility OpenApi allows for but optimizing for schemas that are actually possible to generate from Pydantic
- Not mapping schemas or schema values
- Not pre resolving definitions (they are resolved as needed)

### DIFF
--- a/src/schemas/components/SchemaFormProperty.vue
+++ b/src/schemas/components/SchemaFormProperty.vue
@@ -1,0 +1,54 @@
+<template>
+  <p-label class="schema-form-property" :label="label">
+    <template v-if="description" #description>
+      <div class="schema-form-property__description">
+        <p-markdown-renderer :text="description" class="schema-form-input__markdown" />
+      </div>
+    </template>
+
+    <SchemaFormPropertyInput v-model:value="value" :property="property" />
+  </p-label>
+</template>
+
+<script lang="ts" setup>
+  import { computed } from 'vue'
+  import SchemaFormPropertyInput from '@/schemas/components/SchemaFormPropertyInput.vue'
+  import { SchemaProperty } from '@/schemas/types/schema'
+  import { SchemaValue } from '@/schemas/types/schemaValues'
+
+  const props = defineProps<{
+    property: SchemaProperty,
+    value: SchemaValue,
+    required: boolean,
+  }>()
+
+  const emit = defineEmits<{
+    'update:value': [SchemaValue],
+  }>()
+
+  const value = computed({
+    get() {
+      return props.value
+    },
+    set(value) {
+      emit('update:value', value)
+    },
+  })
+
+  const label = computed(() => {
+    const title = props.property.title ?? ''
+
+    if (!props.required) {
+      return `${title} (Optional)`.trim()
+    }
+
+    return title
+  })
+
+  const description = computed(() => {
+    const { description = '' } = props.property
+    const descriptionWithNewlinesRemoved = description.replace(/\n/g, ' ')
+
+    return descriptionWithNewlinesRemoved
+  })
+</script>
--- a/src/schemas/components/SchemaFormProperty.vue
+++ b/src/schemas/components/SchemaFormProperty.vue
@@ -47,7 +47,7 @@
 
   const description = computed(() => {
     const { description = '' } = props.property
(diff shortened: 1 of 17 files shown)


src/schemas/index.ts
```diff
@@ -0,0 +1,3 @@
+import SchemaForm from '@/schemas/components/SchemaForm.vue'
+
+export { SchemaForm as SchemaFormV2 }
```
pleek91, line 3: Exporting as SchemaFormV2 to not conflict with the existing SchemaForm component

src/schemas/utilities/definitions.ts
```diff
@@ -0,0 +1,8 @@
+import { Schema, SchemaDefinition, SchemaProperty } from '@/schemas/types/schema'
+
+export function getSchemaDefinition(schema: Schema, definition: SchemaDefinition): SchemaProperty {
+  const definitionKey = definition.replace('#/definitions/', '')
+  const definitionSchema = schema.definitions?.[definitionKey]
+
+  return definitionSchema ?? {}
```
znicholasbrown, line 7: Under what circumstances would there be no definition? I'd almost rather this throw an error that we can handle in the component instead of passing back what I have to assume is an invalid definition (though maybe not strictly invalid from a TS perspective)

src/schemas/components/SchemaFormProperty.vue
```diff
@@ -0,0 +1,54 @@
...
+    set(value) {
+      emit('update:value', value)
+    },
+  })
+
+  const label = computed(() => {
+    const title = props.property.title ?? ''
+
+    if (!props.required) {
+      return `${title} (Optional)`.trim()
```
znicholasbrown, line 42: This is a nitpick but I'd like to see the `(Optional)` label treated as less important than the title of the property itself, ESPECIALLY because these tend to get displayed a lot in the same contexts and they can be visually noisy. Could be as simple as making the optional text subdued or approaching it from the opposite direction and adding a chip for "Required" fields. Food for thought

src/schemas/components/SchemaFormProperty.vue
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    return title
+  })
+
+  const description = computed(() => {
+    const { description = '' } = props.property
+    const descriptionWithNewlinesRemoved = description.replace(/\n/g, ' ')
+
+    return descriptionWithNewlinesRemoved
```
znicholasbrown, line 52: What's going on here? This doesn't feel like it should be necessary

znicholasbrown (approved):
couple comments, overall very clean.

src/schemas/components/SchemaFormProperty.vue
pleek91: Agreed. Lets handle that in a future PR 👍

src/schemas/components/SchemaFormProperty.vue
pleek91: Fixes this
image
To display this
image

These descriptions come directly from code comments. And those code comments usually have line breaks in them to format them at a certain line length. But then in the ui they break in weird places and have gaps in them like there are multiple paragraphs.

src/schemas/utilities/definitions.ts
pleek91: Agreed, I'm fine with just throwing an error here rather than returning an empty definition.

pleek91: Updated

src/schemas/components/SchemaFormProperty.vue
znicholasbrown: Ah I see, ok that makes sense. How would you feel about replacing the regex with this:
```
/\n(?!\n)/g
```

That'll do a negative lookahead to make sure we're not incorrectly stripping newline characters that would mark paragraph boundaries

znicholasbrown (approved):

PR #2030, opened 2024-01-26
Fix create block modal not submitting the form

### LINKED ISSUE #2029: Bug: Sometimes cannot create a new block during automation creation
## Describe the issue
When creating an automation, I picked "Send a notification" as the action, went to add a new block, add a new email block, enter information, and then click `Create` but nothing happens. No notable console logs either.

https://github.com/PrefectHQ/prefect-ui-library/assets/22418768/d9134044-eda6-43c1-b5b8-b65627f0aadf

## What did you expect to happen?
I expected the block to be created or if an error occurred for _something_ to happen. I don't see any network activity on click either.

## What else did you try?
I tried to reproduce in other environments to no avail (as in it worked) - so perhaps it's something account/workspace or environment specific. 

I went to the Blocks page instead and tried to create using the add a new block form (outside of the modal). It populated my previous inputs as expected and on `Create`, successfully created the block. So I think the content I had was _valid_.

pleek91:
# Description
The submit button on the create block modal had the wrong attribute for linking the submit button to the form. Fixing that and also not showing the submit button unless the block schema form is being showing. 

closes: https://github.com/PrefectHQ/prefect-ui-library/issues/2029

### DIFF
--- a/src/components/BlockCreateModal.vue
+++ b/src/components/BlockCreateModal.vue
@@ -16,8 +16,8 @@
       <BlockSchemaCreateForm v-if="blockSchema" :id="formId" :block-schema="blockSchema" hide-footer @submit="submit" />
     </template>
 
-    <template #actions>
-      <SubmitButton action="Create" :form-id="formId" />
+    <template v-if="blockSchema" #actions>
+      <SubmitButton action="Create" :form="formId" />
     </template>
   </p-modal>
 </template>


src/components/BlockCreateModal.vue
```diff
@@ -16,8 +16,8 @@
       <BlockSchemaCreateForm v-if="blockSchema" :id="formId" :block-schema="blockSchema" hide-footer @submit="submit" />
     </template>
 
-    <template #actions>
-      <SubmitButton action="Create" :form-id="formId" />
```
collincchoy, line 20: 🤕 

i... wonder how this worked for me "sometimes" but 🤷 .

collincchoy (approved):

src/components/BlockCreateModal.vue
pleek91: Yeah same, it worked fine for me before. But sure seems like it shouldn't have ever worked.

PR #2032, opened 2024-01-26
Add basic support for `anyOf` for schema properties

pleek91:
# Description
Adds support for schema properties that are an `anyOf` definition. Supports switching between schemas while preserving user input for each schema. Switching between schemas restores any previously entered values. 

Note: This does not address matching existing values to their proper schema. That will be addressed in a future PR.

### DIFF
--- a/src/schemas/components/SchemaFormPropertyAnyOfInput.vue
+++ b/src/schemas/components/SchemaFormPropertyAnyOfInput.vue
@@ -1,12 +1,67 @@
 <template>
-  <span>TODO</span>
+  <div class="schema-form-property-any-of-input">
+    <p-button-group v-model="selected" :options="options" small />
+  </div>
+
+  <p-card>
+    <SchemaFormProperty v-model:value="value" :property="property" :required="required" />
+  </p-card>
 </template>
 
 <script lang="ts" setup>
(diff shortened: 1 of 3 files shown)


src/schemas/components/SchemaFormPropertyAnyOfInput.vue
```diff
@@ -1,12 +1,80 @@
...
+    required: boolean,
+  }>()
+
+  const emit = defineEmits<{
+    'update:value': [SchemaValue],
   }>()
+
+  const value = computed({
+    get() {
+      return propertyValues[selectedPropertyIndex.value]
```
znicholasbrown, line 33: This might be a can of worms but is the order of `anyOf` property definitions guaranteed? Could the order change between requests?

znicholasbrown (approved):

src/schemas/components/SchemaFormPropertyAnyOfInput.vue
pleek91: I don't think we should consider the order to be guaranteed. But it shouldn't matter here. Once we set the initial index things will stay in sync as the user switches between definitions.

PR #2033, opened 2024-01-26
Fix property.required sometimes being null 

pleek91:
# Description
When building meta for a schema property we're also checking if the property is required. `required` is an optional field on the parent property and its being defaulted to an empty array.

We had a report of a user's deployments not loading and the error traced down to `required` being `null` rather than `undefined`. I'm not sure why it would be `null` but we generally treat `null` and `undefined` the same. So updating to use a safer check in case either `required` or `properties` are null.


zangell44 (approved):
I don't think this counts

zhen0 (approved):

PR #2037, opened 2024-01-29
Schemas: Set the initial definition index of an `anyOf` property based on the value present

pleek91:
# Description
In order to maintain backwards compatibility we need to able to match a value for an `anyOf` property with the correct definition. So adding that logic to the new schemas codebase. Mostly copied/inspired by the existing [getSchemaValueAllOfDefinition](https://github.com/PrefectHQ/prefect-ui-library/blob/main/src/services/schemas/utilities.ts#L200) in the current codebase. This should ensure we have at least as good of support as we do today for this.

### DIFF
--- a/src/schemas/components/SchemaFormPropertyAnyOf.vue
+++ b/src/schemas/components/SchemaFormPropertyAnyOf.vue
@@ -11,12 +11,13 @@
 <script lang="ts" setup>
   import { ButtonGroupOption } from '@prefecthq/prefect-design'
   import { computed, reactive, ref } from 'vue'
+  import { useWorkspaceApi } from '@/compositions'
   import SchemaFormProperty from '@/schemas/components/SchemaFormProperty.vue'
   import { useSchema } from '@/schemas/compositions/useSchema'
   import { SchemaProperty, isPropertyWith } from '@/schemas/types/schema'
   import { SchemaValue } from '@/schemas/types/schemaValues'
   import { getSchemaDefinition } from '@/schemas/utilities/definitions'
-  import { getSchemaPropertyLabel } from '@/schemas/utilities/properties'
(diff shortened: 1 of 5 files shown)


src/schemas/components/SchemaFormPropertyAnyOf.vue
```diff
@@ -39,7 +43,18 @@
...
+  const initialSelectedPropertyIndex = await getInitialIndexForSchemaPropertyAnyOfValue({
+    schema,
+    property: props.property,
+    value: props.value,
+    api,
+  })
+  
+  if(initialSelectedPropertyIndex === -1) {
+    throw 'not implemented'
+  }
```
pleek91, line 55: Handling this will be some implementation around switching to json.

src/schemas/utilities/properties.ts
```diff
@@ -22,4 +28,103 @@ export function getSchemaPropertyLabel(property: SchemaProperty): string {
...
+  }
+
+  const definitions = getSchemaPropertyAllOfDefinitions(property, schema)
+
+  // block documents are the only reason this utility is async.
+  // if the value is a block document reference we need to fetch the block document from the api
+  // to determine the block type
+  if (isBlockDocumentReferenceValue(value)) {
+    return await getBlockDocumentReferenceDefinitionIndex(value, definitions, api)
+  }
```
pleek91, line 63: The current implementation of this just grabs the first definition that is for a block type. But the example I was working off of is the dbt CLI Profile block which has a union of block types. This was the cleanest way I could think of the ensure the correct block definition was chosen. Shouldn't happen often so I'm fine with making an extra request here.

znicholasbrown (approved):

PR #2042, opened 2024-01-30
chore: delete duplicate deployments list

brandonreid:
This PR deletes the `DeploymentsTable` component in favor of just using the `DeploymentsList` component everywhere. 

The `prefix` prop used on the table component has been ported to the list for consistency in implementation on the flows page.


pleek91:
Idk that the prefix is necessary but since that's the existing functionality it would be a simple change to support the prefix on the list component. Are you apposed to adding that and preserving the urls?

pleek91:
For context the prefix allows for having multiple "union filters" on the same page. Though we haven't really leveraged that much.

brandonreid:
> Idk that the prefix is necessary but since that's the existing functionality it would be a simple change to support the prefix on the list component. Are you apposed to adding that and preserving the urls?

@pleek91 yeah good call. I'm really not apposed to it and having to add these disclaimers to PR's is enough of a sign that I should just support it. Updated.

pleek91 (approved):

PR #2043, opened 2024-01-30
Fix: Use anchor rather than start terminology for anchor date

### LINKED ISSUE #11551: Prefect UI has misleading messaging about `anchor_date` for `Interval` schedules
### First check

- [X] I searched the Prefect documentation for this issue.
- [X] I checked that this issue is related to Prefect and not one of its dependencies.

### Bug summary

Prefect's UI gives users the wrong impression about anchor dates for interval schedules.

1. The Prefect Cloud UI displays Interval schedules with an anchor date as "every <interval> from <anchor_date>" 
2. When editing an Interval schedule the `anchor_date` field is displayed as "Start date"

image

image

One place in the docs also gives a misleading description, see the table of attributes here https://docs.prefect.io/latest/concepts/schedules/?h=schedu#interval. Despite right below it clarifying the actual meaning
> Note that the anchor_date does not indicate a "start time" for the schedule, but rather a fixed point in time from which to compute intervals. If the anchor date is in the future, then schedule dates are computed by subtracting the interval from it. 

### Reproduction

```python3
Set an interval schedule using an anchor date in the Prefect Cloud UI
```

### Error

_No response_

### Versions

```Text
N/A
```

### Additional context

_No response_

zhen0:
Interval schedules have an anchor date, which is the interval that schedules can be calculated from.  As we set out [in the docs, ](https://docs.prefect.io/latest/concepts/schedules/?h=schedu#interval): 

> _the anchor_date does not indicate a "start time" for the schedule, but rather a fixed point in time from which to compute intervals. If the anchor date is in the future, then schedule dates are computed by subtracting the interval from it_

This PR changes some of the language in the UI to make it more clear. 

Closes https://github.com/PrefectHQ/prefect/issues/11551

### DIFF
--- a/src/models/IntervalSchedule.ts
+++ b/src/models/IntervalSchedule.ts
@@ -101,7 +101,7 @@ export class IntervalSchedule implements IIntervalSchedule {
     }
 
     if (this.anchorDate && verbose) {
-      str += ` from ${formatDate(this.anchorDate)} at ${formatTimeNumeric(this.anchorDate)} (${this.timezone ?? 'UTC'})`
+      str += ` using ${formatDate(this.anchorDate)} at ${formatTimeNumeric(this.anchorDate)} (${this.timezone ?? 'UTC'}) as the anchor date`
     }
 
     if (str == '') {
--- a/src/components/IntervalScheduleForm.vue
+++ b/src/components/IntervalScheduleForm.vue
@@ -17,7 +17,7 @@
       </div>
 
       <div>
-        <p-label label="Start date">
+        <p-label label="Anchor date">
           <DateInput v-model="anchorDate" show-time />
         </p-label>
       </div>


src/models/IntervalSchedule.ts
```diff
@@ -101,7 +101,7 @@ export class IntervalSchedule implements IIntervalSchedule {
     }
 
     if (this.anchorDate && verbose) {
-      str += ` from ${formatDate(this.anchorDate)} at ${formatTimeNumeric(this.anchorDate)} (${this.timezone ?? 'UTC'})`
+      str += ` using ${formatDate(this.anchorDate)} at ${formatTimeNumeric(this.anchorDate)} (${this.timezone ?? 'UTC'}) as the anchor date`
```
zhen0, line 104: We could also remove this section if we think the new wording is too verbose?

pleek91 (approved):
LGTM

PR #2044, opened 2024-01-30
Schemas: Add ui types and new mappers for schemas

pleek91:
# Description
Super simple mappers for schemas just to do case conversion and any other transformations we might want to do in the future.


src/schemas/mapper.ts
pleek91, line 1: To keep everything v2 self contained creating a new mapper instance just for anything schema's related. This will help us know what is deprecated at once v2 is release. But eventually this can be absorbed into the primary mapper.

znicholasbrown (approved):

PR #2045, opened 2024-01-30
Schemas: Support for arrays

pleek91:
# Description
Adds ux for properties of `type: 'array'`. Uses the new `PDraggableList` component from prefect-design. Can add, re-order, and delete items from an array. Supports all features of schema properties. 

Using this example flow
```python
class User(BaseModel):
    name: str
    age: int
    birthday: date

@flow
def schema_testing(list: List[User], names: List[str]):
```

Produces this UI
image
image

Currently the user is only given a json input (same flow)
image

### DIFF
--- a/src/schemas/components/SchemaFormProperties.vue
+++ b/src/schemas/components/SchemaFormProperties.vue
@@ -42,11 +42,11 @@
   const props = defineProps<{
     parent: SchemaProperty,
     properties: SchemaProperties,
-    values: SchemaValues | null,
+    values: SchemaValues | undefined,
   }>()
 
   const emit = defineEmits<{
-    'update:values': [SchemaValues | null],
+    'update:values': [SchemaValues | undefined],
   }>()
 
   const properties = computed(() => {
@@ -61,7 +61,7 @@
   })
 
   function getValue(propertyKey: string): unknown {
-    return props.values?.[propertyKey] ?? null
+    return props.values?.[propertyKey] ?? undefined
   }
 
   function setValue(propertyKey: string, value: unknown): void {
--- a/src/schemas/components/SchemaFormProperties.vue
+++ b/src/schemas/components/SchemaFormProperties.vue
@@ -1,16 +1,7 @@
 <template>
   <p-content class="schema-form-properties">
     <template v-for="[key, property] in properties" :key="key">
-      <template v-if="isPropertyWith(property, '$ref')">
-        <SchemaFormProperty
-          :value="getValue(key)"
-          :property="getSchemaDefinition(schema, property.$ref)"
-          :required="getRequired(key)"
-          @update:value="setValue(key, $event)"
-        />
-      </template>
-
-      <template v-else-if="isPropertyWith(property, 'allOf')">
+      <template v-if="isPropertyWith(property, 'allOf')">
         <SchemaFormPropertyAllOf
           :value="getValue(key)"
           :property="property"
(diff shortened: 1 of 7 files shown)


src/schemas/components/SchemaFormProperties.vue
```diff
@@ -1,16 +1,7 @@
...
   <p-content class="schema-form-properties">
     <template v-for="[key, property] in properties" :key="key">
-      <template v-if="isPropertyWith(property, '$ref')">
-        <SchemaFormProperty
-          :value="getValue(key)"
-          :property="getSchemaDefinition(schema, property.$ref)"
-          :required="getRequired(key)"
-          @update:value="setValue(key, $event)"
-        />
-      </template>
```
pleek91, line 11: This got moved into the SchemaFormPropertyInput component where it now does some more complete resolving of references.

src/schemas/components/SchemaFormProperties.vue
```diff
@@ -45,23 +36,19 @@
...
-  import { useSchema } from '@/schemas/compositions/useSchema'
   import { SchemaProperty, SchemaProperties, isPropertyWith } from '@/schemas/types/schema'
   import { SchemaValues } from '@/schemas/types/schemaValues'
-  import { getSchemaDefinition } from '@/schemas/utilities/definitions'
 
   const props = defineProps<{
     parent: SchemaProperty,
     properties: SchemaProperties,
-    values: SchemaValues,
+    values: SchemaValues | undefined,
```
pleek91, line 45: Values can always be undefined because a value might not exist yet on the form's v-model

src/schemas/components/SchemaFormPropertyAllOf.vue
```diff
@@ -33,11 +33,15 @@
...
+      if (isPropertyWith(definition, '$ref')) {
+        return merge(getSchemaDefinition(schema, definition.$ref), property)
+      }
 
-    return merge(property, definition)
-  }, {}))
+      return merge(property, definition)
+    }, {})
+
+    return merge(definitions, props.property)
```
pleek91, line 45: Noticed that the original title was being overridden by any title from the references. Merging the original property over the references makes sure the values on the property itself always win.

src/schemas/components/SchemaFormPropertyInput.vue
```diff
@@ -29,17 +33,27 @@
...
     emit('update:value', value)
   }
 
+  const property = computed(() => {
+    if (isPropertyWith(props.property, '$ref')) {
+      return merge({}, getSchemaDefinition(schema, props.property.$ref), props.property)
+    }
+
+    return props.property
+  })
```
pleek91, line 48: Here's where we're now making sure references are resolved. Used to happen in SchemaFormProperties.

src/schemas/components/SchemaFormPropertyInput.vue
```diff
@@ -67,14 +81,16 @@
     }
 
     if (isSchemaPropertyType(type, 'array')) {
-      // new list input from prefect-design
-      throw 'not implemented'
+      return withProps(SchemaFormPropertyArray, {
+        property: { ...property.value, type },
```
pleek91, line 85: This spread technically isn't necessary. But it helps typescript narrow the type for whole property.

src/schemas/components/SchemaFormPropertyObject.vue
pleek91, line 1: This is just a wrapper so we can have nested properties show up in a p-card.

znicholasbrown (approved):
Code for this looks fine, 2 design requests:
1. Can delete buttons use the small button type instead of regular? they're huge atm
2. Would like to see a better UX around empty lists besides just the add item button, it's a little confusing as-is. When a list is empty can we have some placeholder subdued (maybe even italicized?) text that says something like "{{property.title}} is empty"

pleek91:
@znicholasbrown 

> Can delete buttons use the small button type instead of regular? they're huge atm

I agree. Though that will turn into a menu so I'll play with the layout/sizing once the ux is more final

> Would like to see a better UX around empty lists besides just the add item button, it's a little confusing as-is. When a list is empty can we have some placeholder subdued (maybe even italicized?) text that says something like "{{property.title}} is empty"

@brandonreid I'd love to see some exploration around this. I thought about putting an empty item in the list but we've had some reports of schemas with circular references and automatically putting things in lists is one way that happens. But maybe won't be an issue?

PR #2046, opened 2024-01-31
Display `description` in resume modal when supplied

bunchesofdonald:
This updates the flow run resume modal to display a `description` if one has been supplied when the run was paused.

## Example

A flow run can pause and wait for input with a description:

```python
from prefect import flow, pause_flow_run
from prefect.input import RunInput


class TreeInput(RunInput):
    num_trees: int



@flow
async def plant_trees():
    tree_input = await pause_flow_run(
        wait_for_input=TreeInput.with_initial_data(
            description="How many **trees** to plant?"
        )
    )
    for i in range(tree_input.num_trees):
        print(f"Planting tree {i+1} of {tree_input.num_trees}...")


if __name__ == "__main__":
    plant_trees.serve(name="plant-trees")
```

Resuming this flow run in the modal will display the description:

Screenshot 2024-01-31 at 1 53 46 PM

### DIFF
--- a/src/services/WorkspaceFlowRunsApi.ts
+++ b/src/services/WorkspaceFlowRunsApi.ts
@@ -7,6 +7,7 @@ import { OrchestrationResultResponse } from '@/models/api/OrchestrationResultRes
 import { RunGraphDataResponse } from '@/models/api/RunGraphDataResponse'
 import { FlowRunsFilter, FlowRunsHistoryFilter } from '@/models/Filters'
 import { FlowRun } from '@/models/FlowRun'
+import { FlowRunInputKeyset } from '@/models/FlowRunInputKeyset'
 import { RunHistory } from '@/models/RunHistory'
 import { BatchProcessor } from '@/services/BatchProcessor'
 import { mapper } from '@/services/Mapper'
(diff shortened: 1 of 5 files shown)


src/services/WorkspaceFlowRunsApi.ts
```diff
@@ -73,12 +74,28 @@ export class WorkspaceFlowRunsApi extends WorkspaceApi {
...
 
+  public async getFlowRunInputSchema(id: string, keyset: FlowRunInputKeyset): Promise<Schema> {
+    const data = await this.getFlowRunInput(id, keyset.schema) as SchemaResponse | null
+    if (!data) {
+      throw new Error('Flow run input schema not found')
+    }
     return mapper.map('SchemaResponse', data, 'Schema')
   }
 
+
```
pleek91, line 98: So rather than calling the `getFlowRunInput` method and then casting the type I'd just hit the endpoint directly and say what you expect to get back like
```
public async getFlowRunInputDescription(id: string, keyset: FlowRunInputKeyset): Promise<string | null> {
  const { data } = await this.get<string | null>(`/${id}/input/${keyset.description}`)

  return data
}
```
Same for the input schema one. Otherwise we don't really get the type safety we'd like because even if the `getFlowRunInput` type's change these methods won't get type errors.

bunchesofdonald: That makes sense, thanks Craig!

pleek91 (approved):

PR #2047, opened 2024-01-31
Schemas: Field const support

pleek91:
# Description
Disables any property that has `const` because the default value must be used.


src/schemas/components/SchemaFormProperty.vue
```diff
@@ -2,17 +2,25 @@
   <p-label class="schema-form-property" :label="label">
     <template v-if="description" #description>
       <div class="schema-form-property__description">
-        <p-markdown-renderer :text="description" class="schema-form-input__markdown" />
+        <p-markdown-renderer :text="description" class="schema-form-property__markdown" />
       </div>
     </template>
 
-    <SchemaFormPropertyInput v-model:value="value" :property="property" />
+    <fieldset class="schema-form-property__fields" :disabled="disabled">
```
pleek91, line 9: Just wanna point out `fieldset` is my favorite html element

src/schemas/compositions/useSchemaProperty.ts
```diff
@@ -0,0 +1,21 @@
...
+
+    if (isPropertyWith(value, '$ref')) {
+      return merge({}, getSchemaDefinition(schema, value.$ref), value)
+    }
+
+    return value
+  })
+
+  return property
+}
```
znicholasbrown, line 21: ❤️

znicholasbrown (approved):

PR #2048, opened 2024-01-31
Consolidate status icons

collincchoy:
This PR contains some setup work for swapping work queue health -> new work queue status and some light TLC on "status icons". In it you can find the following:
- Rename `WorkPoolQueueStatusIcon` -> `WorkPoolQueueHealthIcon` - this is to avoid conflicting with future incoming component as both might be around at the same time for a little bit. Suppose we could consider this a breaking change but at least internally, we're not consuming that component downstream. 
- Consolidate the 🟢 / 🔴 styling from Deployment, Work Pool, Work Pool Queue, & Worker status icons into a single `StatusIcon` component - these styled elements will be swapped out for stronger iconography soon so consolidating the change radius for now
  * Note styles in this component are a bit trimmed down from their originating bespoke implementations. I'm attempting to trim out extraneous/conflicting styles but worth taking a closer look in case I missed something. will leave some line-by-line comments for posterity
- Fix spacing and vertical alignment of the `DeploymentStatusIcon` within the `DeploymentsList` component

  | **Before** | **After** |
  | ---- | ---- |
  | image | image |


src/components/WorkPoolStatusIcon.vue
```diff
@@ -39,23 +39,7 @@
 
 <style>
 .work-pool-status-icon { @apply
-  flex
-  items-center
```
collincchoy, line 43: content-less div that's inside of an inline element so this doesn't center the icon as-is

src/components/WorkPoolStatusIcon.vue
```diff
@@ -39,23 +39,7 @@
...
   cursor-help
-  w-2
-  h-2
-  align-middle
```
collincchoy, line 47: Does this even do anything in flex layout? regardless most consuming points that use these icons push their own layouts downward to do vertical alignment when/where necessary.

src/components/WorkPoolStatusIcon.vue
```diff
@@ -39,23 +39,7 @@
...
-  text-inverse
-  dark:text-default
```
collincchoy, line 49: no content inside the dot. copypasta from applying semantic tokens work

src/components/DeploymentsList.vue
```diff
@@ -26,11 +26,11 @@
...
       <template #deployment-name="{ row }">
         <div class="deployment-list__name-col">
-          <span>
-            <p-link :to="routes.deployment(row.id)" class="deployment-list__name">
+          <span class="deployment-list__name">
+            <p-link :to="routes.deployment(row.id)">
               {{ row.name }}
-              <DeploymentStatusIcon v-if="can.access.deploymentStatus" :status="row.status" />
             </p-link>
+            <DeploymentStatusIcon :status="row.status" />
```
collincchoy, line 33: Previously the icon was just inlined up against the name inside of a link. Now the icon is outside of the anchor tag (since it's a tooltip target also). It also looks like there was an intended `mr-2` to put 4px of space between, but since the icon is a _child_ of `<p-link>` there was just "implicit" inline spacing and now the flex items-center lifts the icon up just a tad to vertically center it.

src/components/StatusIcon.vue
```diff
@@ -0,0 +1,25 @@
+<template>
+  <div class="status-icon" :class="`status-icon--${status}`" />
+</template>
+
+<script setup lang="ts">
+  export type Status = 'ready' | 'not_ready'
```
collincchoy, line 6: I considered having this be more generic like a `variant='positive' | 'negative'` or so but after work queue health -> status, the only semantic mismatch is Worker which uses "online" vs "offline". everything else is using "ready" | "not_ready" semantics so figured this works well for most cases and just Worker can pivot (see changes to `WorkerStatusBadge`)

src/components/DeploymentsList.vue
```diff
@@ -239,8 +239,10 @@
 }
 
 .deployment-list__name { @apply
+  inline-flex
```
pleek91, line 242: idk if it wrapped well before but does making this flex cause any layout issues where the name won't wrap when it should?

pleek91 (approved):
Nice, thanks for cleaning that all up!

src/components/DeploymentsList.vue
pleek91: I don't feel strongly. Looks better inline vs flex I think but up to you. Just mostly wanted to make sure the name itself still wrapped.

collincchoy: Hmm took another stab and needs some more tinker time to get inlining space _just right_ but changes as-is should improve the majority of cases i think and i'll maybe revisit styling here later.

pleek91: Sounds good 👍 Ty!

PR #2049, opened 2024-02-01
Schemas: Default values, array improvements, and omit nullish values

pleek91:
# Description
Adds logic so default values from the schema get populated when the form is mounted. This happens at the component level so that way we don't have to recursively travers the schema and resolve every definition ahead of time. This also helps ensure that defaults at every level are respected.

### DIFF
--- a/src/schemas/components/SchemaFormProperties.vue
+++ b/src/schemas/components/SchemaFormProperties.vue
@@ -40,6 +40,7 @@
   import { SchemaProperty, SchemaProperties, isPropertyWith } from '@/schemas/types/schema'
   import { SchemaValues } from '@/schemas/types/schemaValues'
   import { SchemaValue } from '@/types'
+  import { isNullish } from '@/utilities'
 
   const props = defineProps<{
     parent: SchemaProperty,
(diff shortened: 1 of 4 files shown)


src/schemas/components/SchemaFormProperties.vue
```diff
@@ -65,13 +68,30 @@
...
+    flush()
   }
 
   function getRequired(propertyKey: string): boolean {
     return props.parent.required?.includes(propertyKey) ?? false
   }
+
+  const patches: { propertyKey: string, value: SchemaValue }[] = []
+
+  const flush = debounce(() => {
```
pleek91, line 82: This debounce helps collect multiple updates that happen at once and emits them together. Otherwise each emit overwrites the previous value because multiple emits can happen before the prop reflects the change.

src/schemas/components/SchemaFormProperties.vue
```diff
@@ -65,13 +68,30 @@
...
+    const updatedValues = { ...props.values }
+
+    patches.forEach(({ propertyKey, value }) => {
+      updatedValues[propertyKey] = value
+
+      if (isNullish(value)) {
+        delete updatedValues[propertyKey]
+      }
```
pleek91, line 90: This make sure that null/undefined values don't appear in the form values. That way we won't get errors like `'null' is not 'int'`.

src/schemas/components/SchemaFormPropertyArray.vue
```diff
@@ -39,6 +43,21 @@
       emit('update:value', value)
     },
   })
+
+  function getIndexProperty(index: number): SchemaProperty {
```
znicholasbrown, line 47: This function name could be a little clearer, maybe `getPropertyAtIndex`?

znicholasbrown (approved):
1 nit, otherwise LGTM

PR #2050, opened 2024-02-01
Trim down the resume form contents

abrookins:
Isolate the important information (user-provided description, if given) and remove redundant information, like the fact that the flow run is paused and that it requires you to fill out the form -- if the form is present, the message that you should fill it out is implicit, but required fields will alert you if you don't notice.

Also, the current API request to get description input is 404ing in cases where there is no description or the pause data is from an earlier version of Prefect. Adding a try/catch around that request fixed the issue.


pleek91 (commented):
If a flow is suspended I believe this same model is used. And I think the modal will just be empty with a submit button. Do you think maybe we should keep the "Do you want to resume" text (or something else) in case the flow doesn't have input?

abrookins:
Yes, that seems better @pleek91!

src/services/WorkspaceFlowRunsApi.ts
```diff
@@ -75,8 +75,12 @@ export class WorkspaceFlowRunsApi extends WorkspaceApi {
...
 
   public async getFlowRunInputDescription(id: string, keyset: FlowRunInputKeyset): Promise <string | null> {
-    const { data } = await this.get<string | null>(`/${id}/input/${keyset.description}`)
-    return data
+    try {
+      const { data } = await this.get<string | null>(`/${id}/input/${keyset.description}`)
+      return data
+    } catch (error) {
+      return null
+    }
```
pleek91, line 83: @abrookins Why the try catch here?

abrookins: Because otherwise the 404 exception crashes the modal.

abrookins: Ah, I should have mentioned. The form is currently broken without this -- I had to fix it to make my changes.

abrookins: Just updated the PR description to reflect this context

pleek91: oh I see because you're not using a subscription in the modal. I'd recommend not swallowing errors here and using a subscription in the modal.

```
const { response: inputDescription } = await useSubscription(api.flowRuns.getFlowRunInputDescription, props.flowRunId).promise()
```

Or at least `console.error(error)` in your catch so that any errors still show up in the devtools console.

pleek91 (approved):

PR #2052, opened 2024-02-01
Schemas: Add support for string formats

pleek91:
# Description
Adds support for date, datetime, and password string formats. 

image

### DIFF
--- a/src/schemas/components/SchemaFormPropertyDate.vue
+++ b/src/schemas/components/SchemaFormPropertyDate.vue
@@ -1,0 +1,26 @@
+<template>
+  <p-date-input v-model="value" />
+</template>
+
+<script lang="ts" setup>
+  import { computed } from 'vue'
+  import { SchemaProperty } from '@/schemas/types/schema'
+
+  const props = defineProps<{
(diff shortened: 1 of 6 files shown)


src/schemas/components/SchemaFormPropertyDate.vue
```diff
@@ -0,0 +1,44 @@
...
+  const dateFormat = 'yyyy-MM-dd'
+
+  const value = computed({
+    get() {
+      if (isNotNullish(props.value)) {
+        const parsed = parse(props.value, dateFormat, startOfToday())
+
+        if (isInvalidDate(parsed)) {
+          return null
+        }
```
znicholasbrown, line 28: For this and datetime properties - does a partial date throw null here? Thinking through the experience as someone types a date in manually instead of using the picker

znicholasbrown (approved):
One question, not really blocking

src/schemas/components/SchemaFormPropertyDate.vue
pleek91: The component doesn't emit until there is a valid date I believe. So this is only a safe guard for a value that already exists as a string that cannot be parsed.

PR #2056, opened 2024-02-02
Schemas: Add plumbing for `__prefect_kind` and implement `json`

pleek91:
# Description
Adds logic and components for switching between different types of inputs and adds the ability to switch between using a form or using json at any level. 

image

image


src/schemas/components/SchemaFormKindInput.vue
pleek91, line 1: Also including the switch from `null` to `"none"` for the noop kind.

src/schemas/components/SchemaFormPropertyAnyOf.vue
pleek91, line 1: Nothing changed functionally in this file. I was referencing it a few times and organized it a bit to make it more readable.

src/schemas/components/SchemaFormPropertyInput.vue
```diff
@@ -93,6 +93,8 @@
     }
 
     if (isSchemaPropertyType(type, undefined)) {
+      // automatically switch to prefect_kind === 'json'?
+      // value might not be json, so will need to convert it to json...
       throw 'not implemented'
```
znicholasbrown, line 98: In what scenarios would we need to convert the value? If it's coming from the db it should already be serialized so we shouldn't need do anything here I _think_

znicholasbrown (approved):
1 non-blocking comment, otherwise this looks like i'd expect

PR #2059, opened 2024-02-05
Schemas: List enhancements and support prefect_kind

pleek91:
# Description
Adds the ability to switch between prefect kinds for individual list items. Adds menu items for "Move to bottom" and "Move to top". Abstracts the logic for prefect kind as well as schema property input into compositions for reuse. 

image

### DIFF
--- a/src/schemas/components/SchemaFormProperty.vue
+++ b/src/schemas/components/SchemaFormProperty.vue
@@ -33,8 +33,8 @@
   import { computed } from 'vue'
   import SchemaFormPropertyMenu from '@/schemas/components/SchemaFormPropertyMenu.vue'
   import { usePrefectKind } from '@/schemas/compositions/usePrefectKind'
-  import { usePropertyInput } from '@/schemas/compositions/usePropertyInput'
   import { useSchemaProperty } from '@/schemas/compositions/useSchemaProperty'
+  import { useSchemaPropertyInput } from '@/schemas/compositions/useSchemaPropertyInput'
   import { SchemaProperty } from '@/schemas/types/schema'
   import { SchemaValue } from '@/schemas/types/schemaValues'
   import { isNullish } from '@/utilities'
@@ -73,7 +73,7 @@
   }
 
   const { kind } = usePrefectKind(value)
-  const { input } = usePropertyInput(property, value)
+  const { input } = useSchemaPropertyInput(property, value)
 </script>
 
 <style>
--- a/src/schemas/components/SchemaFormProperty.vue
+++ b/src/schemas/components/SchemaFormProperty.vue
@@ -31,13 +31,13 @@
 <script lang="ts" setup>
   import { isNotNullish } from '@prefecthq/prefect-design'
   import { computed } from 'vue'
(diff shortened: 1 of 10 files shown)


src/schemas/components/SchemaFormProperty.vue
pleek91, line 1: Nothing really changed here. Just abstracting pieces of this into compositions for reuse in the list components.

src/schemas/components/SchemaFormPropertyArray.vue
```diff
@@ -58,22 +70,29 @@
...
 }
 
-.schema-form-property-array__handle { @apply
-  cursor-grab
-  mt-2
+.schema-form-property-array__empty { @apply
+  text-subdued
+  text-sm
+  italic;
+  margin-left: 30px;
```
brandonreid, line 96: I'm slightly surprised this worked out to 30px and didn't land on the 8pt grid, but it's an eyeball number regardless.

brandonreid (approved):

src/schemas/compositions/usePrefectKind.ts
```diff
@@ -0,0 +1,34 @@
...
+    },
+    set(__prefect_kind) {
+      kindValuesMap[kind.value] = propertyValue.value
+
+      if (__prefect_kind in kindValuesMap) {
+        propertyValue.value = kindValuesMap[__prefect_kind]
+        return
+      }
+
+      propertyValue.value = { __prefect_kind }
```
znicholasbrown, line 27: Making sure I understand this correctly - this is what stores the value for a given kind view when switching between them, right?

src/schemas/compositions/useSchemaPropertyInput.ts
```diff
@@ -0,0 +1,49 @@
+import { MaybeRefOrGetter, Ref, computed, toValue } from 'vue'
+import SchemaFormPropertyInput from '@/schemas/components/SchemaFormPropertyInput.vue'
+import SchemaFormPropertyKindJson from '@/schemas/components/SchemaFormPropertyKindJson.vue'
+import { SchemaProperty } from '@/schemas/types/schema'
+import { isPrefectKindValue } from '@/schemas/types/schemaValues'
+import { SchemaValue } from '@/types'
+import { withProps } from '@/utilities/components'
+
+// this is a lot easier to just let typescript infer the return type
+// eslint-disable-next-line @typescript-eslint/explicit-function-return-type
```
znicholasbrown, line 10: 👍🏻

znicholasbrown (approved):

src/schemas/components/SchemaFormPropertyArray.vue
pleek91: This was, special lol. I wanna revisit this at the base component to clean this up. This this keeps me moving for now and it looks fine. Its the width of the drag icon itself plus 8px for gap I believe.

src/schemas/compositions/usePrefectKind.ts
pleek91: Correct, this composition handles caching values when switching and setting the correct value when the user changes the `kind`. Implementation components don't need to care about the value when switching kinds.

PR #2060, opened 2024-02-05
Schemas: Add ability to omit specific values

pleek91:
# Description
Adds a menu item to omit a specific value from a form. Omitting the value disabled the property and removes its value from the from values. Enabling the property adds the original value back to the form values. 

Note: You can only omit an optional value

image
image

### DIFF
--- a/src/schemas/components/SchemaFormProperties.vue
+++ b/src/schemas/components/SchemaFormProperties.vue
@@ -32,6 +32,7 @@
 </template>
 
 <script lang="ts" setup>
+  import { isDefined } from '@prefecthq/prefect-design'
   import debounce from 'lodash.debounce'
   import { computed } from 'vue'
   import SchemaFormProperty from '@/schemas/components/SchemaFormProperty.vue'
@@ -40,7 +41,6 @@
   import { SchemaProperty, SchemaProperties, isPropertyWith } from '@/schemas/types/schema'
   import { SchemaValues } from '@/schemas/types/schemaValues'
   import { SchemaValue } from '@/types'
-  import { isNullish } from '@/utilities'
 
   const props = defineProps<{
     parent: SchemaProperty,
@@ -85,7 +85,7 @@
     patches.forEach(({ propertyKey, value }) => {
       updatedValues[propertyKey] = value
 
-      if (isNullish(value)) {
+      if (!isDefined(value)) {
         delete updatedValues[propertyKey]
       }
     })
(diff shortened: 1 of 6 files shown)


src/schemas/components/SchemaFormProperties.vue
pleek91, line 1: There are a few updates around `null` vs `undefined`. `null` is 1-1 with python's `None` vs `undefined` is not a value and is removed from the form values. `undefined` values are removed as part of `JSON.stringify` anyway but also specifically removing them from the form value so keep that object consistent with what gets sent to the api.

src/schemas/components/SchemaFormProperty.vue
```diff
@@ -64,7 +88,7 @@
       return null
     },
     set(value) {
-      emit('update:value', value)
+      emit('update:value', value ?? undefined)
```
pleek91, line 91: Our core prefect-design components generally emit `null` for empty values. So making sure we coerce `null` values to `undefined` here.

src/schemas/components/SchemaFormPropertyMenu.vue
```diff
@@ -15,6 +17,7 @@
 
   const props = defineProps<{
     kind: PrefectKind,
+    disabled?: boolean,
```
pleek91, line 20: When a value is omitted you cannot change the prefect_kind. So adding the disabled prop hides all the kind options and just leaves the slot. Not super clean but works for now. Might revisit this if more things like this pop up.

znicholasbrown: (non-blocking) - what does that look like? Just the label/description of the prop?

znicholasbrown (approved):

src/schemas/components/SchemaFormPropertyMenu.vue
znicholasbrown: scratch that I understand now - thought this was in the property component and not the menu

znicholasbrown:
Not sure how I feel about the strikeout property names - they're a little hard to read and not entirely clear that it's related to the omitting. Would prefer subdued/disabled text with `(Omitted)` or something clearer but can handle in a follow up PR

pleek91:
> Not sure how I feel about the strikeout property names - they're a little hard to read and not entirely clear that it's related to the omitting. Would prefer subdued/disabled text with `(Omitted)` or something clearer but can handle in a follow up PR

Agreed I think it could be improved. But we can tweak that at anytime so I figure stuff like that will get cleaned up toward the end when we're tweaking after trying everything together.

PR #2065, opened 2024-02-06
Schemas: Add support for using workspace variables as a property value

pleek91:
# Description
Adds support the `workspace_variable` prefect kind on the schema form. 

image

image


znicholasbrown (approved):
LGTM

PR #2067, opened 2024-02-06
Schemas: Specify supported `kind` values per implementation

pleek91:
# Description
Adds a prop for specifying which prefect kind features are supported. This will allow each implementation to pick and choose features. For example `jinja` is currently only supported for automations.


src/schemas/components/SchemaFormPropertyMenu.vue
```diff
@@ -8,13 +8,15 @@
     </template>
 
     <template v-if="$slots.default">
```
znicholasbrown, line 10: If you're already pulling in the slots composition, might as well use that here for access consistency

znicholasbrown (approved):

PR #2069, opened 2024-02-07
Schema: Validation errors

pleek91:
# Description
Implements the new workspace level api endpoint `/ui/schemas/validate` to validate user entered values based on the schema provided. 

The example in this screenshot is validating individual items of a list. Which is something we've never been able to do. Also notice the specific errors. 
image

@znicholasbrown this uses prop drilling. Which isn't great. But considering we're already using v-models up and down adding props for validation seems like the best and simplest solution.

### DIFF
--- a/src/schemas/components/SchemaForm.vue
+++ b/src/schemas/components/SchemaForm.vue
@@ -1,7 +1,7 @@
 <template>
   <p-form class="schema-form" @submit="submit">
     <template v-if="schema.properties">
-      <SchemaFormProperties v-model:values="values" :parent="schema" :properties="schema.properties" />
+      <SchemaFormProperties v-model:values="values" :parent="schema" :properties="schema.properties" :errors="errors" />
     </template>
   </p-form>
 </template>
@@ -14,7 +14,7 @@
   import { schemaFormKindsInjectionKey } from '@/schemas/compositions/useSchemaFormKinds'
   import { Schema } from '@/schemas/types/schema'
   import { PrefectKind, SchemaValues } from '@/schemas/types/schemaValues'
-  import { SchemaValuesValidationError } from '@/schemas/types/schemaValuesValidationResponse'
+  import { SchemaValueError } from '@/schemas/types/schemaValuesValidationResponse'
 
   const props = withDefaults(defineProps<{
     schema: Schema,
@@ -35,7 +35,7 @@
   }>()
 
   const api = useWorkspaceApi()
-  const errors = ref<SchemaValuesValidationError[]>()
+  const errors = ref<SchemaValueError[]>([])
 
   const values = computed({
     get() {
--- a/src/schemas/components/SchemaForm.vue
+++ b/src/schemas/components/SchemaForm.vue
@@ -1,5 +1,5 @@
 <template>
-  <p-form class="schema-form">
+  <p-form class="schema-form" @submit="submit">
     <template v-if="schema.properties">
       <SchemaFormProperties v-model:values="values" :parent="schema" :properties="schema.properties" />
     </template>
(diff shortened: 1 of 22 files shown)


src/schemas/components/SchemaForm.vue
pleek91, line 1: This form does not implement any actions like a "save" button. It leaves that up to the implementation. Which is why `loading` is an optional v-model so that at implementation it can be bound to whatever button/action is used.

src/schemas/components/SchemaForm.vue
```diff
@@ -1,32 +1,42 @@
...
 
-  const props = defineProps<{
+  const props = withDefaults(defineProps<{
     schema: Schema,
     values: SchemaValues,
     kinds: PrefectKind[],
-  }>()
+    loading?: boolean | null,
+  }>(), {
+    loading: null,
```
pleek91, line 25: without this `loading` is automatically defined as `false` but I won't this to be truly optional.

src/schemas/components/SchemaFormPropertyArray.vue
```diff
@@ -5,11 +5,12 @@
         No items in this list
       </p>
     </template>
-    <p-draggable-list v-model="value" allow-create allow-delete :generator="generator">
+    <p-draggable-list v-model="value" allow-create allow-delete :generator="generator" :state="state">
```
pleek91, line 8: draggable list doesn't accept `State` yet but this will future proof for that.

src/schemas/maps/schema.ts
pleek91, line 1: Need to convert `Schema` back to `SchemaResponse` for the validation endpoint.

src/schemas/types/schemaResponse.ts
```diff
@@ -23,7 +23,7 @@ export type SchemaPropertyResponse = {
   // exclusiveMaximum?: boolean | number,
   // exclusiveMinimum?: boolean | number,
   format?: SchemaStringFormat,
-  items?: SchemaPropertyResponse,
+  items?: SchemaPropertyResponse | SchemaPropertyResponse[],
```
pleek91, line 26: This type was incorrect and attempting to reverse the mapping revealed that.

src/schemas/mapper.ts
znicholasbrown, line 1: I'm guessing adding the new mappers was causing a conflict with the existing service?

src/schemas/components/SchemaFormPropertyArray.vue
znicholasbrown: Ah I wasn't really thinking of these as inputs since they don't extend the base input class but i see the use case - let's chat at some point to see what that should look like

src/schemas/components/SchemaForm.vue
```diff
@@ -35,4 +45,35 @@
...
+    },
+  })
+
+  async function submit(): Promise<void> {
+    loading.value = true
+
+    try {
+      const { valid, errors: errorsResponse } = await api.schemas.validate(props.schema, values.value)
+
+      errors.value = errorsResponse
```
znicholasbrown, line 66: Just double checking cause I didn't look at the API for this - `/validate` doesn't throw an error if the schema is invalid, right? It returns a 200 + some errors?

src/schemas/components/SchemaForm.vue
znicholasbrown: How do implementers hit `submit`? I think this will obfuscate the footer slot and doesn't expose a default slot of its own

pleek91: `<p-button type="submit" :form="formId" />`

This is something we've been doing lately because it works nice with modals where the modal has its own footer. Might need to add a footer to this form as well. But for now it doesn't seem needed. Easy enough to add if it is needed.

src/schemas/components/SchemaFormPropertyArray.vue
pleek91: might not even be necessary. We'll figure it out if we need.

src/schemas/mapper.ts
pleek91: The existing mapper already has mappers by the same names. So seemed simpler to just create a new mapper.

src/schemas/components/SchemaForm.vue
pleek91: I believe it returns a 403 if the schema itself is invalid.

src/schemas/mapper.ts
znicholasbrown: yeah ok that's what i thought

src/schemas/components/SchemaForm.vue
znicholasbrown: Ahh ok i knew that but forgot we were using that pattern now, thank you

znicholasbrown (approved):

PR #2070, opened 2024-02-08
Disambiguate work queues across work pools in `<WorkQueueCombobox>`

collincchoy:
> [!Warning]
> This PR includes a breaking change to the `<WorkQueueCombobox` component as its component interface now expects to work directly with work queue ids instead of names for it's v-model.

### The problem:
The `<WorkQueueCombobox>` renders a picklist of work queues keyed on and labeled with work queue _name_ only; however **work queue names are only unique within a work pool.**

So if you have more than 1 hybrid (a type that uses queues) work pool in your workspace, then you likely already have 2 work queues with the name "default" belonging to 2 different work pools. So then when using the `<WorkQueueCombobox>`, you end up with duplicate options. 

WorkQueueCombobox seems old. I don't see it being used much so I'm thinking of repurposing it a bit in the following ways:
1. Group the dropdown options by work pool - now 2 work queues named "default" are distinguishable
2. add client-side filtering by 0 or 1+ work pool ids
3. deal with work queue ids instead of names (both as input and output)

#### 👀 Show me
| **Description** | **📸** |
| ---- | ---- |
| Status quo - flat list of all work queues in current workspace with buggy experience for overlapping work queue names | image |
| Work queues options are now grouped by work pool. Picking "default" no longer selects all options labeled as "default". | image |
| Using the workPoolFilter prop, the options can be filtered (client-side) by work pool. When the workPoolFilter prop changes and it would invalidate any of the already-selected options, then those options are automatically removed. | image |
| The component now works with `id`s directly instead of names in order to uniquely identify work queues _outside_ of their respective work pool | image |

### DIFF
--- a/src/components/WorkQueueCombobox.vue
+++ b/src/components/WorkQueueCombobox.vue
@@ -13,17 +13,18 @@
 </template>
 
 <script lang="ts" setup>
-  import { PCombobox, SelectOptionGroup } from '@prefecthq/prefect-design'
+  import { PCombobox, SelectOptionGroup, isSelectOptionNormalized } from '@prefecthq/prefect-design'
   import { useSubscription } from '@prefecthq/vue-compositions'
-  import { computed } from 'vue'
+  import { computed, watch } from 'vue'
   import { useWorkspaceApi } from '@/compositions'
-  import { WorkQueue } from '@/models'
+  import { WorkPoolFilter, WorkQueue } from '@/models'
 
   const props = defineProps<{
     selected: string | string[] | null | undefined,
     emptyMessage?: string,
     allowUnset?: boolean,
     multiple?: boolean,
+    workPoolFilter?: WorkPoolFilter,
   }>()
 
   const emits = defineEmits<{
@@ -52,6 +53,13 @@
   const workQueues = computed(() => workQueuesSubscription.response ?? [])
   const options = computed<SelectOptionGroup[]>(() => {
     const workQueuesGroupedByWorkPool = workQueues.value.reduce<Map<string, WorkQueue[]>>((acc, workQueue) => {
+      // Filter options by work pool id or name
+      if (props.workPoolFilter?.id || props.workPoolFilter?.name) {
+        if (!props.workPoolFilter.id?.includes(workQueue.workPoolId) && !(workQueue.workPoolName && props.workPoolFilter.name?.includes(workQueue.workPoolName))) {
+          return acc
+        }
+      }
+
       const workPoolName = workQueue.workPoolName ?? 'No work pool'
       acc.set(workPoolName, (acc.get(workPoolName) ?? []).concat(workQueue))
       return acc
@@ -76,5 +84,35 @@
     }
 
     return options
+  })
+
+  // Remove previously selected work queues that are no longer options
+  // when the options change
+  watch(options, (newValue, oldValue) => {
+    const newWorkQueueIds = new Set(newValue.flatMap(group => group.options.map((option) => {
+      if (isSelectOptionNormalized(option)) {
+        return option.value
+      }
+      return null
+    })))
+
+    const removedWorkQueueIds = new Set()
+    for (const group of oldValue) {
+      for (const option of group.options) {
+        if (isSelectOptionNormalized(option) && !newWorkQueueIds.has(option.value)) {
+          removedWorkQueueIds.add(option.value)
+        }
+      }
+    }
+
+    if (removedWorkQueueIds.size === 0) {
+      return
+    }
+
+    if (!multiple.value && internalValue.value && removedWorkQueueIds.has(internalValue.value)) {
+      internalValue.value = undefined
+    } else if (Array.isArray(internalValue.value) && internalValue.value.length) {
+      internalValue.value = internalValue.value.filter(id => !removedWorkQueueIds.has(id))
+    }
   })
 </script>
--- a/src/components/WorkQueueCombobox.vue
+++ b/src/components/WorkQueueCombobox.vue
@@ -13,10 +13,11 @@
 </template>
 
 <script lang="ts" setup>
-  import { PCombobox, SelectOption } from '@prefecthq/prefect-design'
+  import { PCombobox, SelectOptionGroup } from '@prefecthq/prefect-design'
   import { useSubscription } from '@prefecthq/vue-compositions'
   import { computed } from 'vue'
   import { useWorkspaceApi } from '@/compositions'
+  import { WorkQueue } from '@/models'
 
   const props = defineProps<{
     selected: string | string[] | null | undefined,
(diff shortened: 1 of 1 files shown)


src/components/WorkQueueCombobox.vue
```diff
@@ -49,20 +51,68 @@
...
-      label: workQueue.name,
-    }))
+  const options = computed<SelectOptionGroup[]>(() => {
+    const workQueuesGroupedByWorkPool = workQueues.value.reduce<Map<string, WorkQueue[]>>((acc, workQueue) => {
+      // Filter options by work pool id or name
+      if (props.workPoolFilter?.id || props.workPoolFilter?.name) {
+        if (!props.workPoolFilter.id?.includes(workQueue.workPoolId) && !(workQueue.workPoolName && props.workPoolFilter.name?.includes(workQueue.workPoolName))) {
+          return acc
+        }
+      }
```
pleek91, line 61: Couldn't we just pass `props.workPoolFilter` into the `getWorkQueues`? Then this options computed would just group and create options for whatever work queues are returned from the api.

src/components/WorkQueueCombobox.vue
```diff
@@ -49,20 +51,68 @@
...
+        label: '',
       })
     }
 
     return options
   })
+
+  // Remove previously selected work queues that are no longer options
+  // when the options change
+  watch(options, (newValue, oldValue) => {
```
pleek91, line 91: Can you expound on why this is necessary? We don't have anything like this in our other combobox components.

src/components/WorkQueueCombobox.vue
collincchoy: That api doesn't support filtering by work pools? 

so I'm filtering client-side

pleek91: hmm we should probably update the api rather than filtering client side. As is it looks like this would support anything in the WorkPoolFilter (like `type`). 

If updating the api isn't an option maybe just add props for `workPoolName` or `workPoolId` so that way its clear what the component supports filtering by.

src/components/WorkQueueCombobox.vue
collincchoy: image

In that example ^, 
1) I pick some work queues that are across work pools
2) I decide no only queues from the "collins little minions" work pool
3)
    a) Without this, the combobox filters down the selectable options - one of which was previously selected and now it shows up as an empty tag without any content
    image
b) With this, the behavior is (imo) more consistent. if the value contains an option that's not a valid option after the options have changed, it's removed.

pleek91: hmm right because this combobox can filter by work pool which is kinda special compared to our other comboboxes. I'd probably just add an `@update:modelValue` to the work pool combobox and clear out the work queues when it changes. But I see the use case. Thanks for the explanation.

src/components/WorkQueueCombobox.vue
collincchoy: Hm. how about just warning if type is provided? 

server-side updates aside, I think I'd rather have this singular object interface for filtering the workpool over more discrete props and then I grabbed the WorkPoolFilter b/c it's semantically correct and the reuse seemed nice for familiarity rather than creating some other structure for like { id?: string[], name?: string[] }... 🤔 but maybe I just do that and lean into structural typing... like a consumer could still pass a WorkPoolFilter 🤷

pleek91: I don't feel strongly other than making it type safe. If we ever add something to WorkPoolFilter and then this component doesn't work as expected because it accepts a WorkPoolFilter but doesn't implement the whole thing is  a land mind for a bug IMO. I'd prefer to update the api to allow filtering by workPool as that's the safest way to ensure things will continue to work properly. But I know you're down a rabbit hole already. So just making this component type safe would be my bar.

collincchoy: Hm yeah +1 more narrowly accurate structural types > nominal typing.

I'm gonna switch it to its own little id/name type. 

it probably is a little off b/c the filters are generally like "api filters" even if they're not labeled as such... so I see where this works out better if the filtering was server-side and this prop is just going straight into api filters. 👍

pleek91: yeah pretty much every (maybe all) "ThingFilter" is a type for the api and we don't usually use the term "filter" for client side stuff. But it be fine as long as the client side enforced the whole filter was implemented or could accurately error for things that were not implemented. But easier and safer to just narrow the type to exactly what you support. 

Thanks for talking that through!

src/components/WorkQueueCombobox.vue
pleek91: yeah, I see your point 👍

src/components/WorkQueueCombobox.vue
```diff
@@ -49,20 +51,68 @@
   const api = useWorkspaceApi()
   const workQueuesSubscription = useSubscription(api.workQueues.getWorkQueues, [{}])
```
zhen0, line 52: @desertaxle - Collin is updating an older component to use with work queue related automations. If we continue to use the [workspace/get_work_queues](https://api.stg.prefect.dev/api/accounts/{account_id}/workspaces/{workspace_id}/work_queues/filter) route, will this conflict with future plans to remove api work queue routes after agents are deprecated?

src/components/WorkQueueCombobox.vue
zhen0: Can you not use the getWorkPoolQueues route to filter by work pool?

collincchoy: The work pool queue apis don't support "breaking out" of a single work pool.

This - without like a "get all work pools -> for each work pool: get all work queues"
image

src/components/WorkQueueCombobox.vue
collincchoy: to be fair, these changes don't really lean _more_ into that route. Another way to look at this is I'm fixing a bug where these components are currently being used? 

For automations, we have another component that wraps this one which I'm going to be using so _transitively_ it's another use of this route but still bounded/contained in the same way as the status quo.

src/components/WorkQueueCombobox.vue
collincchoy: > I'm gonna switch it to its own little id/name type.

addressed in 00b3d75

src/components/WorkQueueCombobox.vue
zhen0: Yep that's fair.  I'm just keen for us to make sure that we are coordinating a few different pieces of work.  Worst case scenario is that we have to update this component again in the future so my comment is non-blocking.

pleek91 (approved):

src/components/WorkQueueCombobox.vue
collincchoy: 👍 in slack

PR #2074, opened 2024-02-09
Render the work pool queue status icon in deployment page header 

collincchoy:
If a deployment has a work pool and work queue, then we currently render the work queue's name on the deployment's page under its name in the header. However, that work queue's _status_ is not currently showing because the boolean check relies on there being a flow run state (with the intention of hiding the status for flow runs that have already terminated since the current status of the terminated flow run's work queue is no longer relevant). 

For a deployment, it could be useful to show the work queue's status here (if there is one to show) so this PR tweaks the condition to make that happen.

| **Before** | **After** |
| ---- | ---- |
| image | image |

### DIFF
--- a/src/components/FlowRunWorkQueue.vue
+++ b/src/components/FlowRunWorkQueue.vue
@@ -3,7 +3,7 @@
     <span>Work Queue</span>
     <WorkQueueIconText :work-queue-name="workQueueName" :work-pool-name="workPoolName" />
 
-    <template v-if="isNotTerminal && workPoolName">
+    <template v-if="showWorkPoolQueueStatus && workPoolName">
       <WorkPoolQueueStatusIcon v-if="can.access.workQueueStatus && workPoolQueue" :work-pool-queue="workPoolQueue" />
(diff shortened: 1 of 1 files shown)


pleek91 (approved):
Makes sense to me

src/components/FlowRunWorkQueue.vue
```diff
@@ -26,7 +26,7 @@
 
   const can = useCan()
 
-  const isNotTerminal = computed(() => props.flowRunState && !isTerminalStateType(props.flowRunState))
+  const showWorkPoolQueueStatus = computed(() => props.flowRunState == undefined || !isTerminalStateType(props.flowRunState))
```
collincchoy, line 29: think i might flip this so it's less confusing. 

Does
```
<template v-if="!hideWorkPoolQueueStatus && workPoolName">
  ...
</template>
...

const hideWorkPoolQueueStatus = computed(() => props.flowRunState && isTerminalStateType(props.flowRunState))
```

make more sense?

PR #2075, opened 2024-02-09
Schemas: Display errors for json values

pleek91:
# Description
Displays errors for json values in a formatted list

image

### DIFF
--- a/src/schemas/components/SchemaFormPropertyErrors.vue
+++ b/src/schemas/components/SchemaFormPropertyErrors.vue
@@ -25,7 +25,7 @@
 <script lang="ts" setup>
   import { SchemaValueError, isSchemaValuePropertyError, isSchemaValueIndexError } from '@/schemas/types/schemaValuesValidationResponse'
   import { isString } from '@/utilities'
-  
+
   defineProps<{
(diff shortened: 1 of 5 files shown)


src/schemas/components/SchemaFormPropertyErrors.vue
```diff
@@ -0,0 +1,38 @@
...
+        </li>
+      </template>
+    </template>
+  </ul>
+</template>
+
+<script lang="ts" setup>
+  import { SchemaValueError, isSchemaValuePropertyError, isSchemaValueIndexError } from '@/schemas/types/schemaValuesValidationResponse'
+  import { isString } from '@/utilities'
+  defineProps<{
```
pleek91, line 28:
```suggestion
  import { isString } from '@/utilities'
  
  defineProps<{
```

src/schemas/components/SchemaFormPropertyMenu.vue
pleek91, line 1: Unrelated, but making sure that the form can always be visible

brandonreid (approved):
Looks like there's a lint error to hit up there, otherwise g2g and super dope

PR #2076, opened 2024-02-09
Schema demo tweaks

pleek91:
# Description
Just a few layout tweaks from things I noticed while doing a demo for the engineering update that I don't wanna forget over the weekend.


collincchoy (approved):
makes sense to me

PR #2080, opened 2024-02-12
Old work queue pages use work queue status

collincchoy:
The old work queue (pre- workers/work pools) pages are still using work queue health instead of the new status. This PR updates so that they use the new status if the flag is on. 

| **Before** | **After** |
| ---- | ---- |
| image | image |
| image | image |


zhen0 (approved):
Thanks @collincchoy

src/components/WorkQueueStatusBadge.vue
```diff
@@ -1,28 +1,36 @@
...
+  const can = useCan()
+
   const workQueueName = computed(() => props.workQueue.name)
   const workQueueId = computed(() => props.workQueue.id)
   const { workQueueStatus } = useWorkQueueStatus(workQueueId)
   const healthy = computed(() => workQueueStatus.value?.healthy)
 
   const label = computed(() => {
+    if (can.access.workQueueStatus) {
+      return { 'ready': 'Ready', 'not_ready': 'Not Ready', 'paused': 'Paused' }[props.workQueue.status]
```
znicholasbrown, line 31: This is weird but I'm guessing it's in lieu of more formal mapping for an otherwise deprecated set of components?

znicholasbrown (approved):

src/components/WorkQueueStatusBadge.vue
collincchoy: yea here's to hoping this file gets deleted before anyone looks at it again 🙈

PR #2088, opened 2024-02-14
Updates for adding status to push pools and their work queues

zhen0:
Previously work pool status and work queue status only referred to work pools which are polled by workers.  We have now introduced the idea of status for push style work pools and their work queues.  This PR updates the wording in the tooltips to remove mention of workers for push pools. 

This PR also updates the workPoolQueueStatusArray component to show work queue status for push pool related work queues on the dashboard. 

After:

image

image

Before:

image


src/components/WorkPoolQueueStatusIcon.vue
```diff
@@ -17,12 +17,15 @@
...
   import { computed } from 'vue'
   import StatusIcon from '@/components/StatusIcon.vue'
+  import { useWorkPool } from '@/compositions/useWorkPool'
   import { WorkPoolQueue } from '@/models'
 
   const props = defineProps<{
     workPoolQueue: WorkPoolQueue,
   }>()
 
+  const { workPool } = useWorkPool(props.workPoolQueue.workPoolName)
```
pleek91, line 27: Assuming this composition accepts a getter you'll want to use one to not break reactivity if the prop changes. 
```suggestion
  const { workPool } = useWorkPool(() => props.workPoolQueue.workPoolName)
```

src/components/WorkPoolStatusIcon.vue
```diff
@@ -15,17 +15,22 @@
...
   import StatusIcon from '@/components/StatusIcon.vue'
-  import { WorkPoolStatus } from '@/models'
+  import { WorkPool } from '@/models'
 
   const props = defineProps<{
-    status: WorkPoolStatus,
+    workPool: WorkPool,
   }>()
 
+  const status = ref(props.workPool.status)
```
pleek91, line 26: You'll want to make this a computed otherwise you'll break reactivity. (Remove the `ref` import as well as it will be unused)
```suggestion
  const status = computed(() => props.workPool.status)
```

pleek91 (changes requested):

zhen0:
Thanks @pleek91 - changes made

pleek91 (approved):

PR #2090, opened 2024-02-14
BugFix: Deployment interval schedules not being created or displayed accurately in the ui

### LINKED ISSUE #11924: Deployment's Schedule incorrectly shows the timezone as "UTC"
### First check

### Bug summary

When creating an interval schedule in the UI for a Deployment and using `UTC` time it'll correctly display the time in the UI once finished if your settings use `UTC` time. But when switching your preferences to display times in the `America/Chicago` timezone the time displayed will updated to reflect the timezone but still show `UTC` instead of `America/Chicago`.

### Reproduction

With your Preferences set to show times in "UTC", add a schedule on a deployment. Use the timezone option `UTC` when creating this schedule.

Then change your preferences to another timezone (America/Chicago). When viewing the schedule in the UI it'll show an updated time reflecting your preferred timezone but still has the label `UTC`.

### Error

I added a schedule like this. 
AddSchedule

It shows up like this when I have the `UTC` preference selected in Prefect's UI.
ViewScheduleWithUTCPreference

When I switch my timezone preference to Chicago time it shows up like this, with the time changed but the `UTC` label remaining.
ViewScheduleWithChicagoPreference

### Browsers

- [X] Chrome
- [X] Firefox
- [ ] Safari
- [ ] Edge

### Prefect version

```Text
Using Prefect Cloud.
```

### Additional context

_No response_

pleek91:
# Description
Interval schedules that use a different timezone than the devices local time could produce weird schedules because the selected timezone was not factored into the date the schedule was being created with. This fixed both the creation and display of the anchor date so that it works the way users expect. 

Closes https://github.com/PrefectHQ/prefect/issues/11924

### DIFF
--- a/src/components/IntervalScheduleForm.vue
+++ b/src/components/IntervalScheduleForm.vue
@@ -18,7 +18,7 @@
 
       <div>
         <p-label label="Anchor date">
-          <DateInput v-model="anchorDate" show-time />
+          <p-date-input v-model="anchorDate" show-time />
         </p-label>
       </div>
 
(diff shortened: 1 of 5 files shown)


znicholasbrown (approved):
Walked through this together so I'm confident this works as expected - thanks @pleek91; we haven't tested this against yaml or python-created schedules so that'd be the next step in testing.

zhen0:
Checked against some anchor date scenarios such as every 2 hours in NYC and Chicago time - LGTM!  Thanks @pleek91

PR #2098, opened 2024-02-16
Schemas: Add v2 schema and properties to the deployment model

pleek91:
# Description
Updates the `Deployment` model and associated types and mappers to include v2 properties. The existing parameter and schema properties are left as is for backwards compatibility.

### DIFF
--- a/src/maps/deployment.ts
+++ b/src/maps/deployment.ts
@@ -1,5 +1,5 @@
 import { sortStringArray } from '@prefecthq/prefect-design'
-import { DeploymentFlowRunCreate, DeploymentFlowRunRequest, DeploymentUpdate, DeploymentUpdateRequest } from '@/models'
+import { DeploymentFlowRunCreate, DeploymentFlowRunRequest, DeploymentUpdate, DeploymentUpdateRequest, SchemaResponse } from '@/models'
 import { DeploymentResponse } from '@/models/api/DeploymentResponse'
 import { Deployment } from '@/models/Deployment'
 import { createObjectLevelCan } from '@/models/ObjectLevelCan'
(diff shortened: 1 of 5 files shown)


brandonreid (approved):
Knowing you coded this, I won't ask why the `as`s here, haha. lgtm

src/maps/deployment.ts
```diff
@@ -3,11 +3,13 @@ import { DeploymentFlowRunCreate, DeploymentFlowRunRequest, DeploymentUpdate, De
...
 import { Deployment } from '@/models/Deployment'
 import { createObjectLevelCan } from '@/models/ObjectLevelCan'
+import { SchemaResponseV2, schemaV2Mapper } from '@/schemas'
 import { MapFunction } from '@/services/Mapper'
+import { Schema } from '@/types/schemas'
 
 export const mapDeploymentResponseToDeployment: MapFunction<DeploymentResponse, Deployment> = function(source) {
   const rawSchema = source.parameter_openapi_schema ?? {}
-  const schema = this.map('SchemaResponse', rawSchema, 'Schema')
+  const schema = this.map('SchemaResponse', rawSchema as Schema, 'Schema')
```
znicholasbrown, line 12: Why is this being cast here?

znicholasbrown (dismissed):

znicholasbrown:
ah that wasn't meant to be an approval

src/maps/deployment.ts
pleek91: I made the response type a union of the old response type and the new response type. Quick hack so that when I'm back cleaning up this model I can just remove the old type and the casting. I thought about writing type guards for that seemed overly complex for this. Open to other ideas though. This is admittedly a little hacky. But since we're just telling typescript what the response type is I don't think it matters much if we continue just telling it here.

pleek91: Fixed the type being used here. TY!

znicholasbrown (approved):

PR #2101, opened 2024-02-16
Schemas: Don't allow switching to a form input if the property has no type

pleek91:
# Description
Removes the "Use form input" if the property has no `type`.

### DIFF
--- a/src/schemas/components/SchemaFormPropertyArrayItem.vue
+++ b/src/schemas/components/SchemaFormPropertyArrayItem.vue
@@ -6,7 +6,7 @@
       <component :is="input.component" v-bind="input.props" />
     </keep-alive>
 
-    <SchemaFormPropertyMenu v-model:kind="kind">
+    <SchemaFormPropertyMenu v-model:kind="kind" :property="property">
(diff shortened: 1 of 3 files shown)


src/schemas/components/SchemaFormPropertyArrayItem.vue
```diff
@@ -6,7 +6,7 @@
       <component :is="input.component" v-bind="input.props" />
     </keep-alive>
 
-    <SchemaFormPropertyMenu v-model:kind="kind">
+    <SchemaFormPropertyMenu v-model:kind="kind" :property="property">
```
collincchoy, line 9: i've been wondering when/if we can use the bind shorthand now... slash what thing we need to flip to be able to

collincchoy (approved):

src/schemas/components/SchemaFormPropertyArrayItem.vue
pleek91: oh yeah! That will be nice. I wonder if volar has caught up 🤔

PR #2102, opened 2024-02-16
HTTP encode API calls to work_pools endpoints

### LINKED ISSUE #11768: Escape work pool names when translating from CLI/UI to API URLs
We had an instance of a customer with a work pool named `' some-word\'` which couldn't be removed via the UI or the API due to the leading space and the trailing `\`.  Work pools are a little different from many of our domain objects, as they are referenced in URLs by their names, and that's the only way to `DELETE` them at the moment.

When the CLI or UI translate work pool names into API URLs, they should each be using HTTP URL encoding, so that example above would become `'%20some-word%5C'` when used in URLs.

hamzamogni:
fixes PrefectHQ/prefect#11768

Currently, work pool names are being fetched from the API using their names, which poses problems when the name contains special characters like `\`, spaces...etc.

This PR fixes the UI part of the issue referenced above, it encodes URLs related to work pool API calls before sending requests.

The built-in `encodeURI` function is used to do that.


znicholasbrown (changes requested):
Hi @hamzamogni - thanks for the contribution! To keep these methods a little cleaner can you please encode just the name argument for each instead of the entire path string? Those shouldn't need to be encoded. Example:

```ts
public async getWorkers(workPoolName: string, filter: WorkPoolWorkersFilter = {}): Promise<WorkPoolWorker[]> {
  const encodedWorkPoolName = encodeURI(workPoolName)
  const { data } = await this.post<WorkPoolWorkerResponse[]>(`/${encodedWorkPoolName}/workers/filter`, request)

  return mapper.map('WorkPoolWorkerResponse', data, 'WorkPoolWorker')
}

```

hamzamogni:
Thank you! I have updated the code accordingly

znicholasbrown (approved):
Thanks so much @hamzamogni !

PR #2113, opened 2024-02-20
Overridable single batch processing deployments n blocks

collincchoy:
This PR aims to enable fixing https://github.com/PrefectHQ/nebula-ui/issues/4196 by allowing the deployment and block documents apis' batch processor callback for processing a GET for a single object to be overridden in derived classes. 

For more on what/why this is useful, see a downstream implementation in - https://github.com/PrefectHQ/nebula-ui/pull/4220


pleek91 (approved):
Add nice find! Implementation makes sense

PR #2114, opened 2024-02-20
Schemas: Fix enum values not being selectable

pleek91:
# Description
Fixes a missing emit that was causing values to not be able to be selected

### DIFF
--- a/src/schemas/components/SchemaFormPropertyString.vue
+++ b/src/schemas/components/SchemaFormPropertyString.vue
@@ -51,7 +51,7 @@
         'onUpdate:modelValue': update,
       })
     }
-    
+
     if (format === 'json-string') {
       return withProps(PCodeInput, {
         lang: 'json',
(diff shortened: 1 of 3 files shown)


src/schemas/components/SchemaFormPropertyString.vue
```diff
@@ -60,13 +60,13 @@
...
     }
-    
+
     if (stringEnum) {
       return withProps(PCombobox, {
         modelValue: props.value,
         state: props.state,
         options: stringEnum.filter(isString),
-        'onUpdate:modelValue': (value) => asType(value, String),
+        'onUpdate:modelValue': (value) => update(asType(value, String)),
```
znicholasbrown, line 69: Enums can have types that aren't strings - is this still correct?

pleek91: hmm this is specifically for properties of type `string`. So I think this is correct? But that probably means I need to account for enums for other things. Numbers and integers probably? What else could an enum be? A boolean?

pleek91: Add added support for integer enums in https://github.com/PrefectHQ/prefect-ui-library/pull/2114/commits/3eeca3f158e701b1607211a14f68743d3269bf05

znicholasbrown: Hm ok I think we might need to expand support there but this is a strict improvement so going to approve

znicholasbrown (approved):

src/schemas/components/SchemaFormPropertyString.vue
pleek91: Love to work with you on that if you have some time. Let me know!

PR #2119, opened 2024-02-20
Schemas: Add support for `None` (`null`) values and improve `anyOf` property layout

pleek91:
# Description
This PR focuses on two related concepts. 

1. Use `undefined` specifically for an "empty" value. Previously both `null` and `undefined` were used interchangeable for this. This allows `null` to be used as an actual value

2. Add a component for properties with a type of `null` which sets the value to `null` when selected. This allows specifically using a `null` (`None`) value. 

## Example
For this flow
```
class User(BaseModel):
  ...

class Param(BaseModel):
  foo: Union[User, int, None] = Field(title="Foo", description="Some description")

@flow
def schema_testing(test: Param):
  pass
```

The "Foo" parameter now has this layout (there should be a card box around the user properties, somethings not working correctly on my local but should be fine, it not I'll address)
image

Which used to look like this
image

### DIFF
--- a/src/schemas/components/SchemaFormPropertyAnyOf.vue
+++ b/src/schemas/components/SchemaFormPropertyAnyOf.vue
@@ -1,15 +1,14 @@
 <template>
-  <div class="schema-form-property-any-of-input">
-    <p-button-group v-model="selectedPropertyIndex" :options="options" small />
-  </div>
-
-  <p-card>
-    <SchemaFormProperty :key="selectedPropertyIndexValue" v-model:value="value" :property="property" :required="required" :errors="errors" />
-  </p-card>
+  <keep-alive>
+    <SchemaFormProperty :key="selectedPropertyIndexValue" v-model:value="value" v-bind="{ property, required, errors }" class="schema-form-property-any-of-input">
+      <p-button-group v-model="selectedPropertyIndex" :options="options" small class="mb-2" />
+    </SchemaFormProperty>
+  </keep-alive>
 </template>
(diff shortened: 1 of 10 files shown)


src/schemas/components/SchemaFormPropertyAnyOf.vue
```diff
@@ -73,10 +72,10 @@
     const selectedProperty = props.property.anyOf[selectedPropertyIndex.value]
 
     if (isPropertyWith(selectedProperty, '$ref')) {
-      return getSchemaDefinition(schema, selectedProperty.$ref)
+      return merge({}, getSchemaDefinition(schema, selectedProperty.$ref), props.property)
     }
 
-    return selectedProperty
+    return merge({}, selectedProperty, props.property)
```
znicholasbrown, line 78: These merges might be backwards - I think property definitions from a selected property should overwrite those of the base property, so props like title and description don't get displayed incorrectly

znicholasbrown (approved):
1 comment otherwise LGTM

src/schemas/components/SchemaFormPropertyAnyOf.vue
pleek91: So what's happening here is for the property label and description we want to always show the value from the property itself. Otherwise in a union of classes you'll never see the property's name. You'll just see "User", when the property itself might be called "Owner". 

Then you see the different available types in the button group. So you can click "User" or possibly something like "None". 

image

PR #2127, opened 2024-02-21
Schemas: wrap property key titles in the `titleCase` utility

pleek91:
# Description
When a property doesn't have a title we use the property's key as the title. Now capitalizing it to match what pydantic does by default when it does create titles.

### DIFF
--- a/src/schemas/components/SchemaFormProperties.vue
+++ b/src/schemas/components/SchemaFormProperties.vue
@@ -46,6 +46,7 @@
   import { SchemaValueError } from '@/schemas/types/schemaValuesValidationResponse'
   import { getSchemaPropertyErrors } from '@/schemas/utilities/errors'
   import { SchemaValue } from '@/types'
+  import { titleCase } from '@/utilities/strings'
 
   const props = defineProps<{
     parent: SchemaProperty,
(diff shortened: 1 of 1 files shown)


collincchoy (approved):
almost want to suggest sentence casing (like `Snake case`) as a more ideal label

At least `titleCase` swaps underscores for spaces.. - not sure if you see the properties camelcased or snakecased tho. 

at the very least i imagine this is better than status quo. 

edit: ah I see a slack thread on this now... - also I guess if the actual titles you see (when they're available) are like `Snake Case` (idk what to call this - space-separated PascalCase..?) then I suppose this should be a consistent fallback for those snake_cased properties you might be seeing without titles.

pleek91:
> almost want to suggest sentence casing (like `Snake case`) as a more ideal label
> 
> At least `titleCase` swaps underscores for spaces.. - not sure if you see the properties camelcased or snakecased tho.
> 
> at the very least i imagine this is better than status quo.
> 
> edit: ah I see a slack thread on this now... - also I guess if the actual titles you see (when they're available) are like `Snake Case` (idk what to call this - space-separated PascalCase..?) then I suppose this should be a consistent fallback for those snake_cased properties you might be seeing without titles.

Yeah I agree with everything you said. Just trying to be consistent with the other titles 👍

PR #2128, opened 2024-02-22
Schemas: Improve list inputs for tuples

pleek91:
# Description
Fixes bug when deleting items from a list. Also prevents adding more values to a list when it has `maxItems`. 

image


src/schemas/components/SchemaFormPropertyArray.vue
```diff
@@ -21,6 +21,11 @@
         />
       </template>
     </p-draggable-list>
+    <template v-if="!allowCreate && property.maxItems">
```
collincchoy, line 24: nit: probably nbd but if for some reason `property.maxItems === 0`, you wouldn't be able to add anything and you wouldn't ever see the message

src/schemas/components/SchemaFormPropertyArray.vue
```diff
@@ -100,4 +112,10 @@
   italic;
   margin-left: 30px;
 }
+
+.schema-form-property-array__max { @apply
+  text-subdued
+  text-sm;
+  margin-left: 30px;
```
collincchoy, line 119: nit: css variable for this magic #?

collincchoy (approved):
LGTM

> Fixes bug when deleting items from a list.
What was the bug/fix for deletion? jc

src/schemas/components/SchemaFormPropertyArray.vue
pleek91: Good suggestion!

pleek91:
> LGTM
> 
> > Fixes bug when deleting items from a list.
> > What was the bug/fix for deletion? jc

Sorry, terrible PR description on my part. When you deleted an item the updated list was doing weird stuff because of the `keep-alive`. Things would be out of place and sometimes the list length wouldn't change.

PR #2131, opened 2024-02-22
Schemas polish

pleek91:
# Description
See comments on changes for details

### DIFF
--- a/src/schemas/components/SchemaFormPropertyKindJson.vue
+++ b/src/schemas/components/SchemaFormPropertyKindJson.vue
@@ -28,6 +28,14 @@
       return props.value.value
     },
     set(value) {
+      if (value?.length === 0) {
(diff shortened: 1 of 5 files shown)


src/schemas/components/SchemaFormPropertyKindJson.vue
pleek91, line 1: Making sure empty json inputs don't send an empty string as the parameter. This will mean users will get "property is required" errors rather than "json is invalid" errors. And allow empty json inputs to pass validation when the property is optional.

src/schemas/components/SchemaFormPropertyNull.vue
pleek91, line 1: Just making sure something is shown to users when the property is type none. Something similar happens in v1.

znicholasbrown (approved):

Issue #1977, opened 2023-12-29
Ensure all content on flowrun dashboard card wraps


WillRaphaelson:
![image](https://github.com/PrefectHQ/prefect-ui-library/assets/26799928/7faa6c51-66ce-4721-8d26-d63a7020451a)

PR #2133, opened 2024-02-23
Design: Update StateListItem.vue

aaazzam:
In #1977 @WillRaphaelson reports that StateListItem cards do not properly wrap. 

It appears some fix has been done to address this, but still possess a small paper cut:
Screenshot 2024-02-23 at 4 38 14 PM

Our use of tag groups almost always collapses tags to a "+N" indicator in the presence of a large title. The collapsed tag group can't be interacted with, and so in the presence of a large title we basically present a count of tags, which is pretty useless IMO. 

This PR simply hides overflow from a shrunken tag container on the state list item. 
Screenshot 2024-02-23 at 4 37 56 PM

This will be revisited in a future PR to prefect-design, to make tag-group "+N" indicators interactable.

### DIFF
--- a/src/components/StateListItem.vue
+++ b/src/components/StateListItem.vue
@@ -178,6 +178,7 @@
 
 .state-list-item__tags { @apply
   min-w-0
+  overflow-hidden
   grow;
 
   grid-area: tags;


brandonreid:
Thanks for flagging this and throwing down a quick fix! I riffed on this PR to instead allow text to wrap: https://github.com/PrefectHQ/prefect-ui-library/pull/2136

PR #2136, opened 2024-02-26
prevent overflowing tags

brandonreid:
This is a riff on @aaazzam 's [PR](https://github.com/PrefectHQ/prefect-ui-library/pull/2133/files). Here instead of allowing tags to overflow, which can sort of half way happen at times and prevents the user from at least hovering the tags to see what they are (not an amazing experience, granted). This removes some unneeded styles which allows long run names to wrap. The ellipsis styles that were there weren't working anyway, so now the full run name can be visible too.

Screenshot 2024-02-26 at 7 59 26 AM

Thanks for attacking these UI issues @aaazzam

### DIFF
--- a/src/components/StateListItem.vue
+++ b/src/components/StateListItem.vue
@@ -134,7 +134,7 @@
 .state-list-item__top-section { @apply
   grid
   w-full
-  items-center
+  items-start
   gap-2;
 
   grid-template-columns: 1fr auto;
--- a/src/components/StateListItem.vue
+++ b/src/components/StateListItem.vue
@@ -153,12 +153,7 @@
 
 .state-list-item__name { @apply
   text-base
-  text-subdued
-  shrink-0
-  whitespace-nowrap
-  grow-0
-  text-ellipsis
-  overflow-hidden;
+  text-subdued;
 
   grid-area: name;
 }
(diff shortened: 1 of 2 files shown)


brandonreid:
@aaazzam I actually forgot to push some code here, so it's actually just removing a lot of borked styles, same outcome though. I'll hit up that tooltip real quick, good lookin out.

brandonreid:
For posterity. Tooltip PR in prefect-design: https://github.com/PrefectHQ/prefect-design/pull/1124

src/components/StateListItem.vue
```diff
@@ -153,12 +153,7 @@
 
 .state-list-item__name { @apply
   text-base
-  text-subdued
-  shrink-0
-  whitespace-nowrap
-  grow-0
-  text-ellipsis
-  overflow-hidden;
+  text-subdued;
```
znicholasbrown, line 156: If we're allowing wrapping here we should align anything else on the line to the top of the bounding box; as it is right now, tags will align to the middle of the content instead, which breaks scanning expectations (see your screenshot for a minor example of this)

brandonreid: Good call. Here's an updated screenshot:

image

brandonreid: Alignment when it's not wrapping:

image

znicholasbrown (approved):

PR #2143, opened 2024-02-29
Feature: Flow run graph with artifacts

brandonreid:
Neccesary updates to support the latest version of the graphs repo with Artifacts selection.

https://github.com/PrefectHQ/prefect-ui-library/assets/6776415/927069fb-115d-43d4-90e0-e6526118e546

Note: Nebula and OSS changes will be required. I'll make sure those PR's are fully prepped for this one before merging.


src/components/FlowRunGraphPopover.vue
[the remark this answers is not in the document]
pleek91: `p-popover` allows you to pass in whatever position you would like. So if you wanted to use `p-popover` hmu and I can show you how.

src/components/FlowRunGraphPopover.vue
```diff
@@ -0,0 +1,64 @@
...
+    setTimeout(() => {
+      document.addEventListener('click', eventHandler)
+      document.addEventListener('focusin', eventHandler)
+    }, 0)
+  })
+
+  onUnmounted(() => {
+    document.removeEventListener('click', eventHandler)
+    document.removeEventListener('focusin', eventHandler)
+  })
```
pleek91, line 38: You could use [`useGlobalEventListener`](https://github.com/prefecthq/vue-compositions/tree/main/src/useGlobalEventListener) for this so you don't have to manager the lifecycle

src/components/FlowRunGraphPopover.vue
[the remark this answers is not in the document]
brandonreid: the-benchwarmers-dang-it

src/components/FlowRunGraphSelectionPanel.vue
```diff
@@ -1,31 +1,36 @@
...
 
-  defineProps<{
-    node: NodeSelection | null,
+  const props = defineProps<{
+    selection: GraphItemSelection | null,
   }>()
 
+  const isSelectionTaskOrFlow = computed(() => {
+    return props.selection && isNodeSelection(props.selection)
+  })
```
pleek91, line 26: Seems like this condition should probably be external to this component. Probably don't even need to mount this component unless there is a selection and its is type `NodeSelection`.

src/components/FlowRunGraphArtifactsPopover.vue
```diff
@@ -0,0 +1,63 @@
...
+    'update:selection': [null],
+  }>()
+
+  const artifactsSelection = computed(() => {
+    if (props.selection && isArtifactsSelection(props.selection)) {
+      return props.selection
+    }
+
+    return null
+  })
```
pleek91, line 38: Same thought here about this component only accepting an ArtifactsSelection rather than a GraphItemSelection. 

You might want a wrapper component that accepts a selection | null and then mounts the appropriate popover component depending on what type of selection there is.

src/components/FlowRunGraphArtifactDrawer.vue
```diff
@@ -0,0 +1,113 @@
+<template>
+  <p-drawer
+    class="flow-run-graph-artifact-drawer"
+    :open="artifactId !== null"
+    placement="right"
+    @update:open="close"
+  >
+    <template v-if="artifactId && artifact">
```
pleek91, line 8: nit: be a little simpler to just check for `artifact` and then use `artifact.id` rather than making this block dependent on two variables.

src/components/FlowRunGraphArtifactDrawer.vue
```diff
@@ -0,0 +1,113 @@
...
+      <div class="flow-run-graph-artifact-drawer__header">
+        <div class="flow-run-graph-artifact-drawer__details">
+          <div v-if="artifact.key">
+            <span class="flow-run-graph-artifact-drawer__detail-title">Key</span>
+            <ArtifactKeyIconText :artifact-id="artifactId" />
+          </div>
```
pleek91, line 14: nit: This would make a nice component and that would make this component much easier to grok/read

src/components/FlowRunGraphArtifactCard.vue
```diff
@@ -0,0 +1,25 @@
...
+  import { useArtifact, useWorkspaceRoutes } from '@/compositions'
+
+  const props = defineProps<{
+    artifactId: string,
+  }>()
+
+  const routes = useWorkspaceRoutes()
+
+  const { artifactId } = toRefs(props)
+  const { artifact } = useArtifact(artifactId)
```
pleek91, line 24: nit
```suggestion
  const { artifact } = useArtifact(() => props.artifactId)
```

src/components/FlowRunGraphPopover.vue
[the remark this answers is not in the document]
brandonreid: After some experimentation and back and forth in slack, `p-popover` has some accommodations specific to trigger+popover combos. For now we'll roll with the custom one here, down the line we might hoist some version of this up to `prefect-design` if we see additional needs for non trigger binded popovers.

src/components/FlowRunGraphPopover.vue
brandonreid: I do need my `setTimeout` so it doesn't immediately call `eventHandler`. Curious your take on this, do I...
A. leave it as is.
B. Update useGlobalEventListener to have an option for like `immediate = false`
C. I could manage a flag that's set after the tick and checked in the eventHandler function

src/components/FlowRunGraphPopover.vue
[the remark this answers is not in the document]
brandonreid: This has been updated to use `p-pop-over` after all, just to leverage the screen edge detection and all that. We had to add an invisible dom node to append the pop over too and it was worth it.

pleek91 (approved):

PR #2144, opened 2024-02-29
Schemas: Updated deployment form component

pleek91:
# Description
Adds a `DeploymentFormV2` component that uses the new schema components. This will be implemented in cloud and oss based on the schemas feature flag.


src/components/DeploymentFormV2.vue
pleek91, line 1: This is basically the same as DeploymentForm minus vee-validate and has updated schema components.

src/maps/deployment.ts
```diff
@@ -8,9 +9,10 @@ import { MapFunction } from '@/services/Mapper'
 import { Schema } from '@/types/schemas'
 
 export const mapDeploymentResponseToDeployment: MapFunction<DeploymentResponse, Deployment> = function(source) {
+  const parameters = readonly(source.parameters)
```
pleek91, line 12: I was noticing that the parameters were getting updated somehow. Like `parametersV2` which should be exactly what is returned from the api was being modified by something. By wrapping the parameters in `readonly` that fixed the issue. I was hoping I'd see an error pop up to find out what was modifying them. But no errors. Its strange. But this fixed it and didn't want to waste any more time tracking down what old code was accidentally updating the parameters by an object reference.

src/schemas/components/SchemaFormPropertyInput.vue
```diff
@@ -114,6 +114,7 @@
         property: { ...property.value, type },
         value: value,
         state: props.state,
+        'onUpdate:value': (value) => emit('update:value', value),
```
pleek91, line 117: fixes a bug where the value wasn't being emitted

src/schemas/components/SchemaFormPropertyMenu.vue
```diff
@@ -44,6 +45,10 @@
       return false
     }
 
+    if (isSchemaPropertyType(props.property.type, undefined) && isDefined(props.property.enum)) {
+      return true
+    }
+
```
pleek91, line 51: Fixes an issue where you couldn't switch back to the enum form.

znicholasbrown (approved):
Just 1 comment - I don't think you need to resolve in this PR if I'm correct about the behavior but a user should be able to update the deployment without providing all parameters; those are only required at runtime.

src/components/DeploymentFormV2.vue
```diff
@@ -0,0 +1,165 @@
...
+    (event: 'submit', value: DeploymentUpdateV2): void,
+    (event: 'cancel'): void,
+  }>()
+
+  async function submit(): Promise<void> {
+    if (shouldValidate.value) {
+      try {
+        const valid = (await Promise.all([
+          validate(),
+          validateParameters(),
```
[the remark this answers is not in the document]
pleek91: I don't think so, because on line 100 we're modifying the schema to remove all required properties. Making everything optional. And then we're sending that modified schema to the validation endpoint.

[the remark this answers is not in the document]
znicholasbrown: i was looking for that code and still skipped over it 😮‍💨

PR #2148, opened 2024-03-01
Schemas: Add the ability to use json as entire parameters value

### LINKED ISSUE #12068: Regression in UI: Missing JSON Input Option for Flow Parameters in New Release
### First check

### Bug summary

The latest update to the Prefect UI has removed the ability to input flow parameters as JSON directly, a feature critical to our deployment pipelines. Previously, our workflows utilized the JSON input method extensively for specifying various parameters, but this option is no longer available in the updated UI.

### Reproduction

1.     Navigate to the "Deployments" section for a specific flow.
2.     Attempt to run a flow with parameters that were typically inputted via JSON in the "Parameters" section.
3.     Observe that the option to input parameters as JSON is missing, replaced by individual fields without the ability to add complex JSON objects directly.

### Error

-     The main issue is the absence of the JSON input option in the "Parameters" section of the flow run configuration.
-     No direct console errors related to this issue, but the functionality gap prevents the execution of flows that depend on JSON parameter input.

### Browsers

- [X] Chrome
- [X] Firefox
- [ ] Safari
- [ ] Edge

### Prefect version

_No response_

### Additional context

_No response_

pleek91:
# Description
Adds the ability to use a single json field for all parameters when creating a flow run
image

Closes https://github.com/PrefectHQ/prefect/issues/12068

### DIFF
--- a/src/components/DeploymentFormV2.vue
+++ b/src/components/DeploymentFormV2.vue
@@ -33,9 +33,17 @@
 
     <p-divider />
 
-    <h3 class="deployment-form__heading">
-      {{ localization.info.parameters }}
-    </h3>
+
+    <div class="flow-run-create-form-v2__header">
+      <h3 class="deployment-form__heading">
+        {{ localization.info.parameters }}
+      </h3>
+      <p-icon-button-menu small>
+        <p-overflow-menu-item v-if="kind === 'json'" label="Use form input" @click="kind = 'none'" />
+        <p-overflow-menu-item v-if="kind === 'none'" label="Use JSON input" @click="kind = 'json'" />
+      </p-icon-button-menu>
+    </div>
+
 
     <SchemaInputV2 v-model:values="parameters" :schema="schema" :errors="errors" :kinds="['none', 'json']" />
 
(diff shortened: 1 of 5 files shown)


src/components/DeploymentFormV2.vue
```diff
@@ -96,6 +105,7 @@
   const infrastructureOverrides = ref(stringify(props.deployment.infrastructureOverrides))
   const enforceParameterSchema = ref(props.deployment.enforceParameterSchema)
   const shouldValidateParameters = ref(true)
+  const { kind } = usePrefectKind(parameters)
```
collincchoy, line 108: not new in this PR but huh i was surprised this is intended to be mutated from outside 🤨

collincchoy (approved):

znicholasbrown (approved):

src/components/DeploymentFormV2.vue
pleek91: What do you mean? The `kind` or the `parameters`?

collincchoy: the `kind`. was just surprised seeing that get mutated directly up in the template.

pleek91: gotcha, yeah we don't do it often but its not an uncommon pattern to have a composition return a ref that you can mutate and then based on that mutation the composition does something. In this case its tracking some in memory values for parameters so that when you switch between kinds it remembers what you had before.

PR #2153, opened 2024-03-04
Schemas: Fix properties that default to json missing their default values

pleek91:
# Description
See code comments for explanation of fixes

### DIFF
--- a/src/components/FlowRunCreateFormV2.vue
+++ b/src/components/FlowRunCreateFormV2.vue
@@ -74,6 +74,7 @@
 <script lang="ts" setup>
   import { showToast } from '@prefecthq/prefect-design'
   import { useValidation, useValidationObserver } from '@prefecthq/vue-compositions'
+  import merge from 'lodash.merge'
   import { computed, h, ref } from 'vue'
   import FlowRunCreateFormTags from '@/components/FlowRunCreateFormTags.vue'
   import FlowRunCreateFormWhen from '@/components/FlowRunCreateFormWhen.vue'
@@ -87,7 +88,7 @@
   import { usePrefectKind } from '@/schemas/compositions/usePrefectKind'
   import { useSchemaValidation } from '@/schemas/compositions/useSchemaValidation'
   import { SchemaValues } from '@/schemas/types/schemaValues'
-  import { isEmptyObject, merge } from '@/utilities/object'
+  import { isEmptyObject } from '@/utilities/object'
   import { isRequired } from '@/utilities/validation'
 
   const props = defineProps<{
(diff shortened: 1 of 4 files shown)


src/components/FlowRunCreateFormV2.vue
pleek91, line 1: Was accidentally using the wrong merge function. Probably need to evaluate where we're using this an see if we can remove it to avoid this in the future.

src/schemas/components/SchemaFormPropertyUnknown.vue
```diff
@@ -36,6 +36,17 @@
...
+  if (!isDefined(props.property.enum)) {
+    const valueOrDefaultValue = isDefined(props.value) ? props.value : props.property.default
+
+    const json: PrefectKindJson = {
+      __prefect_kind: 'json',
+      value: asJson(valueOrDefaultValue),
+    }
+
+    emit('update:value', json)
+  }
```
pleek91, line 48: This was happening inside the computed. Which was bad for two reaons
1. Side effects in computed properties 🤮 
2. Caused the timing to be off and would end up getting reset to `null` by a higher component

src/utilities/types.ts
```diff
@@ -11,7 +11,7 @@ export function asType<T extends() => unknown>(value: unknown, type: T): ReturnT
 }
 
 export function asJson(value: unknown): string | undefined {
-  if (isNullish(value)) {
+  if (!isDefined(value)) {
```
pleek91, line 14: `null` is a valid json value so making sure we don't remove it.

src/components/FlowRunCreateFormV2.vue
znicholasbrown: Oof i didn't realize those were different, good catch

znicholasbrown (approved):
approving to keep the work moving but 1 comment that might be blocking

src/schemas/components/SchemaFormPropertyUnknown.vue
```diff
@@ -36,6 +36,17 @@
...
 
+  if (!isDefined(props.property.enum)) {
+    const valueOrDefaultValue = isDefined(props.value) ? props.value : props.property.default
+
+    const json: PrefectKindJson = {
+      __prefect_kind: 'json',
+      value: asJson(valueOrDefaultValue),
+    }
+
+    emit('update:value', json)
```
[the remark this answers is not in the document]
pleek91: Emitting this basically updates the ui to show a json input rather than this unknown component. This is the same as we were doing before just doing it in setup rather than in the computed.

PR #2162, opened 2024-03-05
Enhancement: add flow run graph states layer

brandonreid:
Adds the root flow run graph states layer. This is feature flagged from the API level. The flag is on for staging and off for prod. State events are opt in so the graph will be fine.

What isn't feature flagged is the empty state. So this will go out, I don't think it's worth feature flagging. If no state is added, it will appear like this on a graph with no nodes:
image

Also updates the selection popover to emit when the popover has been dismissed. This way when dismissing the popover by clicking somewhere outside of the graph, the selection is still cleared. One UX nit I'm still facing is when selecting another of the same type (e.g. a state bar is selected and I immediately click another state bar), it clears the selection rather than selecting the next thing I clicked on. This is still an improvement for now and I'll work out how to check for that down the line.

https://github.com/PrefectHQ/prefect-ui-library/assets/6776415/edb3dd5b-66d7-41f9-b466-7ba528a745a6

### DIFF
--- a/src/maps/runGraphData.ts
+++ b/src/maps/runGraphData.ts
@@ -1,5 +1,5 @@
-import { RunGraphData, RunGraphNode, RunGraphArtifact } from '@prefecthq/graphs'
-import { RunGraphDataResponse, RunGraphNodeResponse, RunGraphArtifactResponse } from '@/models/api/RunGraphDataResponse'
+import { RunGraphData, RunGraphNode, RunGraphArtifact, RunGraphStateEvent, StateType } from '@prefecthq/graphs'
+import { RunGraphDataResponse, RunGraphNodeResponse, RunGraphArtifactResponse, RunGraphStateResponse } from '@/models/api/RunGraphDataResponse'
 import { MapFunction } from '@/services/Mapper'
 import { isKnownArtifactType } from '@/types/artifact'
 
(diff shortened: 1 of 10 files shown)


src/maps/runGraphData.ts
```diff
@@ -31,6 +31,15 @@ export const mapRunGraphArtifactResponse: MapFunction<RunGraphArtifactResponse,
   }
 }
 
+export const mapRunGraphStateResponse: MapFunction<RunGraphStateResponse, RunGraphStateEvent> = function(source) {
+  return {
+    id: source.id,
+    timestamp: this.map('string', source.timestamp, 'Date'),
+    type: source.type as StateType,
```
pleek91, line 38: Seems like if this is true the source type should just use this type. Otherwise maybe a mapper is needed?

src/maps/runGraphData.ts
```diff
@@ -41,11 +50,16 @@ export const mapRunGraphDataResponse: MapFunction<RunGraphDataResponse, RunGraph
     return this.map('RunGraphArtifactResponse', artifact, 'RunGraphArtifact')
   }) ?? []
 
+  const states: RunGraphStateEvent[] = source.states?.map(state => {
+    return this.map('RunGraphStateResponse', state, 'RunGraphStateEvent')
+  }) ?? []
```
pleek91, line 55: Can't you just pass the array of states into the mapper and it'll return an array or undefined?

pleek91 (approved):
Two non blocking comments.

PR #2167, opened 2024-03-05
[Flow run infra overrides] Add an input to override job variables during flow run creation (via run deployment)

collincchoy:
This PR adds a new (behind a feature flag) JSON input to both versions of the `FlowRunCreateForm` so that when configuring a custom run of a deployment, a user can pass through JSON to override job variables from both the deployment and its work pool. 

> [!Note]
> The work to make the submitted value actually override and get picked up at runtime is ongoing. This PR is low-effort, enable setting it through the UI (atm this form submission propagates through to the DB). I will be adding this input in 2 other forms where deployment flow runs can be configured in follow-ups. 

New input outlined in pink below:

image


src/components/JobVariableOverridesInput.vue
```diff
@@ -0,0 +1,24 @@
...
+<script setup lang="ts">
+  import JsonInput from '@/components/JsonInput.vue'
+
+  const modelValue = defineModel<string>()
+</script>
+
+<style>
+.job-variable-overrides-input { @apply
+  min-h-[2.5rem]
+  min-w-full
```
collincchoy, line 20: I ported these styles over from the `<DeploymentForm />` tho I can't... for the life of me figure out what these min-h and min-w were for? 

if you know - i'd love to know 🙃

src/components/JobVariableOverridesInput.vue
```diff
@@ -0,0 +1,24 @@
...
+    :min-lines="3"
+    class="job-variable-overrides-input"
+    show-format-button
+  />
+</template>
+
+<script setup lang="ts">
+  import JsonInput from '@/components/JsonInput.vue'
+
+  const modelValue = defineModel<string>()
```
collincchoy, line 14: 🤺

src/components/DeploymentForm.vue
```diff
@@ -56,13 +56,7 @@
         {{ localization.info.infraOverrides }}
       </h3>
       <p-label label="Infrastructure Overrides (Optional)" :message="overrideErrorMessage" :state="overrideState">
```
collincchoy, line 58: I'm gonna probably change the label here to Job Variables soon.

src/components/FlowRunCreateFormV2.vue
```diff
@@ -158,6 +168,7 @@
       },
       name: name.value,
       parameters: parameters.value,
+      jobVariables: jobVariables.value ? JSON.parse(jobVariables.value) : undefined,
```
pleek91, line 171: Should this just be part of the mapper?

src/components/JobVariableOverridesInput.vue
pleek91: Otherwise when you resize you can get into a weird spot.

pleek91: This component is basically what the default should be for a json input. Someday we should fix that lol

pleek91 (approved):

src/components/FlowRunCreateFormV2.vue
collincchoy: It could be. 

tho it might be better here as it being a string is more of an implementation detail of the form that it's in so that it plops into a textarea. maybe if `DeploymentFlowRunCreate` is actually `FlowRunCreateFormValues` and so then a mapper going from `FlowRunCreateFormValues` -> `DeploymentFlowRunCreateRequest` seems sensible to encapsulate the stringifying. 

that gel w/ ur mental model too?

src/components/JobVariableOverridesInput.vue
collincchoy: Ohhhh I think I did my rem math wrong - I see the min-h stopping you from resizing down too small now. 

still not that sure about the min-w-full - feels like it was maybe an artifact from being in some flex container where it wasn't stretching but not sure

pleek91: `min-w-full` might not be needed. Could be leftover from using `resize` initially.

src/components/FlowRunCreateFormV2.vue
pleek91: Both seem logical. Being an implementation detail makes sense too, probably wouldn't of thought about the mapper if I hadn't seen it in both form components. But one of these will go away in a week or so anyway.

PR #2171, opened 2024-03-06
Schemas: Make sure boolean properties default to false

pleek91:
# Description
Make sure that when a boolean property mounts if the value is undefined it automatically sets it to false. Otherwise the value (`undefined`) will be omitted from the validation and flow run create causing validation errors and runtime errors.

### DIFF
--- a/src/schemas/components/SchemaFormPropertyBoolean.vue
+++ b/src/schemas/components/SchemaFormPropertyBoolean.vue
@@ -1,0 +1,27 @@
+<template>
+  <p-toggle v-model="value" :state="state" />
+</template>
+
+<script lang="ts" setup>
+  import { State, isDefined } from '@prefecthq/prefect-design'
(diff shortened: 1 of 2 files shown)


src/schemas/components/SchemaFormPropertyBoolean.vue
```diff
@@ -0,0 +1,27 @@
...
+    value.value = false
+  }
+</script>
+
+<style>
+.schema-form-property-null { @apply
+  text-subdued
+  text-sm
+}
+</style>
```
znicholasbrown, line 27: Seems like this isn't needed

znicholasbrown (approved):

src/schemas/components/SchemaFormPropertyBoolean.vue
pleek91: Thank you! copypasta...

PR #2176, opened 2024-03-06
Enhancement: Graph empty state message

znicholasbrown:
This PR updates the flow run graph component to better account for terminal and non-terminal states. 

Before (regardless of state):
Screenshot 2024-03-06 at 3 38 56 PM

After (non-terminal):
Screenshot 2024-03-06 at 4 28 28 PM

After (terminal):
Screenshot 2024-03-06 at 4 29 09 PM

As flow run and task run lines become blurrier we might trim this language a bit but for now I think this is at least correct


pleek91 (approved):

brandonreid (approved):
:chefs-kiss:

Ty!

PR #2183, opened 2024-03-07
[Flow run infra overrides] Make flow run job vars available for reading

collincchoy:
PR exposes new `job_variables` field on flow run responses.

### DIFF
--- a/src/models/api/FlowRunResponse.ts
+++ b/src/models/api/FlowRunResponse.ts
@@ -38,5 +38,5 @@ export type FlowRunResponse = {
   work_queue_name: string | null,
   work_pool_name: string | null,
   work_pool_queue_name: string | null,
-  job_variables: Record<string, unknown> | null,
+  job_variables?: Record<string, unknown> | null,
 }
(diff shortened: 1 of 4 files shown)


pleek91 (approved):

src/models/api/FlowRunResponse.ts
```diff
@@ -38,4 +38,5 @@ export type FlowRunResponse = {
   work_queue_name: string | null,
   work_pool_name: string | null,
   work_pool_queue_name: string | null,
+  job_variables?: Record<string, unknown> | null,
```
collincchoy, line 41: `?` for compatibility w/ OSS - open PR over there to add this field so atm this may or may not be there, but coalescing to `{}` in the mapper either way

collincchoy: think we should be able to remove this `| null` also but gonna wait for the apis to stabilize

pleek91: Make sense, thanks!

PR #2191, opened 2024-03-08
[Flow run infra overrides] Override job variables per deployment schedule

collincchoy:
This PR adds a JSON input for flow run job variable overrides into the deployment schedules modal/form --- which enables a user to configure job variable overrides on a per deployment schedule basis through the UI. 

> [!Important]
> This work is feature flagged. atm, api changes are still pending so we need to be careful here to maintain backwards compatibility with the flag off (aka not send the extra field across the wire).

| **Flag OFF** | **Flag ON** |
| ---- | ---- |
| image image | image image |
| image image | image image |


src/components/FlowRunCreateFormV2.vue
```diff
@@ -127,8 +125,7 @@
   const workQueueName = ref<string | null>(props.deployment.workQueueName)
   const retries = ref<number | null>(null)
   const retryDelay = ref<number | null>(null)
-  const jobVariables = ref<string>()
-  const { state: jobVariablesState, error: jobVariablesError } = useValidation(jobVariables, isJson('Job variables'))
+  const jobVariables = ref<string | undefined>(can.access.flowRunInfraOverrides ? '{}' : undefined)
```
collincchoy, line 128: This change is to make the input here initially have `{}` rather than be empty, _but_ it needs to still be undefined if the flag is off so that the field is dropped when it goes over the wire.

image

src/models/DeploymentScheduleCreate.ts
```diff
@@ -3,4 +3,5 @@ import { Schedule } from '@/models/Schedule'
 export type DeploymentScheduleCreate = {
   active: boolean,
   schedule: Schedule,
+  jobVariables?: Record<string, unknown>,
```
collincchoy, line 6: needs to be optional going out for compatibility across api versions

src/components/DeploymentDetails.vue
```diff
@@ -96,7 +96,7 @@
     }
 
     try {
-      await api.deploymentSchedules.createDeploymentSchedule(props.deployment.id, { active: updatedSchedule.active, schedule: updatedSchedule.schedule })
+      await api.deploymentSchedules.createDeploymentSchedule(props.deployment.id, { active: updatedSchedule.active, schedule: updatedSchedule.schedule, jobVariables: updatedSchedule.jobVariables })
```
pleek91, line 99: Any idea why this cannot just be this? 
```suggestion
      await api.deploymentSchedules.createDeploymentSchedule(props.deployment.id, updatedSchedule)
```

src/components/ScheduleFormModal.vue
```diff
@@ -47,19 +52,37 @@
...
+    }
 
-  const submit = (schedule: DeploymentScheduleCompatible): void => {
-    emit('submit', schedule)
+    let jobVariables: Record<string, unknown> | undefined
+    if (!can.access.flowRunInfraOverrides) {
+      jobVariables = undefined
+    } else {
+      jobVariables = internalJobVariables.value ? JSON.parse(internalJobVariables.value) : undefined
+    }
```
pleek91, line 79: is stuff like this only necessary during development and will eventually get cleaned up or is this something we'll have to deal with for backwards compatibility for a while?

pleek91 (approved):

src/components/ScheduleFormModal.vue
collincchoy: the former - this is the "don't send the extra parameter to the api". in hindsight mb i could've done this in the mapper and only have 1 place to update.. 

hoping this doesn't have to last too long.

src/components/DeploymentDetails.vue
collincchoy: Yeah there's a _slight_ typing mismatch 
image

probably why that `if (updatedSchedule.active === null)` check is here. 

I think there's some refactor that could simplify this but didn't get all the way there.

PR #2194, opened 2024-03-11
Restrict logs to only `a` html elements

pleek91:
# Description
Log rows use v-html but first passes the content of the log through `p-sanitize-html` with a custom config. That config is overriding the default config which seems to mean all html tags are allowed. Specifying only the specific tags we want to allow prevents css and js injection.

### DIFF
--- a/src/components/LogRow.vue
+++ b/src/components/LogRow.vue
@@ -39,6 +39,7 @@
   }>()
 
   const config = {
+    ALLOWED_TAGS: ['a'],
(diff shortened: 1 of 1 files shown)


pleek91:
@znicholasbrown can you confirm that we should only be allowing `a` tags?

znicholasbrown (approved):
Yeah this looks good - in my mind passing the bespoke config to `p-sanitize-html` would result in a superset of the configs but in hindsight that doesn't make much sense and WYSIWYG is a lot clearer. Thanks for handling this

PR #2202, opened 2024-03-12
Schemas: Fix prop and v-model not matching up for boolean properties

### LINKED ISSUE #12241: Boolean (optional) parameters are not copied to a new flow
### First check

### Bug summary

After copying a deployment run, the (optional) boolean parameters are not copied, but instead just reset to their defaults.

### Reproduction

1. Create a deployment with a boolean value with a default set (for example, False) .
2. Run the deployment with the boolean set to the non-default setting (in this case: True)
3. After submitting, open the run and copy it to a new run
4. Check if the boolean setting is reset to False again.

### Error

When I copy a deployment run to a new run, the (optional) boolean parameters are just reset to their defaults. Can this please be fixed? I accidentally generated a lot of more computations because of this :')

### Browsers

- [X] Chrome
- [ ] Firefox
- [ ] Safari
- [X] Edge

### Prefect version

_No response_

### Additional context

_No response_

pleek91:
# Description
The v-model was broken because the prop being passed was `value` but the v-model was `modelValue`. Updated all references to use `value`. 

Closes: https://github.com/PrefectHQ/prefect/issues/12241

### DIFF
--- a/src/schemas/components/SchemaFormPropertyBoolean.vue
+++ b/src/schemas/components/SchemaFormPropertyBoolean.vue
@@ -12,7 +12,7 @@
     state: State,
   }>()
 
-  const value = defineModel<boolean | undefined>({ default: undefined })
+  const value = defineModel<boolean | undefined>('value', { default: undefined })
 
   if (!isDefined(value.value)) {
     value.value = asType(props.property.default, Boolean) ?? false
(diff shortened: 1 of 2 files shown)


bunchesofdonald (approved):
Makes sense to me

znicholasbrown (approved):

PR #2207, opened 2024-03-12
Fix deployment table hiding overflowing flow run bar chart popovers.

### LINKED ISSUE #2206: Flow runs bar chart popover cutoff by table
There's an overflow/stacking issue with the flow runs bar chart popover (`flow-runs-bar-chart__pop-over`) from within the Deployments table. 

See in the below screenshot that with a few deployments, the popover when hovering over the flow runs bars is completely hidden. 

> [!Note]
>  the pink line is added for visual effect and not a part of normal behavior.

image

The Deployments table (via `p-table`) has `overflow: hidden` and this popover doesn't portal out so it gets hidden.

collincchoy:
> [!Note]
> In draft for moment - Doing a quick audit of other usages...

| **Before** | **After** |
| ---- | ---- |
| image | image |

Fixes #2206

### DIFF
--- a/src/components/FlowRunsBarChart.vue
+++ b/src/components/FlowRunsBarChart.vue
@@ -1,5 +1,6 @@
 <template>
-  <div ref="chart" class="flow-runs-bar-chart" :class="classes.root" :style="styles.root" @mouseleave="close">
+  <div ref="chart" class="flow-runs-bar-chart" :class="classes.root" :style="styles.root">
+    <!-- @mouseleave="close"> -->
     <template v-for="(flowRun, index) in barFlowRuns" :key="getKey(flowRun, index)">
       <p-pop-over class="flow-runs-bar-chart__bar-container" :placement="placement" :group="group" auto-close>
         <template #target="{ open }">
--- a/src/components/FlowRunsBarChart.vue
+++ b/src/components/FlowRunsBarChart.vue
@@ -1,7 +1,7 @@
 <template>
   <div ref="chart" class="flow-runs-bar-chart" :class="classes.root" :style="styles.root" @mouseleave="close">
     <template v-for="(flowRun, index) in barFlowRuns" :key="getKey(flowRun, index)">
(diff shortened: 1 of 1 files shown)


src/components/FlowRunsBarChart.vue
```diff
@@ -1,7 +1,7 @@
 <template>
   <div ref="chart" class="flow-runs-bar-chart" :class="classes.root" :style="styles.root" @mouseleave="close">
     <template v-for="(flowRun, index) in barFlowRuns" :key="getKey(flowRun, index)">
-      <p-pop-over class="flow-runs-bar-chart__bar-container" :to="chart" :placement="placement" :group="group" auto-close>
```
collincchoy, line 4: `:to` is the teleport target and the default is the body so then this change has the popover content teleporting out of the table thus solving the overflow: hidden issue. 

tho i wonder what use-case `:to="chart"` might be for 🤔

collincchoy:
Ah okay I found an issue with this - while hovering over the bars does correctly render and position the popover content how I'd expect (regardless of a container with overflow rules), moving the mouse from the bar over to the popover content to interact with it now loses focus and the popover closes w/o being able to interact with the content.

src/components/FlowRunsBarChart.vue
```diff
@@ -1,7 +1,8 @@
 <template>
-  <div ref="chart" class="flow-runs-bar-chart" :class="classes.root" :style="styles.root" @mouseleave="close">
```
collincchoy, line 2: Removing this `@mouseleave="close"` fixes the issue where moving your mouse from the bar over to the popover closes the popover. 

however it introduces a UX regression in that the popover lingers onscreen until a click triggers `p-popover`'s `auto-close` behavior. 

Ideally the popover closes on mouseleave (of both the chart _and_ the popover contents) again. 
Needs some changes in prefect-design.

collincchoy:
Need to look into porting some of p-tooltip's hover-out autoclosing behavior into p-popover. 

Maybe a `<p-popover autoclose="'mouseleave' | 'outerclick'" />` or something. 

Closing this PR for now.

PR #2210, opened 2024-03-13
Make state badge prefer its state prop's type if present.

collincchoy:
This PR contains a small change to have the StateBadge use its state type directly for styling if present _else_ fallback to the current method of deriving the type from its name. 

This enables custom flow run states to render in styled like their actual, inherent state type while maintaining the custom name. 

```python
@flow
def hello_world():
    return State(
        type=StateType.COMPLETED,
        name="My custom name",
        message="My custom message.",
    )
```

| **Before** | **After** |
| ---- | ---- |
| image | image |

### DIFF
--- a/src/components/StateBadge.vue
+++ b/src/components/StateBadge.vue
@@ -18,7 +18,7 @@
     dismissible?: boolean,
   }>()
 
-  const type = computed(() => mapStateNameToStateType(props.state?.name).type)
+  const type = computed(() => props.state?.type ?? mapStateNameToStateType(props.state?.name).type)
   const stateWithNoType = computed(() => props.state && !type.value)
   const name = computed(() => props.state?.name ?? 'Unknown')
 


src/components/StateBadge.vue
```diff
@@ -18,7 +18,7 @@
     dismissible?: boolean,
   }>()
 
-  const type = computed(() => mapStateNameToStateType(props.state?.name).type)
+  const type = computed(() => props.state?.type ?? mapStateNameToStateType(props.state?.name).type)
```
collincchoy, line 21: In the example flow in the PR description, `props.state.type === 'completed'` already but the state has a custom name so before, it wouldn't map to any type.

pleek91 (approved):
Ah nice, yeah should have been this way to begin with. TY!

znicholasbrown (approved):
love this

PR #2211, opened 2024-03-13
Enhancement: Virtualize row-grid-layout-list; bump prefect-design

znicholasbrown:
I think this package was artificially holding back downstream packages from getting the latest prefect-design version so bumped that here. 

Related: https://github.com/PrefectHQ/prefect-design/pull/1145

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -43,7 +43,7 @@
                 "vue-tsc": "2.0.6"
             },
             "peerDependencies": {
-                "@prefecthq/prefect-design": "^2.3.3",
+                "@prefecthq/prefect-design": "2.3.7",
                 "@prefecthq/vue-charts": "^2.0.3",
                 "@prefecthq/vue-compositions": "^1.10.0",
                 "vee-validate": "^4.7.0",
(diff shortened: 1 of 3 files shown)


pleek91 (approved):
Hmm a peer with a ^ shouldn't prevent updates in oss or nebula-ui.

znicholasbrown:
Sorry I worded that incorrectly @pleek91 - I didn't dig too deep but I think having a mismatched lowerbound between the application that installs prefect-design and this package means we get 2 versions of prefect-design. Then any components here use the version installed by this package and any components in downstream applications use whatever their version is

pleek91:
> Sorry I worded that incorrectly @pleek91 - I didn't dig too deep but I think having a mismatched lowerbound between the application that installs prefect-design and this package means we get 2 versions of prefect-design. Then any components here use the version installed by this package and any components in downstream applications use whatever their version is

Ah I see, so that is an issue with using globally registered prefect-design components or importing them directly. Using `p-table` in the template (without importing it) will use the globally registered version (registered by oss or nebula-ui). But if you import `PTable` directly then it ends up bundling `PTable` in ui-library using whatever version is targeted here. 

I think we could fix that by always using [`resolveComponent`](https://vuejs.org/api/render-function.html#resolvecomponent)

PR #2215, opened 2024-03-13
Schemas: Add support for interactive workflows

pleek91:
# Description
Adds V2 modal and service methods for interactive workflows enabling users to get the same enhanced deployment parameters ui for flow run input as they do with deployment parameters. 

image

### DIFF
--- a/src/components/FlowRunResumeModalV2.vue
+++ b/src/components/FlowRunResumeModalV2.vue
@@ -1,20 +1,16 @@
 <template>
   <p-modal v-if="flowRun" v-model:showModal="showModal" title="Resume Flow Run">
-    <p-message v-if="serverValidationError" error>
-      {{ serverValidationError }}
-    </p-message>
+    <template v-if="schema">
+      <p-markdown-renderer v-if="description" :text="description" />
 
-    <template v-if="inputSchema">
-      <p-markdown-renderer v-if="inputDescription" :text="inputDescription" />
+      <p-content v-if="schema">
+        <SchemaInputV2 v-model:values="parameters" :schema="schema" :errors="errors" :kinds="['none', 'json', 'workspace_variable']" />
+      </p-content>
     </template>
 
     <p v-else>
       Do you want to resume this flow run?
     </p>
-
-    <template v-if="inputSchema">
-      <SchemaInputV2 v-model:values="parameters" :schema="inputSchema" :errors="[]" :kinds="['none', 'json', 'workspace_variable']" />
-    </template>
 
     <template #actions>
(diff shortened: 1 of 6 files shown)


src/components/FlowRunResumeModalV2.vue
```diff
@@ -0,0 +1,89 @@
...
+      return
+    }
+
+    try {
+      const response = await api.flowRuns.resumeFlowRunV2(props.flowRunId, parameters.value)
+
+      if (response.status != 'ACCEPT') {
+        showToast(localization.error.resumeFlowRun, 'error')
+        console.log('Orchestration failed: ', response.details.reason)
+        serverValidationError = response.details.reason
```
abrookins, line 71: Hey, while you're in here, do you know how we'd clear this value when someone opens the form? I think if you get a server error, the message lingers on the form after you close and reopen it. But it probably shouldn't. 🤔

pleek91: Ah yeah I can figure out something for that 👍

src/components/FlowRunResumeModalV2.vue
```diff
@@ -0,0 +1,89 @@
...
+  import { SchemaV2, SchemaValuesV2, SchemaInputV2 } from '@/schemas'
+  import { getApiErrorMessage } from '@/utilities/errors'
+
+  const props = defineProps<{
+    flowRunId: string,
+  }>()
+
+  const showModal = defineModel<boolean>('showModal')
+  const api = useWorkspaceApi()
+  const parameters = ref<SchemaValuesV2>({})
```
znicholasbrown, line 42: Just double checking - downstream schema inputs will populate this with defaults, right?

znicholasbrown (approved):
one question, otherwise lgtm

src/components/FlowRunResumeModalV2.vue
pleek91: Correct, the component itself will automatically fill in the default values. That way you don't have to worry about it at implementation.

src/components/FlowRunResumeModalV2.vue
pleek91: Update in https://github.com/PrefectHQ/prefect-ui-library/pull/2215/commits/48750e01407ebac2c1594150d8fcba551288e05b to automatically reset the model when its closed.

znicholasbrown (approved):

src/components/FlowRunResumeModalV2.vue
abrookins: You are THE MAN

PR #2217, opened 2024-03-13
Add timeout and refresh logs again

zhen0:
In the UI we refresh our logs when a run enters a completed state but in short runs logs sometimes are not yet uploaded.  The UI then shows a "This run didn't generate logs" message for logs until the page is refreshed. 

This PR adds a timeout and a second log subscription refresh to try polling for logs again.  

Before:
https://github.com/PrefectHQ/prefect-ui-library/assets/40272060/523a8ecc-82d8-473e-882b-4fbd783c7174

Improved:
https://github.com/PrefectHQ/prefect-ui-library/assets/40272060/f840a94d-1112-44ed-a672-3b1ec9c6efaa

Resolves https://github.com/PrefectHQ/nebula-ui/issues/4304 

There's still a potential flicker of "This run didn't generate logs" before the logs appear on the page.  I think that's better than adding a long delay for all logs.  Open to other suggestions!  This is one that I hope future work on web sockets or similar can better address long term.


pleek91 (approved):
That is interesting, hopefully we can make this better by streaming logs via websocket for running runs.

PR #2224, opened 2024-03-14
Allow prefect ui library to be updated when package updates occur upstream

jimid27:
- Relates: https://github.com/PrefectHQ/platform/issues/6282


znicholasbrown (approved):
thanks @jimid27 !

PR #2226, opened 2024-03-14
Schemas: fix reactivity issue

pleek91:
# Description
Debugging a [customer reported issue ](https://prefecthq.slack.com/archives/CJN6L95L7/p1709742115494889) lead me to realizing that the `readonly` that I had added in https://github.com/PrefectHQ/prefect-ui-library/pull/2144 was causing issues if you used the deployment parameters as a default value without first cloning/merging them into a new object. Changing that to a `structuredClone` which will prevent reactivity issues while also severing any object references that could get mutated elsewhere. 

I'm hoping this will no longer be needed after the schemas v1 code is removed from deployments. Added a giant comment with a detailed reproduction so that future me can know how to test this when cleaning up. 

I'd rather not spend more hours digging into schemas code that is going to be deleted to try and fix this. So this is a band-aid.


znicholasbrown (approved):
future you will appreciate this. 

thanks @pleek91

PR #2230, opened 2024-03-15
Add counts by state type to the flow run task runs tab

### LINKED ISSUE #12102: Display count of task runs by state on flow run page
### First check

### Prefect Version

2.x

Users can see a list of task runs on a flow run's page (in the Task Runs tab). It is not possible to see how many tasks are in each possible state. This becomes problematic when there are 100s or 1000s of tasks. In order to determine how many tasks are in each state, the user must apply each state filter and keep track of the counts.

It would be super helpful if we could display a count of tasks in each state somewhere at the top of the Task Runs tab on the flow run page. For example, where it says `n Task runs`, perhaps it could say `n Task runs (all Completed)` or `n Task runs (x Completed, y Failed)`. How it's surfaced is less important than whether it's surfaced at all.

### Example Use

_No response_

### Additional context

Replaces #8028

pleek91:
# Description
Adds a count of tasks by their state type below the total count to easily identify how many tasks

Closes https://github.com/PrefectHQ/prefect/issues/12102

@brandonreid if you have some design feedback on this let me know. 

image

### DIFF
--- a/src/components/FlowRunTaskCounts.vue
+++ b/src/components/FlowRunTaskCounts.vue
@@ -1,0 +1,48 @@
+<template>
+  <span v-if="counts" class="flow-run-task-counts">
+    ({{ counts }})
+  </span>
+</template>
+
+<script lang="ts" setup>
+  import { useSubscription } from '@prefecthq/vue-compositions'
+  import { computed } from 'vue'
+  import { useWorkspaceApi } from '@/compositions/useWorkspaceApi'
+  import { TaskRunsFilter } from '@/models/Filters'
+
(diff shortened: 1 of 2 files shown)


brandonreid:
This looks good from a design perspective imo, at least for now. I know @collincchoy has a checkbox that will align to the left of the count for bulk selection coming. I also have some hanging intent around similar counts on the flow/deployment list pages that will be clickable and serve as a filter. So we may be approaching a time where this header pattern requires some holistic thinking. Again, I think this looks nice for now though.

brandonreid (approved):

PR #2239, opened 2024-03-20
Enhancement: Add hook to axios instance setup

bunchesofdonald:
This adds support to the `Api` class to allow passing in a hook that will receive the `axios` instance so that things like axios interceptors (https://axios-http.com/docs/interceptors) can be attached. This is to support the CSRF implementation here: https://github.com/PrefectHQ/prefect/pull/12354

Related to https://github.com/PrefectHQ/prefect/issues/12347

### DIFF
--- a/src/services/Api.ts
+++ b/src/services/Api.ts
@@ -3,6 +3,8 @@ import axios, { AxiosInstance, AxiosRequestConfig, AxiosResponse, RawAxiosReques
 import { MaybeArray } from '@/types/utilities'
 import { isDefined } from '@/utilities/variables'
 
+export type AxiosInstanceSetupHook = (instance: AxiosInstance) => void
+
 export type PrefectConfig = {
   baseUrl: string,
   token?: string,
@@ -31,9 +33,11 @@ export class Api<T extends PrefectConfig = PrefectConfig> {
   protected apiHeaders: MaybeArray<ApiHeaders> = [getPrefectUIHeaders, getAuthorizationHeaders]
   protected apiBaseUrl: ApiBaseUrl = getPrefectBaseUrl
   protected routePrefix: string | undefined
+  protected instanceSetupHook: AxiosInstanceSetupHook | null
 
-  public constructor(apiConfig: T) {
+  public constructor(apiConfig: T, instanceSetupHook: AxiosInstanceSetupHook | null = null) {
     this.apiConfig = apiConfig
+    this.instanceSetupHook = instanceSetupHook
   }
 
   protected composeBaseUrl(): string {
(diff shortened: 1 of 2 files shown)


src/utilities/api.ts
```diff
@@ -28,32 +29,32 @@ import { WorkspaceWorkQueuesApi } from '@/services/WorkspaceWorkQueuesApi'
 
 // We want the return type to be inferred
 // eslint-disable-next-line @typescript-eslint/explicit-function-return-type
-export function createApi(workspaceConfig: WorkspaceApiConfig) {
+export function createApi(workspaceConfig: WorkspaceApiConfig, instanceSetupHook: AxiosInstanceSetupHook | null = null) {
```
znicholasbrown, line 32:
```suggestion
export function createApi(workspaceConfig: WorkspaceApiConfig, instanceSetupHook?: AxiosInstanceSetupHook) {
```

src/services/Api.ts
```diff
@@ -31,9 +33,11 @@ export class Api<T extends PrefectConfig = PrefectConfig> {
   protected apiHeaders: MaybeArray<ApiHeaders> = [getPrefectUIHeaders, getAuthorizationHeaders]
   protected apiBaseUrl: ApiBaseUrl = getPrefectBaseUrl
   protected routePrefix: string | undefined
+  protected instanceSetupHook: AxiosInstanceSetupHook | null
```
znicholasbrown, line 36:
```suggestion
  protected instanceSetupHook: AxiosInstanceSetupHook | undefined
```

src/services/Api.ts
```diff
@@ -31,9 +33,11 @@ export class Api<T extends PrefectConfig = PrefectConfig> {
...
 
-  public constructor(apiConfig: T) {
+  public constructor(apiConfig: T, instanceSetupHook: AxiosInstanceSetupHook | null = null) {
```
znicholasbrown, line 38:
```suggestion
  public constructor(apiConfig: T, instanceSetupHook?: AxiosInstanceSetupHook) {
```

znicholasbrown (approved):
1 set of suggestions so callers don't need to pass `null` to the api constructor

pleek91:
My only thought is it might make sense to add this to the `PrefectConfig` as an optional property. That way there's just a single config object to pass. Otherwise LGTM

bunchesofdonald:
> My only thought is it might make sense to add this to the `PrefectConfig` as an optional property. That way there's just a single config object to pass. Otherwise LGTM

I had it there initially, but this felt different than the config itself as it's a hook that's just inteded to be an _extension_ of this class, which felt different than the real 'config' things like baseURL. If you feel strongly I'm happy to change it.

PR #2251, opened 2024-03-22
Use p table multiselect behavior in tables

collincchoy:
These table components implement their own selection behavior despite `<p-table />` having support for multiselect built-in. 

This PR updates all tables with selection behavior to use `<p-table>`'s built-in multiselect behavior.
**There should be close to no functional changes expected from this PR** (with the exception of `<WorkPoolQueueTable />` - see comments)

### DIFF
--- a/src/components/DeploymentsList.vue
+++ b/src/components/DeploymentsList.vue
@@ -3,7 +3,7 @@
     <p-list-header sticky>
       <ResultsCount v-if="selectedDeployments.length == 0" label="Deployment" :count="total" />
       <SelectedCount v-else :count="selectedDeployments.length" />
-      <DeploymentsDeleteButton v-if="can.delete.deployment" :selected="selectedDeployments" small @delete="deleteDeployments" />
+      <DeploymentsDeleteButton v-if="can.delete.deployment" :selected="selectedDeployments.map(deployment => deployment.id)" small @delete="deleteDeployments" />
 
       <template #controls>
         <SearchInput v-model="deploymentNameLike" placeholder="Deployment names" label="Search deployments" />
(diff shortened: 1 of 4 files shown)


src/components/WorkPoolQueuesTable.vue
```diff
@@ -1,20 +1,18 @@
 <template>
   <p-content class="work-pool-queues-table">
     <p-list-header sticky>
-      <template v-if="selected">
-        <ResultsCount v-if="selected.length == 0" label="Work Queue" :count="filteredWorkPoolQueues.length" />
-        <SelectedCount v-else :count="selected.length" />
-        <p-button v-if="can.create.work_queue && !selected.length" small icon="PlusIcon" :to="routes.workPoolQueueCreate(workPoolName)" />
+      <ResultsCount v-if="selected.length == 0" label="Work Queue" :count="filteredWorkPoolQueues.length" />
```
collincchoy, line 4: This ResultsCount (i.e. `3 Work Queues`) along with the Plus button to create a new work queue were previously hidden entirely if `can.delete.work_queue === false`.

src/components/WorkPoolQueuesTable.vue
```diff
@@ -110,7 +108,7 @@
 
   const handleDelete = async (): Promise<void> => {
     await workPoolQueuesSubscription.refresh()
-    selected.value = selected.value?.filter(queue => workPoolQueues.value.find(({ id }) => id === queue.id))
+    selected.value = []
```
collincchoy, line 111: just deselect everything on successful deletion. that's what all the other tables do

pleek91 (approved):
Nice, thanks for cleaning this up

src/components/DeploymentsList.vue
```diff
@@ -15,15 +15,7 @@
...
-    <p-table :data="deployments" :columns="columns" class="deployments-list__table">
-      <template #selection-heading>
-        <p-checkbox v-model="model" @update:model-value="selectAllDeployments" />
-      </template>
-
-      <template #selection="{ row }">
-        <p-checkbox v-model="selectedDeployments" :value="row.id" />
-      </template>
-
+    <p-table :selected="can.delete.deployment ? selectedDeployments : undefined" :data="deployments" :columns="columns" class="deployments-list__table" @update:selected="selectedDeployments = $event">
```
collincchoy, line 18: Don't _love_ having to break up the v-model here. 

With a computed `selectedDeploymentsModel` with a setter/getter though, the value becomes possibly undefined which means a bunch of `?.` need to get added around whereas this isolates the nullish-ness. 

It also ends up being more boilerplatey code and see further down in `WorkPoolQueuesTable` where something like that was done before and it led to an innocuous little bug 🤷 .

PR #2252, opened 2024-03-25
Schemas: Better UX for array property enums 

pleek91:
# Description
The list input is nice but not very ergonomic if there is a finite list of things to pick from (an enum).

This updates the array logic to check if a `items` includes an enum and then uses a combobox. 

```
class Color(Enum):
    RED = "red"
    GREEN = "green"
    BLUE = "blue"

@flow
def schema_testing(value: list[Color]) -> None:
  pass
```

Before
image

After
image

Note: One downside to this is combobox doesn't support duplicate values. But that hasn't been an ask from users. And the ux of the list input was a concern. If duplicate values become an issue I think we can solve that by giving users the ability to switch to the list input. For now, json is always an option.


src/schemas/components/SchemaFormPropertyArray.vue
```diff
@@ -45,92 +21,48 @@
...
-
-    return property.default ?? undefined
+  function isSelectModalValue(value: unknown): value is SelectModelValue {
+    return isNumber(value) || isString(value) || isBoolean(value) || isNull(value)
   }
 
-  function isFirstIndex(index: number): boolean {
-    return index === 0
-  }
+  function asSelectModalValue(value: unknown): SelectModelValue[] | undefined {
```
znicholasbrown, line 35:
```suggestion
  function asSelectModelValue(value: unknown): SelectModelValue[] | undefined {
```

src/schemas/components/SchemaFormPropertyArray.vue
```diff
@@ -45,92 +21,48 @@
...
+    if (!Array.isArray(value)) {
+      return undefined
+    }
 
-  function isLastIndex(index: number): boolean {
-    return index === (props.value?.length ?? 0) - 1
+    return value.filter(isSelectModalValue)
```
znicholasbrown, line 40:
```suggestion
    return value.filter(isSelectModelValue)
```

src/schemas/components/SchemaFormPropertyArray.vue
```diff
@@ -45,92 +21,48 @@
...
 
-.schema-form-property-array__empty { @apply
-  text-subdued
-  text-sm
-  italic;
-  margin-left: var(--schema-form-property-array-item-indent);
-}
+      if (isSchemaPropertyPrimitiveType(type) && propertyEnum) {
+        return withProps(PCombobox, {
+          modelValue: asSelectModalValue(props.value),
```
znicholasbrown, line 51:
```suggestion
          modelValue: asSelectModelValue(props.value),
```

src/schemas/components/SchemaFormPropertyArray.vue
```diff
@@ -45,92 +21,48 @@
...
+          state: props.state,
+          options: propertyEnum.filter(isSelectModalValue).filter(isNotNullish),
+          multiple: true,
+          'onUpdate:modelValue': value => emit('update:value', asSelectModalValue(value)),
```
znicholasbrown, line 56:
```suggestion
          'onUpdate:modelValue': value => emit('update:value', asSelectModelValue(value)),
```

src/schemas/components/SchemaFormPropertyArray.vue
```diff
@@ -45,92 +21,48 @@
...
-  text-subdued
-  text-sm
-  italic;
-  margin-left: var(--schema-form-property-array-item-indent);
-}
+      if (isSchemaPropertyPrimitiveType(type) && propertyEnum) {
+        return withProps(PCombobox, {
+          modelValue: asSelectModalValue(props.value),
+          state: props.state,
+          options: propertyEnum.filter(isSelectModalValue).filter(isNotNullish),
```
znicholasbrown, line 54:
```suggestion
          options: propertyEnum.filter(isSelectModelValue).filter(isNotNullish),
```

src/schemas/types/schema.ts
```diff
@@ -55,6 +63,10 @@ export type SchemaProperty = {
   maxItems?: number,
 }
 
+export function isSchemaProperty(value: SchemaProperty | SchemaProperty[] | undefined): value is SchemaProperty {
+  return isDefined(value) && !Array.isArray(value)
+}
```
znicholasbrown, line 68: Do we need to introduce a way to convert from `SchemaProperty[]` to `SchemaProperty`? We'll need to maintain backwards compatibility with v1 for older versions of core.

znicholasbrown (approved):
typo fix, otherwise lgtm

src/schemas/types/schema.ts
pleek91: It crossed my mind to map `items: SchemaProperty[]` to `prefixItems` which is the correct way to do that. But for now I think what we have works fine.

znicholasbrown: ok we can revisit as needed!

src/schemas/components/SchemaFormPropertyArray.vue
pleek91: ty, fixed in https://github.com/PrefectHQ/prefect-ui-library/pull/2252/commits/c17012aa5c3e8215f0c1760703a832a5bf6c41f8

PR #2259, opened 2024-03-26
Schemas: Set default value for json inputs

### LINKED ISSUE #2244: Display parameter schema in json view of parameters
When you switch to json view, you get an empty text input. We should have it display the parameter json schema, with any default values populated as well. This allows users to edit over generate json from scratch.

pleek91:
# Description
When switching to json we can fill in the default values so that the user has something to work from rather than just showing them an empty json input. 

image

closes https://github.com/PrefectHQ/prefect-ui-library/issues/2244

### DIFF
--- a/src/schemas/components/SchemaFormPropertyKindJson.vue
+++ b/src/schemas/components/SchemaFormPropertyKindJson.vue
@@ -6,14 +6,19 @@
 </template>
 
 <script lang="ts" setup>
-  import { State } from '@prefecthq/prefect-design'
+  import { State, isDefined } from '@prefecthq/prefect-design'
   import { computed } from 'vue'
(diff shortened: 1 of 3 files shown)


collincchoy:
Updating the value using a form/input and then switching to JSON doesn't seem to update the value in the JSON input.

Trying to hunt down where/why but is this expected behavior atm? 

i.e. - a schema with a single boolean property that's default false. 
1. load up a form.
2. see that there's a toggle that's off ✔️ 
3. flip to json and see false ✔️ 
4. flip back to form
5. flip toggle on ✔️ 
6. flip to json and see false ❌ 

The reverse flow of updating using the JSON input and then switching back to form also doesn't sync like I'd expect 🤔

collincchoy (approved):
my other comment here is out-of-scope for the issue here 👍

PR #2261, opened 2024-03-27
Add pause-deployment and resume-deployment actions from nebula-ui

pleek91:
# Description
First piece of ui work in the broader goal of bringing automations to oss. Sets up the ability for automation actions to be defined in ui-library that can be shared with oss and cloud. 

Starting by bringing over the `pause-deployment` and `resume-deployment` actions into ui-library. Creating exports that nebula-ui can import and extend.

### DIFF
--- a/src/automations/components/AutomationActionInput.vue
+++ b/src/automations/components/AutomationActionInput.vue
@@ -1,0 +1,39 @@
+<template>
+  <component :is="input.component" v-bind="input.props" class="automation-action-input" />
+</template>
+
+<script lang="ts" setup>
+  import { computed } from 'vue'
+  import AutomationActionPauseDeploymentInput from '@/automations/components/AutomationActionPauseDeploymentInput.vue'
+  import AutomationActionResumeDeploymentInput from '@/automations/components/AutomationActionResumeDeploymentInput.vue'
+  import { AutomationAction } from '@/automations/types'
+  import { withProps } from '@/utilities'
+
+  const props = defineProps<{
+    action: Partial<AutomationAction>,
+  }>()
+
+  const emit = defineEmits<{
+    'update:action': [Partial<AutomationAction>],
+  }>()
+
+  const input = computed(() => {
+    switch (props.action.type) {
+      case 'pause-deployment':
+        return withProps(AutomationActionPauseDeploymentInput, {
+          action: props.action,
+          'onUpdate:action': value => emit('update:action', value),
+        })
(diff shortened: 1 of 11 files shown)


src/automations/components/AutomationActionInput.vue
```diff
@@ -0,0 +1,39 @@
...
+          action: props.action,
+          'onUpdate:action': value => emit('update:action', value),
+        })
+      case 'resume-deployment':
+        return withProps(AutomationActionResumeDeploymentInput, {
+          action: props.action,
+          'onUpdate:action': value => emit('update:action', value),
+        })
+      case undefined:
+        throw new Error('AutomationActionInput.vue action.type is undefined')
```
pleek91, line 33: This is a little weird. Because of the type `Partial<AutomationAction>` the types say type can be undefined. Which we don't want, and the implementation in nebula-ui makes sure this cannot happen. But the types allow for it. I tried a few ways to fix this but each thing I tried didn't really work and also created complexity. 

So opting for this for now. Matches existing functionality but also makes sure that we get good type errors when adding new actions.

src/automations/components/AutomationActionPauseDeploymentInput.vue
pleek91, line 1: Copied directly from ui-library

src/automations/components/AutomationActionResumeDeploymentInput.vue
pleek91, line 1: Copied directly from ui-library

src/automations/components/AutomationDeploymentCombobox.vue
pleek91, line 1: Copied directly from ui-library

src/automations/types/actions.ts
pleek91, line 1: There's a bunch of commented types here. I started with this file and copied over and enhanced the types for all the actions before realizing how much work needed to be done for each action. So opted to only move over the two actions but left these types here because I already did the work and they will be needed in future PRs.

src/automations/types/actions.ts
```diff
@@ -0,0 +1,234 @@
+import { Require } from '@/types/utilities'
+import { createTuple, isNullish, isRecord, isString } from '@/utilities'
+
+export const { values: automationActionTypes, isValue: isAutomationActionType } = createTuple([
```
collincchoy, line 4: 💯

collincchoy (approved):

PR #2267, opened 2024-04-01
Move pause/resume automation actions into oss

pleek91:
# Description
Moves all the necessary bits of the pause/resume automation logic into oss. This includes moving some base model and api models and services. These continue the pattern of having a distinction between an Automation and a CloudAutomation.


src/automations/types/automation.ts
pleek91, line 1: This base automation class omits actions and triggers which are then added in cloud's CloudAutomation which extends this. Because actions is a discriminating union we cannot just add actions here and "extend" the actions in cloud.

src/services/index.ts
pleek91, line 1: This file wasn't sorted like we generally do. Haven't noticed any problems when testing but if we run into any sort of "used before it was declared" errors this is probably why.

znicholasbrown (approved):
A couple of non-blocking questions

src/automations/types/automation.ts
collincchoy: dang. hm so then automations created in this project _could_ omit those properties that otherwise probably should be there? I'm thinking we'll need those in here at some point where something in here tries to access those properties but thinking mb this is just incremental?

pleek91: I'm not sure. My thought is that this project might not even have that logic at all and this just has the pieces that the components in this project need when reading. But this is incremental for sure. Likely other stuff will be added. Just trying to move as little as possible in each PR. Just enough to support the action being moved.

collincchoy (approved):

src/automations/components/AutomationActionPauseAutomationInput.vue
```diff
@@ -0,0 +1,32 @@
+<template>
+  <p-content class="automation-action-pause-automation">
```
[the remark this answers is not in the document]
pleek91: ah this is a good catch. I added `input` to all the action components when I moved them into ui-library so the component export didn't conflict with the type export. I'll come back though and update these classes.

src/automations/components/AutomationAutomationCombobox.vue
```diff
@@ -0,0 +1,31 @@
+<template>
+  <AutomationCombobox v-model:selected="internalSelected" empty-message="Infer Automation" class="automation-automation-combobox">
```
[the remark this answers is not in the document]
pleek91: Yeah this isn't great. But this matches the other components. For example there is a DeploymentCombobox and then there is an AutomationDeploymentCombobox.

PR #2281, opened 2024-04-04
Move send notification action from cloud

pleek91:
# Description
Moves all the pieces of the "Send a notification" action from cloud. Noteworthy changes

- The AutomationActionSendNotificationInput component no longer requires passing in the automation. It used to use the automation's trigger to determine the default subject and body. There is now a separate utility in nebula-ui that determines the default value for any action. That way actions can be completely agnostic of the rest of the automation. 
- The template validation endpoint was also moved into the WorkspaceAutomationsApi service. This endpoint has not been implemented in the oss api yet.

### DIFF
--- a/src/automations/components/AutomationActionSendNotificationInput.vue
+++ b/src/automations/components/AutomationActionSendNotificationInput.vue
@@ -1,0 +1,108 @@
+<template>
+  <p-content class="automation-action-send-notification">
+    <p-label label="Block" :state="blockDocumentState" :message="blockDocumentError">
+      <BlockCapabilityBlockDocumentInput v-model="blockDocumentId" :state="blockDocumentState" capability="notify" required use-modal />
+    </p-label>
+
+    <template v-if="isAnonymousBlock">
+      <p-message info>
+        This notification action is using an anonymous block. This is most likely because it was migrated from an existing notification.
+      </p-message>
+    </template>
+
+    <p-label label="Subject" :state="subjectState" :message="subjectError">
+      <p-code-input v-model="subject" :state="subjectState" lang="jinja" />
+    </p-label>
+
+    <p-label label="Body" :state="bodyState" :message="bodyError">
+      <p-code-input v-model="body" :state="bodyState" rows="5" lang="jinja" />
+    </p-label>
+
+    <p-message info>
+      Event fields and any of the following objects present in the triggering event can be used in notification templates: flow, deployment, flow run, work pool, work queue, metric.
+    </p-message>
+  </p-content>
+</template>
+
+<script lang="ts" setup>
+  import { useValidation, ValidationRule } from '@prefecthq/vue-compositions'
+  import { computed } from 'vue'
+  import { AutomationActionSendNotification } from '@/automations/types/actions'
+  import BlockCapabilityBlockDocumentInput from '@/components/BlockCapabilityBlockDocumentInput.vue'
+  import { useBlockDocument, useWorkspaceApi } from '@/compositions'
+  import { timeout } from '@/utilities/time'
+  import { isRequired } from '@/utilities/validation'
+
+  const props = defineProps<{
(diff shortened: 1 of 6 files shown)


src/automations/components/AutomationActionSendNotificationInput.vue
```diff
@@ -0,0 +1,108 @@
...
+
+    if (source === 'validator') {
+      await timeout(3000, signal)
+    }
+
+    if (signal.aborted) {
+      return
+    }
+
+    const valid = await api.automations.validateTemplate(value)
```
znicholasbrown, line 67: recognizing this doesn't exist in the public API yet, can you still add it to the client-side API class?

src/automations/components/AutomationActionSendNotificationInput.vue
```diff
@@ -0,0 +1,108 @@
+<template>
+  <p-content class="automation-action-send-notification">
+    <p-label label="Block" :state="blockDocumentState" :message="blockDocumentError">
+      <BlockCapabilityBlockDocumentInput v-model="blockDocumentId" :state="blockDocumentState" capability="notify" required use-modal />
+    </p-label>
+
+    <template v-if="isAnonymousBlock">
+      <p-message info>
+        This notification action is using an anonymous block. This is most likely because it was migrated from an existing notification.
```
znicholasbrown, line 9:
```suggestion
        This notification action is using an anonymous block, most likely because it was migrated from an existing notification.
```
Is there call to action here? I'm not sure what purpose this is serving otherwise

src/automations/components/AutomationActionSendNotificationInput.vue
```diff
@@ -0,0 +1,108 @@
...
+    <p-label label="Subject" :state="subjectState" :message="subjectError">
+      <p-code-input v-model="subject" :state="subjectState" lang="jinja" />
+    </p-label>
+
+    <p-label label="Body" :state="bodyState" :message="bodyError">
+      <p-code-input v-model="body" :state="bodyState" rows="5" lang="jinja" />
+    </p-label>
+
+    <p-message info>
+      Event fields and any of the following objects present in the triggering event can be used in notification templates: flow, deployment, flow run, work pool, work queue, metric.
```
znicholasbrown, line 22:
```suggestion
      In addition to any fields present on the triggering event, the following objects can be used in notification templates: <p-code inline>flow</p-code>, <p-code inline>deployment</p-code>, <p-code inline>flow_run</p-code>, <p-code inline>work_pool</p-code>, <p-code inline>work_queue</p-code>, and <p-code inline>metric</p-code>.
```

Check me on the specifics but I think this language is a little clearer and hopefully the code snippets provide a better jumping-off point.

znicholasbrown (approved):
A couple of small comments, nothing blocking since I understand this is a port.

src/automations/components/AutomationActionSendNotificationInput.vue
pleek91: Yes added in https://github.com/PrefectHQ/prefect-ui-library/pull/2281/commits/0bfa2f00ac731b79aed40c56b5b99c1d1b03036c

src/automations/components/AutomationActionSendNotificationInput.vue
pleek91: No not really. I'm trying to see if I can remove this. But I believe this is here because it'll show the block but its not one of the values in the list of regular blocks.

src/automations/components/AutomationActionSendNotificationInput.vue
pleek91: Sounds good to me.

src/automations/components/AutomationActionSendNotificationInput.vue
pleek91: You were super fast! ❤️

PR #2292, opened 2024-04-11
Chore: Map state timestamps from string => date

znicholasbrown:
This was inconsistent with other timestamps and meant we couldn't operate on them in components without doing a conversion there; this seemed better.


pleek91 (approved):

znicholasbrown:
I'm guessing this is going to fail tests, going to use those to figure out what this breaks instead of trying to go through all the components 🤷🏻

znicholasbrown:
Oh that was easier than i thought

PR #2295, opened 2024-04-12
Add a useWorkPoolId composition

zhen0:
On the events feed (and a few other places) we need to get a work pool by ID instead of name. This adds a composition to make that easier. 

Related to https://github.com/PrefectHQ/nebula-ui/pull/4509


znicholasbrown (approved):
Thanks @zhen0 very convenient

src/services/WorkspaceWorkPoolsApi.ts
```diff
@@ -27,6 +27,18 @@ export class WorkspaceWorkPoolsApi extends WorkspaceApi {
...
+    const filter = {
+      workPools: {
+        id: [id],
+      },
+    }
+    const request = mapper.map('WorkPoolsFilter', filter, 'WorkPoolsFilterRequest')
+    const { data } = await this.post<WorkPoolResponse[]>('/filter', request)
+
+    return mapper.map('WorkPoolResponse', data[0], 'WorkPool')
+  }
```
pleek91, line 40: I literally have a branch I was working on today where I wrote this exact same composition and service method! 😂 great minds!

I wrote this just slightly differently to account for deleted work pools. This way we'll get an error if the work pool doesn't exist 

```suggestion
  public async getWorkPoolById(workPoolId: string): Promise<WorkPool> {
    const workPools = await this.getWorkPools({ workPools: { id: [workPoolId] } })
    const workPool = workPools.find(workPool => workPool.id === workPoolId)

    if (!workPool) {
      throw new BatchLookupError(workPoolId)
    }

    return workPool
  }
```

PR #2301, opened 2024-04-12
Move automation routes from cloud

pleek91:
# Description
In order for components in ui-library to link to automation pages we need to move the automation routes into this project.

### DIFF
--- a/src/automations/types/createAutomationQuery.ts
+++ b/src/automations/types/createAutomationQuery.ts
@@ -1,0 +1,37 @@
+import { AutomationAction } from '@/automations/types/actions'
+import { isRecord } from '@/utilities/object'
+
+export type CreateEventAutomationQuery = {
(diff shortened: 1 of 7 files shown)


src/automations/types/createAutomationQuery.ts
```diff
@@ -0,0 +1,37 @@
+import { AutomationAction } from '@/automations/types/actions'
+import { isRecord } from '@/utilities/object'
+
+export type CreateEventAutomationQuery = {
+  from: 'event',
+  event: { id: string, occurred: Date },
```
pleek91, line 6: One thing to note. In cloud this is type `Event` (a workspace event). But I don't want to move all the stuff associated with that type yet. So inlining just the two properties used here so this type is more narrow than `Event`.

znicholasbrown (approved):

PR #2303, opened 2024-04-12
Chore: refreshChannel when we create a new work pool

dylanbhughes:
Ensures that our default work pool subscriptions are accurate.


dylanbhughes:
This change isn't really relevant to the open source, moving to nebula-ui

PR #2307, opened 2024-04-12
Chore: make the props simpler and remove the counter

dylanbhughes:
After working with @collincchoy to place the usage counter for automations, I've decided to move the counter for work pools in Nebula so that the placement is similar. The `PageHeadingWorkPools` still needs a prop to help it hide the create WorkPool button but it's now much simpler.

### DIFF
--- a/src/components/PageHeadingWorkPools.vue
+++ b/src/components/PageHeadingWorkPools.vue
@@ -20,8 +20,3 @@
   const routes = useWorkspaceRoutes()
 </script>
 
-<style>
-.page-heading-work-pools__total-counter { @apply
-  text-subdued
-}
-</style>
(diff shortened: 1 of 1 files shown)


collincchoy (approved):

src/components/PageHeadingWorkPools.vue
```diff
@@ -1,37 +1,23 @@
 <template>
   <page-heading class="page-heading-work-pools" :crumbs="crumbs">
     <template #after-crumbs>
-      <p-button v-if="can.create.work_pool && displayCreateButton" small icon="PlusIcon" :to="routes.workPoolCreate()" />
-    </template>
-    <template v-if="workPoolsLimit && currentWorkPools" #actions>
-      <span class="page-heading-work-pools__total-counter">
```
collincchoy, line 7: Can you remove this class from the `<style>` block below?

PR #2308, opened 2024-04-15
Add richer descriptions for automation actions

pleek91:
# Description
This PR attempts to make the description of an automations actions richer and more descriptive by adding the ability for each action to define a component that displays information about a specific instance of that action. 

This PR adds the pattern by enforcing each action type has a description component associated with it. Which could be as simple as choosing the "default" description component which just uses the action's label. Or a custom component that adds more rich information. 

Description components are added for all of the automation actions defined here in ui-library. And allows for extending in nebula-ui. 

image

image
image

image


collincchoy (approved):
Cool - this looks great!

PR #2320, opened 2024-04-17
Enhancement: Update modal buttons and add auto-close to the parameters and job variable modals

pleek91:
# Description
When viewing a run deployment action there are two modals that are basically the same for viewing parameters and job variables. Update them to use `auto-close` so users can click on the mask to close the modal and changing the button from the default cancel button to a primary "Close" button like we've done on the ExtraInfoModal.

### DIFF
--- a/src/automations/components/AutomationActionDescriptionRunDeployment.vue
+++ b/src/automations/components/AutomationActionDescriptionRunDeployment.vue
@@ -19,12 +19,22 @@
     </template>
   </div>
 
-  <p-modal v-model:show-modal="showParametersModal" title="Parameters">
+  <p-modal v-model:show-modal="showParametersModal" title="Parameters" auto-close>
     <p-code-highlight :text="stringify(action.parameters)" lang="json" />
+    <template #cancel>
+      <p-button @click="closeParametersModal">
+        Close
+      </p-button>
+    </template>
   </p-modal>
 
-  <p-modal v-model:show-modal="showJobVariablesModal" title="Job Variables">
+  <p-modal v-model:show-modal="showJobVariablesModal" title="Job Variables" auto-close>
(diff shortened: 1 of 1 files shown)


collincchoy (approved):

src/automations/components/AutomationActionDescriptionRunDeployment.vue
```diff
@@ -42,8 +52,8 @@
     action: AutomationActionRunDeployment,
   }>()
 
-  const { value: showParametersModal, setTrue: openParametersModal } = useBoolean()
-  const { value: showJobVariablesModal, setTrue: openJobVariablesModal } = useBoolean()
+  const { value: showParametersModal, setTrue: openParametersModal, setFalse: closeParametersModal } = useBoolean()
```
collincchoy, line 55: huh. i haven't thought of using this over `useShowModal` but i suppose they're the same.. 

seems confusing. y'know from a "one way of doing things" sort of way

pleek91: oh dang, I forgot that existed 😂 I should just use that. I feel like `useShowModal` existed and then `useBoolean` came around because we wanted something more generic. Should I change this or does it not really matter?

pleek91: I updated it cause now that you mentioned it it started bothering me.

collincchoy: i'm on the fence between the 2, but would err toward consistency and least surprise / most-familiar.

I _think_ that'd be useShowModal to me. reading `useBoolean` in a vacuum is surprisingly not immediately intuitive to me what it does, but i get how we prob got here lol. def nbd tho - 🤷

collincchoy (approved):

PR #2321, opened 2024-04-18
Enhancement: flow run list information density

znicholasbrown:
This PR introduces a few new props to the flow run list item component set to allow hiding the flow name and ancillary details (like work pool, deployment, work queue)

cc @cicdw

### DIFF
--- a/src/components/FlowRunFilteredList.vue
+++ b/src/components/FlowRunFilteredList.vue
@@ -19,7 +19,7 @@
       </template>
     </p-list-header>
 
-    <FlowRunList v-model:selected="selected" :flow-runs="flowRuns" :selectable="selectable && can.delete.flow_run" @bottom="next" />
+    <FlowRunList v-model:selected="selected" v-bind="{ hideDetails, hideFlowName }" :flow-runs="flowRuns" :selectable="selectable && can.delete.flow_run" @bottom="next" />
 
     <PEmptyResults v-if="empty">
       <template #message>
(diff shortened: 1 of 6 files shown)


src/components/FlowRunFilteredList.vue
```diff
@@ -19,7 +19,7 @@
       </template>
     </p-list-header>
 
-    <FlowRunList v-model:selected="selected" :flow-runs="flowRuns" :selectable="selectable && can.delete.flow_run" @bottom="next" />
+    <FlowRunList v-model:selected="selected" v-bind="{ hideDetails, hideFlowName }" :flow-runs="flowRuns" :selectable="selectable && can.delete.flow_run" @bottom="next" />
```
znicholasbrown, line 22: This component isn't being used on the dashboard but is used on the flow page where we can remove the flow name since that's a known context

znicholasbrown:
Before:
Screenshot 2024-04-18 at 12 08 56 PM


After:
Screenshot 2024-04-18 at 12 08 54 PM

cicdw (approved):
nice!! 😍

src/components/FlowRunFilteredList.vue
pleek91: Its being used in quite a few places. But the defaults are false so we have to intentionally hide things. So this seems fine to me

pleek91 (approved):
Nice, great low hanging fruit

src/components/FlowRunFilteredList.vue
znicholasbrown: Yeah that was the intent - this won't change anything but the dash by default

PR #2323, opened 2024-04-19
Fix schema form properties with no default value from defaulting to `null` (`None`)

pleek91:
# Description
For most form components `null | undefined` is treated the same. But for something like a json input `null` is actually a valid value. So we need to make sure we always use `undefined` for an empty value and not null. 

Also noticed that definitions were not getting mapped so added the mapper to those properties.

### DIFF
--- a/src/schemas/maps/schema.ts
+++ b/src/schemas/maps/schema.ts
@@ -6,7 +6,7 @@ import { mapValues } from '@/utilities/object'
 
 export const mapSchemaResponseToSchema: MapFunction<SchemaResponse, Schema> = function(source) {
   return {
-    definitions: source.definitions,
+    definitions: this.map('SchemaPropertiesResponse', source.definitions, 'SchemaProperties'),
     position: source.position,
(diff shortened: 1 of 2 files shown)


src/schemas/maps/schema.ts
```diff
@@ -6,7 +6,7 @@ import { mapValues } from '@/utilities/object'
 
 export const mapSchemaResponseToSchema: MapFunction<SchemaResponse, Schema> = function(source) {
   return {
-    definitions: source.definitions,
+    definitions: this.map('SchemaPropertiesResponse', source.definitions, 'SchemaProperties'),
```
znicholasbrown, line 9: oh i'm surprised we weren't already doing this

znicholasbrown (approved):

src/schemas/maps/schema.ts
pleek91: Same, but this really only matters for block definitions which we're not really using.

PR #2328, opened 2024-04-24
Prune dangerous trash icons. 

aaazzam:
In prefect-ui-library we accidentally use destructive (i.e. red) colors for non-destructive actions. 

e.g. we use a destructive trash icon button to launch a modal, and separately that modal has a button to confirm delete. 

- the button to launch the modal is _not_ a destructive action - so it shouldn't be red (changed).
- the button to confirm deletion _is_ a destructive action, so it should continue to be red (unchanged).

separately this choice pays dividends in our design library so that we don't need to worry about outline&destructive vs filled&destructive buttons


pleek91 (approved):
yep yep yep

PR #2329, opened 2024-04-24
Feature: `FormattedDate` component

znicholasbrown:
This PR adds a new component, `RichTimestamp`, which we should use anywhere we display a normalized/formatted timestamp to ensure the long form version of the timestamp is always accessible.


src/components/RichTimestamp.vue
```diff
@@ -0,0 +1,65 @@
...
+        case 'numeric':
+          return formatDateTimeNumeric(new Date(props.timestamp))
+        case 'approximate':
+        default:
+          return secondsToApproximateString(props.timestamp)
+      }
+    }
+
+    return props.timestamp
+  })
```
znicholasbrown, line 51: This probably isn't perfect, open to alternate suggestions

pleek91 (approved):
Personally I find "RichTimestamp" a little nondescript. But the only suggestion I have is `DateFormatted` or something similar. Approved as is though

pleek91 (approved):

znicholasbrown:
Discussed sync and made some minor updates to what this supports, thanks @pleek91 !!

PR #2334, opened 2024-04-25
Fix generated deployment parameter not respecting `None` as a default value

pleek91:
# Description
Found a couple places where we were checking for nullish values when we should be only checking for undefined. This would mean that `None` values were not treated as real values.

### DIFF
--- a/src/schemas/components/SchemaFormProperties.vue
+++ b/src/schemas/components/SchemaFormProperties.vue
@@ -79,7 +79,7 @@
   }
 
   function getValue(propertyKey: string): unknown {
-    return props.values?.[propertyKey] ?? undefined
+    return props.values?.[propertyKey]
   }
 
   function setValue(propertyKey: string, value: unknown): void {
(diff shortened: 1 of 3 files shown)


src/schemas/components/SchemaFormProperties.vue
```diff
@@ -79,7 +79,7 @@
   }
 
   function getValue(propertyKey: string): unknown {
-    return props.values?.[propertyKey] ?? undefined
+    return props.values?.[propertyKey]
```
pleek91, line 82: Since `null` (none) is a valid value we shouldn't coerce it to undefined.

src/schemas/components/SchemaFormPropertyAnyOf.vue
```diff
@@ -96,12 +96,14 @@
 
   const property = computed(() => {
     const selectedProperty = props.property.anyOf[selectedPropertyIndex.value]
+    // eslint-disable-next-line no-unused-vars
+    const { anyOf, ...property } = props.property
```
pleek91, line 100: unrelated change that strips the `anyOf` from the schema for an individual `anyOf` property. Wasn't hurting anything currently but would cause issues with validating a single property.

znicholasbrown (approved):

PR #2339, opened 2024-04-25
Enhancement: Deployments list

znicholasbrown:
This PR introduces a new component called `DeploymentList`, which is a non-breaking redesign of the existing `DeploymentsList` component. The existing component's CSS classes have been modified because they were inconsistent with the component name (and inconsistent internally) and would have conflicted with those of the new component. 

In service of the new component, this PR includes some minor visual updates, API extensions, and component extensions, none of which are breaking. 
- Updated the styles of deployment status badges for visual clarity and added a small pulse animation to ready statuses:

| Before    | After |
| -------- | ------- |
| Screenshot 2024-04-25 at 7 12 03 PM  | [Screenshot 2024-04-25 at 7 12 14 PM ](https://github.com/PrefectHQ/prefect-ui-library/assets/27291717/b9608525-d517-44f5-8a3b-a7467f431d80)   |
| Screenshot 2024-04-25 at 7 22 37 PM |  Screenshot 2024-04-25 at 7 22 03 PM |

- Introduces a new `FlowPopover` component - I've stripped the functionality from it for now while we iterate on content / potentially replace the underlying mechanism but the bones are good
- Updated the font face for `ResultsCount` and `SelectedCount` to better reflect brand styles:

| Before    | After |
| -------- | ------- |
| Screenshot 2024-04-25 at 7 18 05 PM |  Screenshot 2024-04-25 at 7 18 08 PM  |

- Updated `SearchInput` to allow passing the new `small` variant prop
- Added new and existing but missing filter deployment filter criteria to filters models
- Enabled filtering deployments lists on the name of the deployment OR the name of its parent flow
- Added a new route param handler for deployment statuses (we're not using these filters anywhere that I'm aware of yet but will soon)
- Improved mobile-scaling of table items, including fixing an issue that required refreshing the page to show/hide columns appropriately
- General UX improvements

Here's some screenshots (some table improvements are included that have already been released):

Before:

Screenshot 2024-04-25 at 7 28 12 PM

After:

Screenshot 2024-04-25 at 7 28 16 PM

Before (mobile):

Screenshot 2024-04-25 at 7 28 28 PM

After (mobile):

Screenshot 2024-04-25 at 7 28 25 PM


src/components/DeploymentStatusBadge.vue
```diff
@@ -1,18 +1,42 @@
 <template>
-  <template v-if="deployment.status">
-    <p-tag>
-      <DeploymentStatusIcon :status="deployment.status" />
-      {{ titleCase(deployment.status) }}
-    </p-tag>
-  </template>
+  <p-tag v-if="deployment.status" :class="classes.root" class="deployment-status-badge">
```
znicholasbrown, line 2: Removed the template wrapper around the root element here to avoid double-nesting `templates` unecessarily

src/components/DeploymentsList.vue
```diff
@@ -1,5 +1,5 @@
 <template>
```
znicholasbrown, line 1: Changes to this component are purely for consistency / to avoid CSS conflicts.

src/components/FlowRunPopOver.vue
```diff
@@ -78,6 +78,7 @@
   rounded-full
   overflow-hidden
   cursor-pointer
+  z-10
```
znicholasbrown, line 81: This was ending up behind certain content when it shouldn't have been.

src/components/DeploymentList.vue
```diff
@@ -0,0 +1,352 @@
...
+        <span />
+      </template>
+
+      <template #deployment="{ row }">
+        <div class="deployment-list__deployment">
+          <router-link :to="routes.deployment(row.id)" class="deployment-list__name">
+            {{ row.name }}
+          </router-link>
+
+          <FlowPopover :flow-id="row.flowId" class="deployment-list__flow-name" />
```
pleek91, line 54: nit: some of these elements have classes but no styles. Probably don't need them.

src/components/DeploymentList.vue
```diff
@@ -0,0 +1,352 @@
...
+      <template #updated="{ row }">
+        <FormattedDate :date="row.updated" class="deployment-list__timestamp" />
+      </template>
+
+      <template #created="{ row }">
+        <FormattedDate :date="row.created" class="deployment-list__timestamp" />
+      </template>
+
+      <template #schedules="{ row }">
+        <div class="deployment-list__schedules">
```
pleek91, line 67: This would make a nice little component.

src/components/DeploymentList.vue
```diff
@@ -0,0 +1,352 @@
...
+    },
+  ])
+
+  const columnClasses = (column: TableColumn<Deployment>): ClassValue => [`deployment-list__${snakeCase(column.label)}-column`]
+  const rowClasses = (row: Deployment): ClassValue => {
+    return [
+      {
+        'deployment-list__row--subdued': row.paused,
+      },
+    ]
```
pleek91, line 227:
```suggestion
    return {
      'deployment-list__row--subdued': row.paused,
    }
```

src/components/DeploymentList.vue
```diff
@@ -0,0 +1,352 @@
...
+
+  const selectedDeployments = ref<Deployment[]>([])
+
+  function refresh(): void {
+    subscriptions.refresh()
+  }
+
+  const emit = defineEmits<{
+    (event: 'delete'): void,
+  }>()
```
pleek91, line 238: nit: Nice to have all the "define" macros at the top of the script so you can see the component's api

src/components/DeploymentStatusBadge.vue
pleek91: I wish you could just add v-if on the root template tag

src/components/FlowPopover.vue
```diff
@@ -0,0 +1,74 @@
+<template>
+  <p-pop-over :placement="placement">
+    <template #target="{ open, close, toggle }">
+      <!-- TODO: This doesn't actually open/close right now -->
+      <slot v-bind="{ open, close, toggle, flow }">
+        <FlowRouterLink :flow-id="flowId">
+          <div v-if="flow" class="flow-popover__link" v-bind="$attrs">
+            <p-icon icon="PFlow" small class="flow-popover__link-icon" />
+            <span class="flow-popover__link-name">{{ flow.name }}</span>
+          </div>
```
pleek91, line 10: Is this not the same as FlowIconText?

src/components/SearchInput.vue
```diff
@@ -3,7 +3,7 @@
     <template v-if="label">
       <p-label for="search" :label="label" class="search__label" />
     </template>
-    <p-text-input v-model="internalValue" type="search" :placeholder="placeholder" class="search__input">
+    <p-text-input v-model="internalValue" type="search" :small="small" :placeholder="placeholder" class="search__input">
```
pleek91, line 6: This is where you can use the new shorthand! Not a real suggestion, just an example
```suggestion
    <p-text-input v-model="internalValue" type="search" :small :placeholder class="search__input">
```

src/components/SearchInput.vue
```diff
@@ -19,6 +19,7 @@
     modelValue: string | null | undefined,
     placeholder?: string,
     label?: string,
+    small?: boolean,
```
pleek91, line 22: Should this be a `size` prop to be consistent with the direction adam is going with prefect-design?

src/formatters/DeploymentStatusRouteParam.ts
```diff
@@ -0,0 +1,17 @@
...
+  protected override parse(value: LocationQueryValue): DeploymentStatus {
+    if (value === null || !isDeploymentStatus(value)) {
+      throw new InvalidRouteParamValue()
+    }
+
+    return value
+  }
+
+  protected override format(value: DeploymentStatus): LocationQueryValue {
+    return `${value}`
```
pleek91, line 15: Isn't this the same? 
```suggestion
    return value
```

src/maps/filters.ts
```diff
@@ -235,6 +235,10 @@ export const mapDeploymentFilter: MapFunction<DeploymentFilter, DeploymentFilter
     },
     is_schedule_active: toEquals(source.isScheduleActive),
     work_queue_name: toAny(source.workQueueName),
+    flow_or_deployment_name: toLike(source.flowOrDeploymentNameLike),
```
pleek91, line 238: oh nice, does this mean we can fuzzy search across both?!

pleek91 (approved):
Overall LGTM! Comments are non blocking

src/components/DeploymentList.vue
znicholasbrown: Agreed, done!

src/components/DeploymentList.vue
znicholasbrown: Ah I just realized the type for this is wrong as defined in ui-lib, it's requiring this to be an array of strings or records. Let me see if that's fixable.

znicholasbrown: Ok I just copied what we had in prefect-design for now which makes a lot more sense

src/formatters/DeploymentStatusRouteParam.ts
znicholasbrown: lol yes, nice catch

src/components/DeploymentList.vue
pleek91: ah right. Hmm I feel like this is more accurate if it still works
```
export type ClassValue = string | string[] | Record<string, boolean> | (string | Record<string, boolean>)[]
```

src/maps/filters.ts
znicholasbrown: yes!

src/components/DeploymentList.vue
pleek91: ah yeah that one makes more sense too. Should probably delete the type from ui-library.

znicholasbrown: Yeah that works for this component at least, let's see how the rest of the build behaves

src/components/SearchInput.vue
znicholasbrown: Ah yeah, that'll require a small shim for `small` in the meantime but I'm ok with that.

pleek91: not saying you should or have to use it, just pointing it out since you noticed it in my PR yesterday

src/components/SearchInput.vue
znicholasbrown: Ah yeah, that'll require a small shim for small in the meantime but I'm ok with that.

src/components/SearchInput.vue
znicholasbrown: Ah sorry that comment got added to the wrong thread - love the suggestion, applying it!

PR #2342, opened 2024-04-29
Sync value between form and json when entering flow parameters

### LINKED ISSUE #12564: Bug: Override parameters not used in JSON view
### First check

### Bug summary

When creating a new flow run via "Copy to new run", parameters from the previous run are preserved in the parameter input form.
However, when switching to the JSON input, they are reset to defaults.

### Reproduction

```
from prefect import flow

@flow
def main(
        param: str = "foo",
):
    pass


if __name__ == "__main__":
    main.serve(
        name='default',
    )
 ```
 
- Run the above script (and so create a new deployment)
- Navigate to the deployment page, click "Run" -> "Custom run"
- Change the param value (e.g. to "bar")
- Click run and navigate to the newly created run
- From the top-right context menu on the run page, click "Copy to new run"
- See changed value in input form (expected)
- Click on three-dot button at the top and switch to "Use JSON input"
- See original value "foo"

Same applies if parameters are passed to the deployment

```
from prefect import flow

@flow
def main(
        param: str = "foo",
):
    pass


if __name__ == "__main__":
    main.serve(
        name='default',
        parameters={"param": "bar"}
    )
 ```

When creating a custom run for this flow, the input form shows "bar" but the JSON input shows "foo".

### Error

image
image

### Browsers

- [X] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### Prefect version

```Text
Version:             2.16.8
API version:         0.8.4
Python version:      3.10.12
Git commit:          11cb641c
Built:               Fri, Mar 29, 2024 11:01 AM
Profile:             default
Server type:         ephemeral
Server:
  Database:          sqlite
  SQLite version:    3.37.2
```

### Additional context

possibly related to #12106

### LINKED ISSUE #12566: Custom values of optional parameters not used in input form
### First check

### Bug summary

When assigning a Pydantic model with custom values as default to a flow parameter, the default values of the model are displayed in the flow submission form instead of the custom values.

When switching to "JSON input", the correct custom values are displayed.

This is only happening with Pydantic 2 and optional parameters. With Pydantic 1 or non-optional parameters, everything works as expected.

### Reproduction

```
from prefect import flow
from pydantic import BaseModel, Field
from typing import Optional


class MyModel(BaseModel):
    param1: Optional[str] = Field(title="Param1", default="foo")


@flow
def main(
        param: MyModel = MyModel(param1="bar")
):
    pass


if __name__ == "__main__":
    main.serve(
        name='default',
    )
  ```
  
- Run the above script (and so create a new deployment)
- Navigate to the deployment page, click "Run" -> "Custom run"
- See "foo" (the model default) in input form
- Click on three-dot button and switch to "JSON input"
- See "bar" (the custom assigned default)

### Error

image
image

### Browsers

- [X] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### Prefect version

```Text
Version:             2.16.8
API version:         0.8.4
Python version:      3.10.12
Git commit:          11cb641c
Built:               Fri, Mar 29, 2024 11:01 AM
Profile:             default
Server type:         ephemeral
Server:
  Database:          sqlite
  SQLite version:    3.37.2
```

### Additional context

possibly related to https://github.com/PrefectHQ/prefect/issues/12106 and #12564

pleek91:
# Description
Previously the form and json input modes in enhanced deployment parameters were treated like two separate forms. Switching between them would cause values to get out of sync and sometimes work to be lost. 

This enhancement updates the two modes to work like a single form working on a single set of values. So switching to json will cause any values entered/visible on the form to be stringified. And when switching to the form after editing the json the value will be parsed and the form will show the correct values. 

Closes https://github.com/PrefectHQ/prefect/issues/12566
Closes https://github.com/PrefectHQ/prefect/issues/12564

Note: When editing as json the form will require the input to be valid according to the schema before parsing and switching to the form.  

Using flow
```
from prefect import flow
from pydantic import BaseModel, Field
from typing import Optional


class MyModel(BaseModel):
    param1: Optional[str] = Field(title="Param1", default="from model")
    param2: Optional[str] = Field(title="Param2", default="from model")
    param3: Optional[str] = Field(title="Param3", default="from model")


@flow
def main(
        param: MyModel
):
    pass


if __name__ == "__main__":
    main.serve(
        name='sample_flow',
        parameters={"param": { "param3": "from deployment" }}
    )
```

Initial form
image

Values updated
image

Switch to json
image

Update json with invalid value and attempt to switch back to the form
image

Fix json to be valid
image

Switch to form
image


src/schemas/maps/schemaValue.ts
```diff
@@ -0,0 +1,99 @@
...
+    default:
+      throw new Error(`mapSchemaValueJson missing case for kind: ${to}`)
+  }
+}
+
+function mapSchemaValueJson(json: PrefectKindJson, to: PrefectKind): SchemaValue {
+  switch (to) {
+    case 'jinja':
+      return json.value
+    case 'workspace_variable':
```
pleek91, line 47: this is not correct (note to self)

src/schemas/maps/schemaValue.ts
```diff
@@ -0,0 +1,99 @@
...
+      throw new Error(`mapSchemaValueJson missing case for kind: ${to}`)
+  }
+}
+
+function mapSchemaValueNone(none: unknown, to: PrefectKind): SchemaValue {
+  const kind = getPrefectKindFromValue(none)
+
+  if (kind !== 'none') {
+    throw new Error(`mapSchemaValueNone called with value of kind: ${kind}`)
+  }
```
pleek91, line 67: I think this isn't needed any more

src/schemas/compositions/usePrefectKindValue.ts
```diff
@@ -0,0 +1,85 @@
...
+
+      if (!valid) {
+        return
+      }
+    }
+
+    // store the current value for the current kind
+    const currentKind = getPrefectKindFromValue(schemaValue.value)
+
+    valueMap[currentKind] = structuredClone(getRawValue(schemaValue.value))
```
pleek91, line 44: try catch this

pleek91: add a comment about why we're using both structuredClone and getRawValue

src/schemas/compositions/usePrefectKindValue.ts
```diff
@@ -0,0 +1,95 @@
...
+        if (to === 'none') {
+          schemaValue.value = valueMap[to]
+          return
+        }
+
+        schemaValue.value = valueMap[to] ?? { __prefect_kind: to }
+      }
+
+      throw error
+    }
```
pleek91, line 65: @znicholasbrown

src/schemas/maps/schemaValue.ts
pleek91, line 1: @znicholasbrown this changed to throw errors when a value cannot be mapped. Then the `usePrefectKindValue` composition catches that and handles it accordingly.

znicholasbrown (approved):

PR #2348, opened 2024-05-01
Move deployment status trigger template from cloud

pleek91:
# Description
First PR of moving ui components from cloud for triggers that are supported in oss. Everything base trigger related and related to the deployment status trigger. 

Sets up some initial abstraction into AutomationTriggerEventInput.vue which will be the component cloud and oss will implement for their respective forms.


src/automations/components/AutomationWithinInput.vue
```diff
@@ -0,0 +1,77 @@
...
+        return true
+      }
+
+      if (value <= max) {
+        return true
+      }
+
+      return 'Time cannot be greater than 30 days'
+    }
+  }
```
pleek91, line 76: These utilities did not exist in ui-library. Rather than starting down the rabbit hole of migrating validation utilities (which conflict with vee-validate rules) I chose to just past them into the component.

src/automations/types/deploymentStatusTrigger.ts
```diff
@@ -0,0 +1,27 @@
...
+  posture: AutomationTriggerEventPosture,
+  status: DeploymentStatus,
+  time: number,
+}
+
+/**
+ * This is used to determine if an AutomationTrigger can be mapped into a DeploymentStatusTrigger
+ */
+export function isDeploymentStatusTrigger(trigger: AutomationTrigger): boolean {
+  return isAutomationTriggerEvent(trigger) &&
```
pleek91, line 21: Calling out I change line 21. It used to check if the posture was `Proactive` or `Reactive`. But posture is unique to event triggers now, so just checking that its an event trigger.

znicholasbrown (approved):

PR #2349, opened 2024-05-01
Remove default open accordion in Dashboard

aaazzam:
The dashboard uses a PAccordion component to collapse groups of flow runs.

Today's behavior is to render the first accorrdion item open by default, which can cause stuttery load times since it fetches all of the subsequent flow runs underneath it. Relatedly, this is the only component which would break under the new PAccordion API so I'm going to remove this behavior for now.


znicholasbrown (approved):

znicholasbrown:
This is a deliberate tradeoff of load speed vs visibility - since the flows still show up as expected I think this is good.

pleek91:
Nice, I like this

PR #2374, opened 2024-05-03
Chore: Tailwind config preset and TS migration

znicholasbrown:
While a chore, this PR includes some notable changes:
1. The package now expliticly exports all files instead of just those in `/dist` - this is a convenience for downstream consumers to allow better file discovery/not limit to exports in `/dist`
2. Postcss config has been migrated to the modern `.cjs` syntax
3. Tailwind plugin has been removed in favor of `presets` syntax (a plugin still exists for adding utilities but this doesn't need to be consumed by downstream applications anymore)
4. Tailwind config has been migrated to ES modules
5. Standardizes tsconfig with other packages in our ecosystem; a future enhancement would be to centralize and extend this config but that felt out of scope of this PR.
6. Renames `style.css` => `index.css`; this is consistent with other packages in the ecosystem and is more in line with our use of the file as an import layer
7. Vite config has been migrated to `.mts` syntax, consistent with prefect-design


package.json
```diff
@@ -31,11 +31,10 @@
         ".": {
             "import": "./dist/prefect-ui-library.mjs",
             "require": "./dist/prefect-ui-library.umd.js"
-        },
-        "./dist/style.css": "./dist/style.css"
```
znicholasbrown, line 35: Downstream applications will need to update references to this file (now available via `src/styles`

src/components/FlowRunsAccordion.vue
```diff
@@ -1,7 +1,7 @@
 <template>
   <template v-if="flowIds">
     <p-accordion :sections="flowIds" class="flow-runs-accordion">
-      <template #header="{ section: flowId, id, toggle, content, selected }">
+      <template #heading="{ section: flowId, id, toggle, content, selected }">
```
znicholasbrown, line 4: This was failing the build

src/components/NotificationForm.vue
```diff
@@ -14,6 +14,7 @@
       </p-label>
     </div>
 
+    <!-- @vue-ignore -->
     <p-button-group v-model="selectedBlockTypeId" :options="buttonGroup" />
```
znicholasbrown, line 18: This generic changed recently so this is a shim

.eslintrc.json
```diff
@@ -36,7 +36,7 @@
     }
   },
   "rules": {
-    "import/no-default-export": "error",
+    "import/no-default-export": "off",
```
pleek91, line 39: eh I'm pretty in favor of not using default exports. What's the thought here?

package.json
pleek91: This doesn't make sense to me. The css is still part of the build and generally needs to be designated as an export in the package.json or things like ts will error when trying to import it. At least that's how I understand it.

src/tailwind.config.ts
```diff
@@ -39,9 +33,4 @@ const config = {
...
     },
   },
-  plugins: [prefectDesignPlugin],
-}
-
-function PrefectUiLibraryPlugins() { /* noop */ }
-
-module.exports = plugin(PrefectUiLibraryPlugins, config)
\ No newline at end of file
+} satisfies Config
```
pleek91, line 36: this is a great way to use satisfies. Gonna log that away in my brain for future use.

vite.config.mts
pleek91, line 1: I think using `mts` should be a temporary solution where the long term solution is moving to `module` for the tsconfig

.eslintrc.json
znicholasbrown: Just that we were ignoring it in a lot of the files that I was working on and it doesn't seem necessary - why prevent default exports? I'm fine to revert this tbh I don't feel very strongly

package.json
znicholasbrown: The change to the `files` entry will address this by making both the source code available (via `/src/styles`) and the build artifacts available (via `/dist`). Retaining this mapping doesn't make a lot of sense to me because all it was saying was you could find a file at `/dist/style.css`, which is still true but there are a lot of other files there too. The import/require mappings are still correct because they're signals to node about how to handle the different import styles.

pleek91 (approved):

PR #2380, opened 2024-05-06
Move custom event trigger ui over from cloud and related event comboboxes

pleek91:


collincchoy (approved):

src/components/EventResourceCombobox.vue
```diff
@@ -0,0 +1,100 @@
...
+<script lang="ts" setup>
+  import { useSubscription } from '@prefecthq/vue-compositions'
+  import { computed } from 'vue'
+  import { useWorkspaceApi } from '@/compositions/useWorkspaceApi'
+
+  const selected = defineModel<string | string[] | null | undefined>('selected', { required: true })
+
+  const props = defineProps<{
+    multiple?: boolean,
+    additionalOptions?: RelatedResourceGroup[],
```
pleek91, line 21: oss doesn't have webhooks. So allowing for additional options to be passed in at implementation. This is so that cloud can define its own version of this component and provide it.

PR #2403, opened 2024-05-14
enhancement: max active runs and and catchup toggle to schedule UI

WillRaphaelson:
Adds catchup and max scheduled runs to the schedule edit / create modal.

image

### DIFF
--- a/src/components/DeploymentScheduleCard.vue
+++ b/src/components/DeploymentScheduleCard.vue
@@ -49,8 +49,7 @@
 
 
   const updateInternalState = (): void => {
-    internalValue.value = { active: props.deploymentSchedule.active, schedule: props.deploymentSchedule.schedule, jobVariables: props.deploymentSchedule.jobVariables }
-    // internalValue.value = { active: props.deploymentSchedule.active, schedule: props.deploymentSchedule.schedule, jobVariables: props.deploymentSchedule.jobVariables, catchup: props.deploymentScheduleCompatible.catchup, maxActiveRuns: props.deploymentScheduleCompatible.maxActiveRuns }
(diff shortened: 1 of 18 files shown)


WillRaphaelson:
closes [#13205](about:blank)

abrookins:
How do you clear max active runs / return to the default?

src/components/ScheduleFormModal.vue
```diff
@@ -23,6 +23,26 @@
...
     <FlowRunJobVariableOverridesLabeledInput v-if="can.access.deploymentScheduleFlowRunInfraOverrides" v-model="internalJobVariables" />
 
+    <p-label
+      label="Max active runs"
+      :message="maxActiveRunsErrorMessage"
+      :state="maxActiveRunsState"
+      class="interval-schedule-form__column--span-3"
+    >
+      <p-number-input v-model="internalMaxActive" min="0" :state="maxActiveRunsState" />
+    </p-label>
```
znicholasbrown, line 33: I don't think there's a way to nullify this value rn - any schedule created will have a value for this, defaulting to 0. I'd recommend we take page from the schemas work and provide a way to omit this value.

src/components/DeploymentScheduleCard.vue
```diff
@@ -28,13 +28,14 @@
   const props = defineProps<{
     deployment: Deployment,
     deploymentSchedule: DeploymentSchedule,
+    deploymentScheduleCompatible: DeploymentScheduleCompatible,
```
znicholasbrown, line 31: I'm not sure I understand what this prop is doing - the `DeploymentSchedule` has the `catchup` and `maxActiveRuns` values, right? Is there a reason we can't pull them from that prop? I may be misunderstanding how these are used.

znicholasbrown (commented):
A couple of questions @WillRaphaelson, this looks p good!

abrookins:
The catchup field probably needs a better label, even with the tool tip. It's a little inscrutable. Maybe "Catch up late runs?"

PR #2406, opened 2024-05-15
Bugfix: `useFlowRuns`

znicholasbrown:
This PR updates the `useFlowRuns` composition to optimize network requests. Previously, `useFlowRuns` called both the `/count` and `/filter` endpoints, even if the count was not used by the caller. This doubled the number of network requests per use, which was often unnecessary.

To address this I've introduced a new method, `usePaginatedFlowRuns`, for cases where pagination is needed. References have been updated to use either `useFlowRuns` or `usePaginatedFlowRuns` as appropriate.

This change reduces unnecessary calls to the `/count` endpoint, improving performance without affecting functionality.

Note: I've got some downstream PRs ready for this - they'll be much smaller


pleek91:
I'm curious what implementation(s) of this that weren't using the count were problematic. This is something we'd run into with all implementations of usePagination. But we generally use both the count and the list together.

pleek91 (approved):

znicholasbrown:
Sorry, I should have detailed that here; two of the biggest offenders were `useNextFlowRun` and `useLastFlowRun` in terms of requests volume - we don't use these compositions a ton so one potential solution would be to change their use. However, I decided to got his route because we have 5 other components that use `useFlowRuns` directly without consuming the count or any of the underlying pagination architecture:
- `FlowRunMenu`
- `FlowRunParentFlowRun`
- `FlowRunTagsInput`
- `WorkPoolQueueUpcomingFlowRunsList`

And I'd expect this number to increase as we swap out direct calls to the API (which is something we've been working toward)

PR #2415, opened 2024-05-16
Feature: Autonomous task runs

znicholasbrown:
This PR updates some of the library internals to provide the mechanics to show task runs on what is currently the "Flow Runs" page. It also includes some QOL updates around existing components/compositions, including an update to `useTaskRuns` to match that of #2406 (including a new paginated task runs composition)


src/compositions/usePagination.ts
```diff
@@ -76,7 +77,7 @@ export function usePagination<
   const pages = computed(() => Math.ceil(total.value / getLimit()))
 
   const countSubscriptionParameters = computed(() => {
-    if (page.value) {
+    if (isDefined(page.value)) {
```
znicholasbrown, line 80: I could use a gut check on this one - I don't quite understand what's up here but it seems like this composition has a weird catch-22 where infinite mode sets the page to 0, which causes this parameters computed to exit early and so the fetch functions never run. I'm not sure why we've not seen this before, I suspect it's something of a race condition?

aaazzam (approved):

src/compositions/usePagination.ts
collincchoy: 🧐 infinite mode should coalesce the page getter to 0 which would have this return null to disable the count subscription. 

Hmmm maybe open a separate issue to track and not block this stuff?

src/components/TaskRunListItem.vue
```diff
@@ -1,43 +1,67 @@
...
+          <template v-if="taskRun.startTime">
+            {{ formatDateTimeNumeric(taskRun.startTime) }}
+          </template>
+          <template v-else-if="taskRun.expectedStartTime">
+            {{ formatDateTimeNumeric(taskRun.expectedStartTime) }}
+          </template>
+        </p-icon-text>
+      </template>
+
+      <template v-if="showFlowRun && visible" #relationships>
```
collincchoy, line 40: can u eli5 what the intersection observer and visible toggle here are solving? 

is this like - a list of task run list items are rendered, some might be offscreen b/c scroll and for those we want to lazily load this slot?

collincchoy (approved):

src/components/TaskRunListItem.vue
znicholasbrown: Yup exactly - these make a bunch of secondary calls to grab related items based on the ids/names passed here; we don't want to both making those calls if the user can't even see the list item.

src/compositions/usePagination.ts
znicholasbrown: I'm gonna dig a little more - without this change I couldn't get any of my filter calls to run until I updated one of the filters.

PR #2417, opened 2024-05-16
Fix: Remove old flow run infra override access checks

zhen0:
https://github.com/PrefectHQ/prefect/pull/12742 removed experimental flags for flow run infra overrides.  However the UI still contains access checks meaning that users in OSS can not access or view the flow run infra overrides section of the custom run form (and elsewhere). 

This PR removes the access checks. Accompanying PR in https://github.com/PrefectHQ/prefect/pull/13401

### DIFF
--- a/src/components/FlowRunCreateFormV2.vue
+++ b/src/components/FlowRunCreateFormV2.vue
@@ -124,7 +124,7 @@
   const workQueueName = ref<string | null>(props.deployment.workQueueName)
   const retries = ref<number | null>(null)
   const retryDelay = ref<number | null>(null)
-  const jobVariables = ref<string | undefined>(can.access.flowRunInfraOverrides ? '{}' : undefined)
+  const jobVariables = ref<string | undefined>('{}')
 
   const { errors, validate: validateParameters } = useSchemaValidation(schema, parameters)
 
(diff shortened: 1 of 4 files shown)


abrookins:
@zhen0 Yes, this looks right, we should have removed this alongside the other PR.

abrookins (approved):

znicholasbrown (commented):
There's an unused `can` import in the automation actions run deployment input file now, otheriwse this LGTM!

znicholasbrown (approved):

collincchoy:
I think we need to handle/remove the `deploymentScheduleFlowRunInfraOverrides` flag also. 

I split what-was-originally 1 flag for flow-run-infra-overrides into 2 as the server-side only had partial support so we could roll out some of the functionality but hold on the scheduling updates (which needed server-side support). 

I'm _guessing_ the current state now though is that we want both on everywhere - in which case, we should update  usages of `deploymentScheduleFlowRunInfraOverrides` to default to "on" behavior.

src/components/FlowRunCreateFormV2.vue
```diff
@@ -124,7 +124,7 @@
   const workQueueName = ref<string | null>(props.deployment.workQueueName)
   const retries = ref<number | null>(null)
   const retryDelay = ref<number | null>(null)
-  const jobVariables = ref<string | undefined>(can.access.flowRunInfraOverrides ? '{}' : undefined)
+  const jobVariables = ref<string | undefined>('{}')
```
collincchoy, line 127: I feel like a bunch of handling could get simplified now by removing this `| undefined` in places.... 

but this should all work as-is

collincchoy (approved):

src/components/FlowRunCreateFormV2.vue
collincchoy: it looks the API itself calls this field optional so technically the typing is _correct_ but it'd certainly make everything simpler if we stuck to a single type especially if we're defaulting to a defined value. 

probably not worth the headache atm tho tbh

PR #2419, opened 2024-05-16
Fix: state colors in dark mode using prefers-color-scheme over in-app toggle

collincchoy:
The state colors aren't properly being set when in dark mode. 

More specifically, the `dark:` tailwind prefix isn't working correctly when used within prefect-ui-library because the tailwind config is configuring `darkMode: 'media'` which uses a media query on prefers-color-scheme instead of the class-based strategy that prefect-design uses. This means that if your browser color theme preference does not match the in-app, manually selected color theme (via the toggle), then ui-library is likely using the wrong theme which mismatches with theming elsewhere in the app.

| **Before** | **After** |
| ---- | ---- |
| image | image |


src/tailwind.config.ts
```diff
@@ -34,4 +34,5 @@ export default {
       },
     },
   },
+  presets: [],
```
collincchoy, line 37: > By default, presets themselves extend Tailwind’s [default configuration](https://github.com/tailwindlabs/tailwindcss/blob/master/stubs/config.full.js) just like your own configuration would. If you’d like to create a preset that completely replaces the default configuration, include an empty presets key in the preset itself:

The default configuration sets `darkMode: 'media'` whereas the prefect-design config uses `darkMode: 'class'`. Because this(this file's) config is used as a preset in the top-level tailwind.config.ts but _after_ the prefectDesignConfig (like `presets: [prefectDesignConfig, baseConfig]`), this default is overriding what prefect-design sets. 

An alternate fix could be to swap the order of the presets array, but opting to go this route instead as it seems more correct. prefect-design's tailwind config should be the base and then ui-library-specific things should go on _top_. _however_ because we're going to be extending prefect-design's config already as a preset, we don't need to pull in the default configuration for the base _here_ so setting presets to an empty array should keep this config just as it reads in this file.

znicholasbrown (approved):

PR #2420, opened 2024-05-16
Fix: Add default notification body/subject templates for custom automation triggers

collincchoy:
Fixes a bug where selecting a custom trigger during automation creation, and then attempting to use a Send Notification (or in cloud a Call a webhook) action fails because of missing default notification body + subject templates.

### DIFF
--- a/src/automations/defaultActionValues.ts
+++ b/src/automations/defaultActionValues.ts
@@ -58,6 +58,24 @@ Flow run URL: {{ flow_run|ui_url }}
 State message: {{ flow_run.state.message }}
 `.trim()
 
+const NOTIFICATION_BODY_CUSTOM = `
+Automation: {{ automation.name }}
+Description: {{ automation.description }}
+
+Event: {{ event.id }}
+Resource:
+{% for label, value in event.resource %}
+{{ label }}: {{ value }}
+{% endfor %}
+Related Resources:
+{% for related in event.related %}
+    Role: {{ related.role }}
+    {% for label, value in related %}
+    {{ label }}: {{ value }}
+    {% endfor %}
+{% endfor %}
+`.trim()
+
 export function getDefaultNotificationBody(template: AutomationTriggerTemplate): string {
   switch (template) {
     case 'deployment-status':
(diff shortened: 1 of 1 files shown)


src/automations/defaultActionValues.ts
```diff
@@ -58,6 +58,24 @@ Flow run URL: {{ flow_run|ui_url }}
...
+{% for label, value in event.resource %}
+{{ label }}: {{ value }}
+{% endfor %}
+Related Resources:
+{% for related in event.related %}
+    Role: {{ related.role }}
+    {% for label, value in related %}
+    {{ label }}: {{ value }}
+    {% endfor %}
+{% endfor %}
```
collincchoy, line 76: This is from the bottom of the [automation docs](https://docs.prefect.io/latest/concepts/automations/?h=automation#templating-with-jinja)

> In addition to those shortcuts for flows, deployments, and work pools, you have access to the automation and the event that triggered the automation. See the [Automations API](https://app.prefect.cloud/api/docs#tag/Automations) for additional details.

src/automations/defaultActionValues.ts
```diff
@@ -68,8 +86,11 @@ export function getDefaultNotificationBody(template: AutomationTriggerTemplate):
       return NOTIFICATION_BODY_WORK_POOL_STATUS
     case 'work-queue-status':
       return NOTIFICATION_BODY_WORK_QUEUE_STATUS
+    case 'custom':
+      return NOTIFICATION_BODY_CUSTOM
     default:
-      throw new Error(`Default notification body missing for template type: ${template}`)
+      const exhaustiveCheck: never = template
+      throw new Error(`Default notification body missing for template type: ${exhaustiveCheck}`)
```
collincchoy, line 93: weak opinion that this should be `return ''` instead of throwing a runtime error.
altho now that this exchaustiveCheck: never is here, we _should_ get issues at build-time now instead.

znicholasbrown (approved):

pleek91 (approved):
Thank you!

PR #2430, opened 2024-05-19
Enhancement: More efficient next flow run queries

znicholasbrown:
This PR adds two new routes the UI API that consume a new utility API for getting a map of flow ids to their next, partially-hydrated, runs. This PR, like https://github.com/PrefectHQ/prefect-ui-library/pull/2416 and others, is part of an effort to reduce the number of calls we're making on pages. This PR replaces a max 50 calls (1 per flow in the flow table) with a single  batched call. 

Note: this is a mostly-invisible change; the only visual difference is that this replaces the flow run icon with the state icon for the run, since previously this was making a separate call to the flow runs batcher otherwise (!!!)


src/models/NextFlowRun.ts
znicholasbrown, line 1: I expect to make this model slightly more generic in the near future, to better match the API model, which is just an abridged version of the flow run model.

src/components/FlowRunName.vue
pleek91, line 1: Why accept the individual pieces when I think all the implementations have the whole flow run?

pleek91: Oh because the next flow run by flow endpoint doesn't return the whole flow run?

Are we really saving much by stripping down that response? If it returned the whole flow run then it would be compatible with all of our existing components. Fully support the endpoint and its ability to be batched. Not sure about the stripped down model though.

src/services/UiApi.ts
```diff
@@ -34,4 +44,16 @@ export class UiApi extends WorkspaceApi implements IUiApi {
 
     return data
   }
+
+  public async getNextRunsByFlow(flowIds: string[]): Promise<UiNextFlowRunByFlow> {
+    const request = { 'flow_ids': [...new Set(flowIds)] }
```
pleek91, line 49: Do we need to dedupe this? Creating sets is slow. Probably not a big deal since this shouldn't get called many times. But wouldn't the api dedupe this as well in its query?

src/components/FlowRunName.vue
znicholasbrown: I think we're saving specifically in the department of doing a join to hydrate the state object - outside of that I'm not totally sure but regardless part of the reason we can batch this is because we're using an under-hydrated model so even if this returned some more fields it wouldn't return the entire object

src/services/UiApi.ts
znicholasbrown: I think the API definitely will handle that - the case I was solving for here was when multiple components for the same flow id made the request - my thinking was that deduping to start could save us round-trips when we run into limits on the batcher. E.g. If i make 201 calls (batcher limits 200) and the 199th and 201st call are the same ID. Obviously a contrived scenario but I could see us running into it pretty easily

znicholasbrown: Oh but actually now that I walk the code.. this deduping would need to happen at the batch level and not at the route level 🤦🏻

pleek91 (approved):
Discussed concerns in slack. Approving to keep things rolling

PR #2431, opened 2024-05-19
Enhancement: List limits

znicholasbrown:
This PR updates the deployment, block documents, variables, and flows list components to take advantage of the new p-pager limit functionality. Each table now has a default items per page limit of 10 with the ability to change the default up to the previous max of 50.
 
There are several reasons to lower the default limit on these:
1. Improved performance; many of these tables make calls on a per-row basis or to more expensive endpoints; each item in the list adds to the overall load time and increases the potential number of times that a user might experience content shifting on the page as data is hydrated
2. There's [research/guidance](https://catsy.com/blog/how-many-products-should-i-show-per-page) to suggest that the ideal number of items to display in a list is somewhere in the 5-50 range, skewing lower depending on the complexity of the objects being shown. Following this guidance lowers the amount of cognitive overhead required to scan these lists and helps users better learn and navigate the product.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -43,7 +43,7 @@
                 "vue-tsc": "^2.0.17"
             },
             "peerDependencies": {
-                "@prefecthq/prefect-design": "^2.10.4",
+                "@prefecthq/prefect-design": "^2.10.5",
                 "@prefecthq/vue-charts": "^2.0.3",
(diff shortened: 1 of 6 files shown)


pleek91 (approved):

pleek91:
Nice love this

PR #2434, opened 2024-05-20
Fix: missing vertical space on work pool card.

collincchoy:
Updates to `<p-card />` removed the intended gap on these cards

| **Before** | **After** |
| ---- | ---- |
| image | image |

### DIFF
--- a/src/components/WorkPoolCard.vue
+++ b/src/components/WorkPoolCard.vue
@@ -1,6 +1,6 @@
 <template>
-  <p-card>
-    <div class="work-pool-card">
+  <p-card class="work-pool-card">
+    <p-content secondary>
       <div class="work-pool-card__header">
         <div class="work-pool-card__heading">
           <div class="work-pool-card__name">
@@ -38,7 +38,7 @@
           {{ lastPolled }}
         </div>
       </div>
-    </div>
+    </p-content>
   </p-card>
 </template>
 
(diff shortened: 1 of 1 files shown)


src/components/WorkPoolCard.vue
```diff
@@ -1,40 +1,42 @@
...
-  <p-card class="work-pool-card">
-    <div class="work-pool-card__header">
-      <div class="work-pool-card__heading">
-        <div class="work-pool-card__name">
-          <p-link class="work-pool-card__name" :to="routes.workPool(workPool.name)">
-            {{ workPool.name }}
-          </p-link>
-          <WorkPoolStatusIcon :work-pool="workPool" />
+  <p-card>
+    <div class="work-pool-card">
```
pleek91, line 3: fine as is but be nice to do to keep the block class as the root and to also use p-content rather than having custom styling here. 
```
<p-card class="work-pool-card">
  <p-content secondary>
```

pleek91 (approved):

src/components/WorkPoolCard.vue
collincchoy: `<p-card class="work-pool-card">` like just for naming but remove the styles? previously, that class was actually giving styles here

pleek91: Yeah just for naming and remove the styles.

collincchoy: I added p-content. but... left the styling on the class. dang it.

PR #2436, opened 2024-05-20
Chore: Improve truncated deployment name/flow column styles.

collincchoy:
Follow-up to https://github.com/PrefectHQ/prefect-design/pull/1301 which removes titles at the `p-table` level. This PR _adds_ `title` attributes to the truncated text in this table.


src/components/FlowPopover.vue
```diff
@@ -63,11 +63,11 @@
 }
 
 .flow-popover__link-name { @apply
-  shrink
   truncate
 }
 
 .flow-popover__link-icon { @apply
+  shrink-0
```
collincchoy, line 70: think this was backwards. 

i noticed the icon shrinking as i brought the width down. this is to stop that - the icon should stay the same size.

pleek91 (approved):

PR #2442, opened 2024-05-21
remove variable value search

WillRaphaelson:
removes the call to filter variables by value


znicholasbrown (approved):
I'm guessing this isn't possible anymore?

WillRaphaelson:
> I'm guessing this isn't possible anymore?
 yep, we're removing this from the api.

PR #2464, opened 2024-05-24
Enhancement: Show pagination even if there's just 1 page

znicholasbrown:
This PR removes the constraint that we only show pagination info when tables contain more than 1 page. This was a confusing constraint generally.


pleek91 (approved):
I like it

PR #2466, opened 2024-05-24
Don't require parameters that exist on the deployment when creating a "Run deployment" automation

pleek91:
# Description
When creating an automation with a "Run deployment" action all required parameters from the deployment schema were marked as required in the ui even if the deployment already had a value. This prevented users from using the "omit value" feature to make sure that an automation did not overwrite the deployment. 

Marking all parameters that the deployment has a default value for as optional means these values can be omitted. That way when the action runs the deployment's value for the parameter will be used (even if it has been changed since the automation was create)

Related to: https://github.com/PrefectHQ/nebula-ui/issues/4738


znicholasbrown (approved):
makes sense to me

PR #2471, opened 2024-05-28
Enhancement: JSON Variables

WillRaphaelson:
This creates a separate layer of models, mappers, api methods for Variables V2, which support any json type, in order to put them behind a feature flag.


WillRaphaelson:
@znicholasbrown no rush here, but at some point I added
```
'access:jsonVariables',
```

to the can.ts file in /services. Its not being used anywhere I dont think, is this needed?

znicholasbrown:
Yup you can remove that @WillRaphaelson

src/components/VariablesPageEmptyState.vue
```diff
@@ -9,7 +9,7 @@
     </template>
 
     <template #description>
-      Variables store non-sensitive strings.
+      Variables store non-sensitive pieces of JSON.
```
znicholasbrown, line 12:
```suggestion
      Variables store non-sensitive strings.
```

This is still correct for the existing version of variables.

src/models/api/VariableV2Request.ts
```diff
@@ -0,0 +1,11 @@
+export type VariableV2CreateRequest = Partial<{
+  name: string | null,
+  value: unknown,
```
collincchoy, line 3: Is this the only difference between v1 and v2? That previously this was a string, but now it could be... what - an object?

znicholasbrown (approved):
One suggestion, otherwise lgtm

src/models/api/VariableV2Request.ts
WillRaphaelson: any json compatible type!

PR #2474, opened 2024-05-29
Json string var type

WillRaphaelson:
this pr stops the ui from adding extra quotes around strings when sending to the api


znicholasbrown (approved):
makes sense - thanks @WillRaphaelson / @pleek91

PR #2476, opened 2024-05-29
Skip populating default values when creating "Run deployment" automation actions

pleek91:
# Description
Parameter forms are generated based on json schemas which can include default values for properties.  When a form is generated it also populates the default values. This makes sense when creating a flow run directly in the ui, the user sees exactly what parameters the flow run will be created with. 

But in the context of automations that means the automation will always be executed with the default parameters from the deployment at the time the action was created. Which means updating/changing the deployments defaults has no effect on future executions. 

This PR updates the parameters form for the "Run deployment" action so it **does not populate the default values from the deployment**. That way when the automation action is executed it will always be executed with the latest deployment defaults unless the user has specifically overwritten the value in the automation action itself. 

image

Related to https://github.com/PrefectHQ/prefect-ui-library/pull/2466
Closes https://github.com/PrefectHQ/nebula-ui/issues/4738


src/schemas/compositions/useSchemaFormSettings.ts
pleek91, line 1: The majority of this PR is combining two existing compositions (useSchemaFormKinds and useSchema) into a single composition (this composition) so I could add a 3rd provided value (skipDefaultValueInitialization) without creating another composition. This composition now serves as a settings object that is provided for the entire form.

src/schemas/components/SchemaFormProperty.vue
```diff
@@ -114,7 +116,7 @@
     },
   })
 
-  if (!props.skipDefaultValueInitialization && !isDefined(props.value) && isDefined(property.value.default)) {
+  if (!skipDefaultValueInitializationForAllProperties && !props.skipDefaultValueInitialization && !isDefined(props.value) && isDefined(property.value.default)) {
```
pleek91, line 119: This is the primary change. Checking both the local prop and the global form setting to see if the default value initialization should be skipped.

znicholasbrown (approved):

PR #2478, opened 2024-05-29
Enhancement: Image artifact

dylanbhughes:
Screenshot 2024-05-29 at 2 59 49 PM

Displays an image artifact from the UI


src/components/ArtifactDataImage.vue
```diff
@@ -0,0 +1,26 @@
+<template>
+  <div class="artifact-data-image">
+    <img loading="eager" :src="artifact.data">
```
pleek91, line 3: nit: eager is the default
```suggestion
    <img :src="artifact.data">
```

pleek91 (approved):

PR #2488, opened 2024-05-31
Fix TimezoneSelect defaulting to `UTC`

pleek91:
# Description
Different implementations can have different logic for what `null` means for a timezone. For example user preferences in cloud use `null` for "Local" time (the browser's default timezone). But because of this component defaulting to `UTC` even when the user had no value for timezone in their settings (and it would say "Local" when viewing preferences) it would say `UTC` when editing (even though that is not what was saved in their profile). 

Removing the null coalesce fixes this. And other implementations of this component specify `UTC` as the default and continue to work as intended. 

**Preferences**
image

**Editing Before**
image

**Editing After**
image

### DIFF
--- a/src/components/TimezoneSelect.vue
+++ b/src/components/TimezoneSelect.vue
@@ -25,7 +25,7 @@
 
   const internalValue = computed({
     get() {
-      return props.modelValue ?? 'UTC'
+      return props.modelValue
     },
     set(val) {
       emit('update:modelValue', val)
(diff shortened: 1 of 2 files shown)


dylanbhughes (approved):

src/components/TimezoneSelect.vue
```diff
@@ -13,27 +13,15 @@
...
     hideUnset?: boolean,
   }>()
 
-  const emit = defineEmits<{
-    (event: 'update:modelValue', value: string | null): void,
-  }>()
-
-  const internalValue = computed({
-    get() {
-      return props.modelValue ?? 'UTC'
```
pleek91, line 28: Specifically removing the `?? 'UTC'` is the fix here. Updating the syntax to use `defineModel` as well.

PR #2505, opened 2024-06-06
Update pagination for flow runs

pleek91:
# Description
Moving over from using the deprecated `usePaginationX` compositions which attempted to do pagination using multiple api calls into a single call using the new `/paginate` endpoint for flow runs. This means

- No more "infinite" mode. Paging only
- No more multiple calls
- No more flickering when an interval poll happens

Putting this up for review but can't merge this yet until I have the nebula-ui and oss PRs ready because this is a breaking change to the `usePaginatedFlowRuns` composition.

### DIFF
--- a/src/components/FlowRunFilteredList.vue
+++ b/src/components/FlowRunFilteredList.vue
@@ -5,7 +5,7 @@
         <p-select-all-checkbox v-model="selected" :selectable="flowRuns.map((flowRun) => flowRun.id)" item-name="flow run" />
       </template>
 
-      <ResultsCount v-if="selected.length == 0" :count="total" label="Flow run" />
+      <ResultsCount v-if="selected.length == 0" :count label="Flow run" />
       <SelectedCount v-else :count="selected.length" />
       <FlowRunsDeleteButton v-if="can.delete.flow_run" :selected="selected" @delete="deleteFlowRuns" />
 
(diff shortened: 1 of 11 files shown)


src/compositions/filters.ts
```diff
@@ -559,6 +559,76 @@ export function useDeploymentsFilter(defaultValue: MaybeReactive<DeploymentsFilt
   return useUnionFilter<DeploymentsFilter>(defaultValue, defaultDeploymentSort)
 }
 
+function usePaginationUnionFilter<T extends PaginationUnionFilter>(defaultValue: MaybeReactive<T>, defaultSort: Exclude<T['sort'], undefined>): UseFilter<WithPage<T>> {
```
pleek91, line 562: Worth pointing out the `WithPage` part of the return type. Because `p-pager` complains about `page` being `number | undefined` I opted to have these compositions accept the normal filter (where page is optional) but always return `number` for page and default the value to `1`. This matches the api logic and makes the types for these filters much easier to work with at implementation.

pleek91: Also, this composition is a straight copy/paste of the `useUnionFilter` composition and just changed the types and added the page default. So probably not worth digging into how this actually works unless you're just curious.

src/components/FlowRunFilteredList.vue
```diff
@@ -65,23 +71,24 @@
   const selected = ref<string[]>([])
   const searchTerm = ref('')
   const searchTermDebounced = useDebouncedRef(searchTerm, 500)
+  const { value: limit } = useLocalStorage('flow-run-list-limit', 10)
```
collincchoy, line 74: What's the use in having this in local storage?

src/components/FlowRunFilteredList.vue
```diff
@@ -65,23 +71,24 @@
...
-  const { filter, clear, isCustomFilter } = useFlowRunsFilterFromRoute(merge({}, props.filter, {
+  const { filter, clear, isCustomFilter } = useFlowRunsPaginationFilterFromRoute(merge({}, props.filter, {
     flowRuns: {
       nameLike: searchTermDebounced,
     },
+    limit,
   }), props.prefix)
 
-  const { flowRuns, total, subscriptions, next } = usePaginatedFlowRuns(filter, {
+  const { flowRuns, count, pages, subscription } = usePaginatedFlowRuns(filter, {
```
collincchoy, line 83: > `subscriptions` -> `subscription`

:chefs-kiss:

collincchoy (approved):

src/components/FlowRunFilteredList.vue
pleek91: I'm just following the pattern @znicholasbrown has been using for pagination. There's a dropdown in `p-pager` that lets you select how many results to show per page. And if you change the value it remembers.

znicholasbrown (approved):

Issue #2404, opened 2024-05-14
limit Custom Run submit button to 1 click


jakekaplan:
When doing a custom run and clicking `submit` twice in quick succession a user is able to kick off 2 flow runs when they almost certainly intended to kick off one flow run. This can be very confusing and scary to users when they see things happen twice. We should lock the `Submit` button after a single click to prevent abuse

![Screenshot 2024-05-14 at 6 35 04 PM](https://github.com/PrefectHQ/prefect-ui-library/assets/40362401/0b6374f3-7ab5-4a61-b752-cb21de64fb6d)

![Screenshot 2024-05-14 at 6 35 43 PM](https://github.com/PrefectHQ/prefect-ui-library/assets/40362401/8bcab129-86cb-4b8c-a5b3-838eea3c3da3)

PR #2525, opened 2024-06-12
Add the ability to disable the flow run create form submit button

pleek91:
# Description
Part of fixing https://github.com/PrefectHQ/prefect-ui-library/issues/2404

Will need implementation in oss and cloud ui

### DIFF
--- a/src/components/FlowRunCreateFormV2.vue
+++ b/src/components/FlowRunCreateFormV2.vue
@@ -68,7 +68,7 @@
       <p-button @click="emit('cancel')">
         Cancel
       </p-button>
-      <p-button variant="default" type="submit">
+      <p-button variant="default" type="submit" :disabled>
         Submit
       </p-button>
     </template>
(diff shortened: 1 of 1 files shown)


src/components/FlowRunCreateFormV2.vue
```diff
@@ -126,6 +127,10 @@
   const { errors, validate: validateParameters } = useSchemaValidation(schema, parameters)
 
   async function submit(): Promise<void> {
+    if (props.disabled) {
+      return
+    }
```
znicholasbrown, line 132: I'm surprised this is needed with the submit button disabled? But also doesn't hurt.

znicholasbrown (approved):

src/components/FlowRunCreateFormV2.vue
pleek91: I tried a few different things and button smashing was still creating double runs sometimes. This I wasn't able to get past.

pleek91: I think just the order of execution and vue's rendering means the button can still be clicked multiple times before it actually gets disabled. This makes sure the submit never happens more than once.

PR #2536, opened 2024-06-20
Chore: Add missing DeploymentComboboxOption import.

collincchoy:
can't see options in DeploymentCombobox anymore.

Regression from https://github.com/PrefectHQ/prefect-ui-library/pull/2528


pleek91 (approved):
Ouch, thank you!

PR #2558, opened 2024-06-26
Disable validation checkboxes from parameter forms when `enforce_parameter_schema` is enabled

pleek91:
# Description
Disables the "validate parameters before submitting" checkboxes on all deployment parameter forms when enforce parameter schema is enabled. `enforce_parmeter_schema` is now  defaulted to `true` in 3.x. This is a intermediary step before removing the checkbox altogether.

image


collincchoy (approved):
looks right to me!

collincchoy (approved):

PR #2571, opened 2024-07-01
Fix "This run didn't generate logs" being displayed when logs exist

### LINKED ISSUE #2565: Delay in logs and summary for crashed flow runs
When watching a flow run execute, a user can see the flow run state transition to 'Crashed'. For about 10 seconds after seeing the flow run transition states, the UI will say 'This run didn't generate logs', and 'A summary has not been generated for this flow run'. This can cause users to exit the page prematurely before seeing more descriptive information about why the flow crashed. Attached are pictures of a flow run at the instant it crashes, and then when it eventually displays more information.

Possible quick fixes:
-Add a loading wheel for ~10 seconds
-Change text to say 'A summary has not been generated for this flow run **yet**." and "This run didn't generate logs **yet**."

Screenshot 2024-06-27 at 3 42 15 PM
Screenshot 2024-06-27 at 3 42 52 PM

pleek91:
# Description
Because there can be a delay between a run being in a terminal state and the log events being processed, logs might not show up before the UI has stopped polling for them. This can specifically happy with runs that crashed immediately. 

To account for this we want to keep polling for logs for a limited time after the run has completed. I've specified 30 seconds which should be plenty long enough to account for any log delay. If it is not we probably want to think about this more holistically and not as just a ui problem. 

Closes https://github.com/PrefectHQ/prefect-ui-library/issues/2565
Related https://github.com/PrefectHQ/nebula-ui/issues/4304

### DIFF
--- a/src/components/FlowRunLogs.vue
+++ b/src/components/FlowRunLogs.vue
@@ -41,11 +41,12 @@
 </template>
 
 <script lang="ts" setup>
-  import { ref, computed, toRefs, watch } from 'vue'
+  import { SubscriptionOptions, useNow } from '@prefecthq/vue-compositions'
+  import { differenceInSeconds } from 'date-fns'
+  import { ref, computed } from 'vue'
   import { LogLevelSelect, LogsContainer, LogsSort } from '@/components'
   import { useLogsSort, useWorkspaceApi } from '@/compositions'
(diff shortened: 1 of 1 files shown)


znicholasbrown (approved):
nice, makes sense to me.

PR #2573, opened 2024-07-01
Fix: run deployment form state shared by row index

### LINKED ISSUE #14461: Quick run deployment parameters shared across deployments
### First check

### Bug summary

_Sometimes_ default parameters are being overwritten when trying to quick run deployments from the listing view. It happens when I run one, search for another, and then run that one and then the latter's default parameters will sometimes be overridden/wrong/missing. 

This also happens when paging. 

### Reproduction

I've been able to repro shared form state across the Run Deployment model for 2 different deployments. It's weird?
have 2 deployments with flows that have an overlapping parameter name (ex. 2 flows with a name parameter) - the rest of the params can diverge, but you need 1 required one to trigger the modal.
search for 1 by name so that 1 result shows up, quick run, put a value in for name. search for 2nd by name so that 1 result shows up, quick run, and see the name from the other deployment.
:bulb: seems order-dependent. i.e. 1st row leaks form state by parameter name when the row's deployment changes. Doesn't happen when viewing many deployments and running between them without shifting the results order.
:bulb: can reproduce with pagination instead of search. You just need to be lucky and have a deployment land at the same row index # with an overlapping param name and at least 1 required param

### Error

No error - just confusing UX and sometimes accidentally trigger with wrong parameters

### Browsers

- [X] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### Prefect version

```Text
2.19.7
```

### Additional context

Bug is in prefect cloud UI as well.

collincchoy:
Closes https://github.com/PrefectHQ/prefect/issues/14461

### DIFF
--- a/src/components/FlowList.vue
+++ b/src/components/FlowList.vue
@@ -20,6 +20,7 @@
       :data="flows"
       :columns="columns"
       :column-classes="columnClasses"
+      :row-key="flow => flow.id"
       class="flow-list__table"
       @update:selected="selectedFlows = $event"
     >
--- a/src/components/FlowList.vue
+++ b/src/components/FlowList.vue
@@ -20,7 +20,7 @@
       :data="flows"
       :columns="columns"
       :column-classes="columnClasses"
-      :row-key="flow => flow.id"
+      :row-key="(flow: Flow) => flow.id"
       class="flow-list__table"
       @update:selected="selectedFlows = $event"
     >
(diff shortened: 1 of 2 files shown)


src/components/FlowList.vue
```diff
@@ -20,6 +20,7 @@
       :data="flows"
       :columns="columns"
       :column-classes="columnClasses"
+      :row-key="flow => flow.id"
```
collincchoy, line 23: Not really relevant but figured 🤷

znicholasbrown (approved):

PR #2582, opened 2024-07-02
Enhancement: ANSI color support in logs

znicholasbrown:
This PR builds on #2569 by adding support for ansi color codes in logs:

Before:

Screenshot 2024-07-02 at 5 53 05 PM

After:

Screenshot 2024-07-02 at 5 52 40 PM


src/components/LogRow.vue
```diff
@@ -36,55 +38,88 @@
...
   import { formatTimeNumeric } from '@/utilities/dates'
+  import { ansiiColorRegex, ansiToTailwind } from '@/utilities/logs'
   import { urlRegex } from '@/utilities/urls'
 
   const props = defineProps<{
     log: Log,
   }>()
 
+  const taskRunId = computed(() => props.log.taskRunId)
+  const { taskRun } = useTaskRun(taskRunId)
```
pleek91, line 49:
```suggestion
  const { taskRun } = useTaskRun(() => props.log.taskRunId)
```

pleek91 (approved):
I'm a little hesitant to support long term if its not 100% accurate but I think overall the approach is fine.

PR #2588, opened 2024-07-08
Remove old work queues part 1

collincchoy:
Part 1 of work queue removal. 

Following this I will:
1. cut a new version and bump in downstream apps. Removed route definitions so that the routes will 404 (those that shouldn't have any remaining references in app code). The 1 workQueue(id) route will remain since it's still used in 2 small places - this PR "shims" it by stubbing in a page component that fetches the work queue to redirect to the work pool's corresponding work queue page. Everything should still work - no breaking changes yet.
2. upgrade in downstream apps. Can then start removing all work queue related files (pages, components, & route definitions) from downstream app repositories
3. come back to ui-library and cull out the rest -- at this point there will be breaking changes but the downstream repo's will have been already updated to handle them.
4. cut a new version and upgrade in downstream repos

### DIFF
--- a/src/router/index.ts
+++ b/src/router/index.ts
@@ -158,7 +158,7 @@ export function createWorkspaceRouteRecords(components: Partial<WorkspaceRouteCo
         {
           name: 'workspace.work-queues.work-queue',
           path: 'work-queue/:workQueueId',
-          component: components.workQueue,
+          component: components.workQueue ?? import('@/components/WorkQueueToWorkPoolQueueRedirect.vue'),
         },
         {
           name: 'workspace.work-queues.work-queue-edit',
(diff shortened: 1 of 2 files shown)


src/router/index.ts
```diff
@@ -142,31 +142,10 @@ export function createWorkspaceRouteRecords(components: Partial<WorkspaceRouteCo
...
-          name: 'workspace.work-queues.work-queue-create',
-          path: 'create',
-          component: components.workQueueCreate,
-          meta: {
-            can: 'create:work_queue',
-          },
-        },
         {
           name: 'workspace.work-queues.work-queue',
           path: 'work-queue/:workQueueId',
```
collincchoy, line 147: Going to hold onto this route for now. There are 2 components that use it so for now this route renders a stub that redirects to the newer work-pool.work-queue route.

pleek91 (approved):

PR #2605, opened 2024-07-10
Add the flow run's automation to the page heading

pleek91:
# Description
If a flow run was run from a "Run deployment" automation action we can show the automation in the page heading along with the other information like the flow, deployment, work pool, etc. 

image

Related to https://github.com/PrefectHQ/nebula-ui/issues/4357

**Note**
Generally each related entity in the page heading also appears in the list item. For now I'm not adding the automation to the list item because currently there isn't a way to batch fetch the automations for a list of flow runs without making a very general events query with a potentially wide occurred span.


src/components/FlowRunAutomation.vue
```diff
@@ -0,0 +1,65 @@
...
+      return null
+    }
+
+    // purposefully adding some margin around the expectedStartTime
+    // because the automation being executed should happen prior to the time the run was scheduled
+    const until = addMinutes(props.flowRun.expectedStartTime, 5)
+    const since = subMinutes(until, 5)
+
+    return {
+      resource: {
```
collincchoy, line 30: I think this would match automation created/updated events?

maybe should identify a particular event? automation.triggered or executed?

src/components/FlowRunAutomation.vue
```diff
@@ -0,0 +1,65 @@
...
+
+  const { events } = useWorkspaceEvents(() => {
+    if (!props.flowRun.expectedStartTime) {
+      return null
+    }
+
+    // purposefully adding some margin around the expectedStartTime
+    // because the automation being executed should happen prior to the time the run was scheduled
+    const until = addMinutes(props.flowRun.expectedStartTime, 5)
+    const since = subMinutes(until, 5)
```
collincchoy, line 27:
```suggestion
    const since = subMinutes(props.flowRun.expectedStartTime, 5)
```
?

or isn't this otherwise the same as:
```suggestion
    const since = props.flowRun.expectedStartTime
```
?

collincchoy (approved):
2 q's/comments, but otherwise lgtm

src/components/FlowRunAutomation.vue
pleek91: Ah yes good catch, forgot to target the specific event I was looking at. Updated in https://github.com/PrefectHQ/prefect-ui-library/pull/2605/commits/051d0484be639fe6cef8e3ce3e39aa5268da3d2a

src/components/FlowRunAutomation.vue
pleek91: I updated this in https://github.com/PrefectHQ/prefect-ui-library/pull/2605/commits/0af48c5d5883f71b4880aa31bf5bbbf5f77a3ab9 because this was 1) unclear and 2) incorrect. I wanted a 5 minute buffer on either side.

collincchoy (approved):

PR #2606, opened 2024-07-11
Converge old work queue and work pool queue models

collincchoy:
Following in the footsteps of https://github.com/PrefectHQ/prefect-ui-library/pull/2588 and https://github.com/PrefectHQ/prefect-ui-library/pull/2601, this PR continues on the journey of purging the blurry lines of a `WorkQueue` vs `WorkPoolQueue`. 

The aforementioned PRs removed a majority of the old WorkQueue components and some models, but this PR goes on to fully _remove_ the `WorkQueue*` models in favor of `WorkPoolQueue` ones. 

Can we do that?? - you may ask. 

Well actually, the server treats both sets of APIs the _same_. They use the same exact schemas for both requests and schemas. For example, compare these 2 create work queue routes - they use the same action(request body) and response schemas - https://github.com/PrefectHQ/prefect/blob/6dbf284f9c99b3dfbbecf22638c5d71d621f6b93/src/prefect/server/api/workers.py#L403-L408 and https://github.com/PrefectHQ/prefect/blob/6dbf284f9c99b3dfbbecf22638c5d71d621f6b93/src/prefect/server/api/work_queues.py#L37-L40

Comments left inline


src/components/WorkPoolQueueCreateForm.vue
```diff
@@ -56,7 +56,7 @@
   const routes = useWorkspaceRoutes()
   const { validate, pending } = useValidationObserver()
 
-  const name = ref<string>()
+  const name = ref('')
```
collincchoy, line 59: `WorkQueueCreate` requires name, but for some reason the ui's WorkPoolQueueCreate and WorkQueueCreate types diverged on this.

https://github.com/PrefectHQ/prefect/blob/6dbf284f9c99b3dfbbecf22638c5d71d621f6b93/src/prefect/server/schemas/actions.py#L845-L846

This form already uses an `isRequired` validation rule here too so shouldn't be any functional change here.

src/components/WorkQueueCombobox.vue
```diff
@@ -17,7 +17,7 @@
   import { useSubscription } from '@prefecthq/vue-compositions'
   import { computed, watch } from 'vue'
   import { useWorkspaceApi } from '@/compositions'
-  import { WorkQueue } from '@/models'
+  import { WorkPoolQueue } from '@/models'
```
collincchoy, line 20: This PR completely removes the old `WorkQueue` model since to the server, they're both the exact same. The only meaningful differences between the types (not-server-side) have been removed in previous PR's barring the deprecated `workQueue.filter` -- which this PR also removes as it's 1) deprecated and 2) I couldn't find any usage of it anymore.


So throughout this PR, you'll see a lot of "renaming" of `WorkQueue` -> `WorkPoolQueue` for consistency.

The only places you should still see just `WorkQueue` is when deliberately working with queues w/o their parent work pool - i.e. get all work queues in workspace.

src/maps/filters.ts
```diff
@@ -556,7 +541,14 @@ export const mapWorkPoolWorkersFilter: MapFunction<WorkPoolWorkersFilter, WorkPo
 
 export const mapWorkPoolQueuesFilter: MapFunction<WorkPoolQueuesFilter, WorkPoolQueuesFilterRequest> = function(source) {
   return removeEmptyObjects({
-    work_pools: this.map('WorkPoolFilter', source.workPools, 'WorkPoolFilterRequest'),
+    work_queues: {
```
collincchoy, line 544: 😱 - this was actually wrong before. the Read Work Queues API that's under Work Pools uses a _work queue_ filter here (https://app.stg.prefect.dev/api/docs#tag/Work-Pools/operation/read_work_queues_api_accounts__account_id__workspaces__workspace_id__work_pools__work_pool_name__queues_filter_post or https://github.com/PrefectHQ/prefect/blob/6dbf284f9c99b3dfbbecf22638c5d71d621f6b93/src/prefect/server/api/workers.py#L472-L473)

At first, I just aliased mapWorkPoolQueuesFilter = mapWorkQueuesFilter and similarly for the types, and then realized i could just cut the old one out and only use 1.

src/models/WorkPoolQueueCreate.ts
```diff
@@ -1,7 +1,7 @@
-export type WorkPoolQueueCreate = Partial<{
+export type WorkPoolQueueCreate = {
   name: string,
```
collincchoy, line 2: as mentioned earlier, name is actually required from the API here.

pleek91 (approved):
Nice, love seeing this cleaned up

PR #2610, opened 2024-07-11
Chore: consistent status label conversions

collincchoy:
Noticed these separate get<blank>StatusLabel functions when going through work queue stuff - this PR consolidates the deployment, work pool, and work pool queue status mappings to use those functions.


pleek91 (approved):
❤️

PR #2611, opened 2024-07-12
Enhancement: Duplicate deployment

dylanbhughes:
This PR adds a page for duplicating a deployment from the UI. Often a user has an existing deployment and wants to create a new one with different parameters. This is a quick and easy way to do so.


src/components/DeploymentFormV2.vue
```diff
@@ -107,6 +119,18 @@
   const { validate } = useValidationObserver()
   const { errors, validate: validateParameters } = useSchemaValidation(schema, parameters)
   const { state: overrideState, error: overrideError } = useValidation(jobVariables, isJson('Job variables'))
+  const { state: nameState, error: nameError } = useValidation(name, (value) => {
+    if (props.mode === 'update') {
+      return true
+    }
```
znicholasbrown, line 125: Doesn't this mean I could pass validation if I'm updating instead of duplicating and provide an invalid name?

src/components/DeploymentFormV2.vue
```diff
@@ -4,9 +4,16 @@
...
+    <template v-if="mode === 'duplicate'">
+      <p-label label="Name" :state="nameState" :message="nameError">
+        <p-text-input v-model="name" />
+      </p-label>
+    </template>
+    <template v-else>
+      <p-label label="Name">
+        <p-text-input :model-value="name" disabled />
+      </p-label>
+    </template>
```
znicholasbrown, line 16:
```suggestion
    <p-label label="Name" :state="nameState" :message="nameError" :disabled="mode === 'update'" >
        <p-text-input v-model="name" />
    </p-label>
```
You can use the same markup here and conditionally disable the input, I think

znicholasbrown (approved):
A couple of small comments, otherwise lgtm

src/components/DeploymentFormV2.vue
dylanbhughes: We disable the input for updating a deployment because a deployment name can't be updated

znicholasbrown: ah yeah good point

PR #2624, opened 2024-07-23
Fix variables sometimes displaying old or incorrect values when using the "click to view" preview

### LINKED ISSUE #2623: Click to view in the variables table brings up the wrong variables informations
The video shows this better than I can describe it but the variables table is showing incorrect information when I click on the "click to view" button for a long variable. 

To reproduce:
1. Create a variable with any value. 
2. Create a new variable with a long, json value. 
3. Once the variable is created, click on the "click to view" button. Note that it shows the name and value for a different variable. 
4. Refresh the page. 
5.  Click on "click to view" again - it should show the correct info. 

https://github.com/user-attachments/assets/814f043c-7a8a-4720-aeca-f00968772a4e

pleek91:
# Description
Closes https://github.com/PrefectHQ/prefect-ui-library/issues/2623

Adds a `rowKey` to the table to make sure row components are not reused across variables and fixes some reactivity issues with the value preview.

### DIFF
--- a/src/components/VariablesTable.vue
+++ b/src/components/VariablesTable.vue
@@ -28,6 +28,7 @@
       :data="variables"
       :columns="columns"
       :column-classes="columnClass"
+      :row-key="(variable: Variable) => variable.id"
       @update:selected="selectedVariables = $event"
     >
       <template #name="{ row }">


dylanbhughes (approved):

src/components/VariablesTable.vue
```diff
@@ -28,6 +28,7 @@
       :data="variables"
       :columns="columns"
       :column-classes="columnClass"
+      :row-key="(variable: Variable) => variable.id"
```
znicholasbrown, line 31: knew this would come in handy

znicholasbrown (approved):

PR #2626, opened 2024-07-24
Update flow run create forms to set `enforce_parameter_schema` on individual flow runs

pleek91:
# Description
UI side of https://github.com/PrefectHQ/prefect/pull/14718

Updates the "validate parameters before submitting" checkbox and its logic. 

- Changes the label to "Enforce parameter schema"
- Defaults to the value of the deployment's setting for `enforce_parameter_schema`
- Toggling the checkbox sets the value for `enforce_parameter_schema` on the flow run create request
- Unchecking the checkbox (or leaving it unchecked if the deployment defaults it to unchecked) will prevent all parameter validation (both when submitting and when the run is scheduled)

**Note**
Also updated the naming of some of these components and types to remove the "V2" suffix now that the "V1" components and types have been removed.


src/components/FlowRunCreateForm.vue
```diff
@@ -21,20 +21,7 @@
...
-              </template>
-              <div class="w-fit">
-                <p-checkbox v-model="shouldValidate" disabled label="Validate parameters before submitting" />
-              </div>
-            </p-tooltip>
-          </template>
-          <template v-else>
-            <p-checkbox v-model="shouldValidate" label="Validate parameters before submitting" />
-          </template>
+          <p-checkbox v-model="enforceParameterSchema" label="Enforce parameter schema" />
```
cicdw, line 24: I kind of like "Validate" better than "Enforce"; it implies there's a failure mode associated with it

```suggestion
          <p-checkbox v-model="enforceParameterSchema" label="Validate parameters" />
```

cicdw (changes requested):
small nit: i think the user-facing name for this should still be "validate" but the rest LGTM, thank you!

src/components/QuickRunParametersModal.vue
```diff
@@ -19,20 +19,7 @@
...
-            </template>
-            <div class="w-fit">
-              <p-checkbox v-model="validate" disabled label="Validate parameters before submitting" />
-            </div>
-          </p-tooltip>
-        </template>
-        <template v-else>
-          <p-checkbox v-model="validate" label="Validate parameters before submitting" />
-        </template>
+        <p-checkbox v-model="enforceParameterSchema" label="Enforce parameter schema" />
```
cicdw, line 22: ah i see here as well should be "validate"

cicdw (approved):

src/services/WorkspaceDeploymentsApi.ts
```diff
@@ -62,14 +62,14 @@ export class WorkspaceDeploymentsApi extends WorkspaceApi {
     return data
   }
 
-  public async createDeployement(deployment: DeploymentCreate): Promise<Deployment> {
+  public async createDeployment(deployment: DeploymentCreate): Promise<Deployment> {
```
znicholasbrown, line 65: subtle

znicholasbrown (approved):

PR #2637, opened 2024-07-30
Fix quick run modal validation and creating multiple runs from frustration clicks

pleek91:
# Description
This fixes two issues

1. Multiple runs created when user clicks "Run" multiple times. 
This would happen because the run button was not disabled while the async call was happening. Which made it appear that nothing was happening resulting in a second (and sometimes third) click by the user. This would result with as many runs being created as there were clicks. Disabling the button while the run is being created prevents multiple runs and also indicates to the user that work is happening.

2. Validation errors not being displayed in the form
If a validation error occurred the modal was closed and a toast was displayed. Fixed the pre-submit validation so that errors are displayed within the form itself and prevented the modal from closing when any error occurs so that users don't lose their work.

### DIFF
--- a/src/components/QuickRunParametersModal.vue
+++ b/src/components/QuickRunParametersModal.vue
@@ -3,6 +3,7 @@
     <SchemaFormV2
       :id="formId"
       v-model:values="parameters"
+      v-model:loading="loading"
       :schema="deployment.parameterOpenApiSchema"
       :validate="enforceParameterSchema"
       :kinds="['json', 'workspace_variable']"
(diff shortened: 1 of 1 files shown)


src/components/QuickRunParametersModal.vue
```diff
@@ -25,7 +26,7 @@
 
     <template #actions>
       <slot name="actions">
-        <p-button type="submit" primary :form="formId">
+        <p-button type="submit" :disabled="loading" primary :form="formId">
```
znicholasbrown, line 29:
```suggestion
        <p-button type="submit" :loading primary :form="formId">
```
any reason not to use the loading state here? that should also disable the button

znicholasbrown (approved):

PR #2641, opened 2024-08-05
Allow api config to be a getter

pleek91:
# Description
Small change to allow the api config to be a getter. Which will simplify some of our implementation in other projects.


znicholasbrown (approved):
nice

PR #2645, opened 2024-08-05
Enhancement: Disable deployments

### LINKED ISSUE #2594: Add toggles for new Deployment property: "disabled"
Depends on: https://github.com/PrefectHQ/prefect/issues/14529

- [x] Toggle on the right side of each item deployment list on the deployments page
- [x] Toggle at the top of the deployments page above the schedule and trigger sections
- [x] Disable the `pause` and schedule toggles when a deployment is `disabled`
- [x] ~Disable editing deployment schedules when a deployment is `disabled`~

dylanbhughes:
Closes #2594


src/components/DeploymentDetails.vue
```diff
@@ -30,7 +31,18 @@
           <div v-for="automation in relatedAutomations" :key="automation.id" class="deployment-details-triggers__value">
             <AutomationIconText :automation-id="automation.id" />
           </div>
+
+          <!-- If the button has a :to set, it doesn't seem to respect the :disabled param -->
```
znicholasbrown, line 35: This feels like a bug we can solve, opened a ticket [here](https://github.com/PrefectHQ/prefect-design/issues/1406)

src/components/DeploymentDetails.vue
```diff
@@ -16,7 +16,8 @@
       <template #label>
         <div class="deployment-details__schedule-label">
           <span>Schedules</span>
-          <DeploymentToggle v-if="can.update.deployment && deployment.schedules.length" :deployment="deployment" @update="emit('update')" />
+          <!-- We need to display this toggle if the deployment is paused so a user can enabled it -->
+          <DeploymentToggle v-if="can.update.deployment && (deployment.schedules.length > 3 || deployment.paused)" :deployment="deployment" @update="emit('update')" />
```
znicholasbrown, line 20: What's with the magic number 3 here?

src/components/DeploymentDisableToggle.vue
```diff
@@ -0,0 +1,64 @@
+<template>
+  <p-tooltip text="Deployment enabled" side="left">
```
znicholasbrown, line 2: Maybe we can make this a little more explicit with something like:
`Deployment ${enabled ? 'enabled' : 'disabled'} (click to ${enabled ? 'disable' : 'enable'})`, as-is it might be a little confusing

src/components/DeploymentList.vue
```diff
@@ -75,14 +75,17 @@
...
       <template #action="{ row }">
-        <DeploymentMenu
-          class="deployment-list__menu"
-          small
-          show-all
-          :deployment="row"
-          flat
-          @delete="refresh"
-        />
+        <div class="deployment-list__action">
```
znicholasbrown, line 78: Not blocking, in case you weren't aware - if you don't want to wrap this with a div + new class, you can use the existing column class passed at the table root `.deployment-list__action-column`

src/components/DeploymentSchedulesFieldset.vue
```diff
@@ -6,7 +6,7 @@
 
     <ScheduleFormModal v-if="deployment.can.update" :active="null" :schedule="null" :job-variables="{}" @submit="createSchedule">
       <template #default="{ open }">
-        <p-button size="sm" icon="PlusIcon" @click="open">
+        <p-button size="sm" icon="PlusIcon" :disabled="deployment.disabled" @click="open">
```
znicholasbrown, line 9: Should schedules not be modifiable if a deployment is disabled? I'd sort of expect I could modify it all I want, it just wouldn't submit work

src/components/DeploymentStatusIcon.vue
```diff
@@ -23,6 +23,8 @@
         return 'One or more processes are actively polling this deployment'
       case 'not_ready':
         return 'This deployment has not been polled in the past 60 seconds'
+      case 'disabled':
+        return 'This deployment is disabled and will not be polled'
```
znicholasbrown, line 27: This probably applies to the `not_ready` status above as well but I find this wording to be particularly unclear; will work still build up (e.g. late runs)? I don't have a great suggestion (sorry) but would like if we could make this a little clearer, I don't quite know what the behavior is though

src/components/RunMenu.vue
```diff
@@ -5,7 +5,7 @@
         ref="runButton"
         class="run-menu__run-button"
         icon-append="PlayIcon"
-        :disabled="deployment.deprecated"
+        :disabled="deployment.deprecated || deployment.disabled"
```
znicholasbrown, line 8: I think same question here, though maybe it's ok if this is disabled; is this an API restriction as well?

znicholasbrown (approved):
Some comments but nothing blocking

src/components/DeploymentDetails.vue
dylanbhughes: Discussed with the team generally when I demoed this initially -- mostly a matter of taste w/ regards to the space. Happy to put it in a config file somewhere if that's the correct pattern

src/components/DeploymentSchedulesFieldset.vue
dylanbhughes: I think flipping a schedule is an action that a user expects to have a result. If I flip the toggle and nothing happens, that'd be pretty confusing if I'm not paying attention to the disabled toggle.

src/components/DeploymentStatusIcon.vue
dylanbhughes: This is a great note -- I can definitely clean this language up

src/components/DeploymentSchedulesFieldset.vue
znicholasbrown: hm ok that makes sense - easy to walk back if it's confusing as well

src/components/RunMenu.vue
dylanbhughes: Correct -- attempting to create a flow run on a disabled deployment returns a 409

src/components/DeploymentStatusIcon.vue
znicholasbrown: That's already a lot cleaner! How does this behavior impact existing runs?

dylanbhughes: Doesn't mess with them at all, only prevents new runs

PR #2655, opened 2024-08-13
Chore: Tag often-used models

znicholasbrown:
Tagging these models will allow us to more easily and implicitly handle discrimination when using these models in unions.


pleek91:
@znicholasbrown couple ts errors popped up from adding these. Couple mocks (do we even need mocks anymore?) and one around artifacts, which is an interesting one that might influence this decision.

znicholasbrown:
Yeah @pleek91 we probably don't need mocks anymore tbh but it's pretty easy to update those so I've opted to do that instead; cleaned up the artifact kinds as well so this should be g2g

src/components/ArtifactResultCard.vue
znicholasbrown, line 1: This isn't referenced anywhere so removing it here since it has some outdated types

src/components/ArtifactCard.vue
```diff
@@ -43,10 +43,10 @@
   import { computed } from 'vue'
   import FormattedDate from '@/components/FormattedDate.vue'
   import { localization } from '@/localization'
-  import { Artifact } from '@/models'
+  import { Artifact, ArtifactCollection } from '@/models'
 
   const props = defineProps<{
-    artifact: Artifact,
+    artifact: Artifact | ArtifactCollection,
```
znicholasbrown, line 49: With the intro of the `kind` field, `Artifact` and `ArtifactCollection` aren't exactly equivalent; this allows them to be passed interchangeably

pleek91 (approved):

PR #2656, opened 2024-08-13
Update work pools to support object level permissions

pleek91:
# Description
Adds object level permissions to work pools. In oss these always return `true` but these will be overridden in cloud to adhere to work pool acls

part of ENTRP-27


znicholasbrown (approved):
nice to see the object level can work paying off

PR #2658, opened 2024-08-13
Chore: update route names for disabling deployments

dylanbhughes:
After some code review feedback for the backend PR we decided to update the routes for disabling/enabling deployments.


znicholasbrown (approved):
thanks for seeing this through @dylanbhughes! 🥳

PR #2666, opened 2024-08-13
Enhancement: Update favicon compositions

znicholasbrown:
These compositions were polling aggressively and needlessly - I've updated them to take the entire flow run or task run object and to instead rely on the caller for data freshness.


znicholasbrown:
This is technically a breaking change but these compositions are only used in 2 places each as part of the page components; will update those as part of the package bump when this goes out.

pleek91 (approved):
Ah yes I've run into this aggressive polling before. Good find

znicholasbrown:
i was losing my mind trying to figure out where this was happening 😬

PR #2674, opened 2024-08-14
Hide active slots on v2 limit create form

abrookins:
Showing the "active slots" field, which we use internally to track holders of an active limit, is confusing on the Create Concurrency Limit form. It will seldom be used on that form, and it conflicts (semantically) with both the "active" flag and the "concurrency limit" field, making users unsure how to proceed.

The API endpoint to create limits defaults to 0 for this property, so we can skip it from the POST.

See: https://github.com/PrefectHQ/prefect/issues/13361

I also get confused by this every time I use the form. 😂


cicdw (approved):
LGTM

pleek91 (approved):
LGTM

PR #2682, opened 2024-08-23
Enhancement: Allow resizing deployment name column

znicholasbrown:
This PR updates some column sizing on the deployment table and allows resizing the name column. This isn't a perfect solution but is probably the least-involved way we can handle this atm. 

https://github.com/user-attachments/assets/1c72368d-83ee-459d-94e7-ff145d182188

### DIFF
--- a/src/components/DeploymentList.vue
+++ b/src/components/DeploymentList.vue
@@ -172,6 +172,9 @@
   const columns = computed<TableColumn<Deployment>[]>(() => [
     {
       label: 'Deployment',
+      minWidth: '200px',
+      width: '200px',
+      maxWidth: '1000px',
     },
     {
       label: 'Status',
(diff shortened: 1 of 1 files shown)


znicholasbrown:
In case the tradeoffs here aren't clear; since this is doing a pretty dumb resize of the column. Expanding it past the min width needed to display the table will cause the table to scroll horizontally; not ideal but maybe not a deal breaker. Additionally the resize won't persist through pagination.

collincchoy (approved):
i can dig it 👍

collincchoy:
> In case the tradeoffs here aren't clear; since this is doing a pretty dumb resize of the column. Expanding it past the min width needed to display the table will cause the table to scroll horizontally; not ideal but maybe not a deal breaker. Additionally the resize won't persist through pagination.

i kinda like that it won't persist so if you get yourself into a janky view, it'll reset pretty quickly/easily.

znicholasbrown:
@collincchoy good point

PR #2687, opened 2024-08-27
Update some work queue related stuff to be compatible with cloud acls

pleek91:
# Description
Work queue's don't have ACLs of their own, but they need to respect the ACL of the work pool they belong to. So I'm adding an object level `can` to the `WorkPoolQueue` which cloud can use for this.


collincchoy (approved):
Ah yeah i guess this makes sense since we don't always/usually have immediate access to a work queue's parent pool. 

👍

PR #2689, opened 2024-08-27
Update permission checks when working with work queues

pleek91:
# Description
Now that work pool queues have an object level can object we can check that when appropriate. Also adds permission checks for creating and bulk deleting work queues based on the work pool.

### DIFF
--- a/src/components/WorkPoolQueuesTable.vue
+++ b/src/components/WorkPoolQueuesTable.vue
@@ -8,7 +8,7 @@
         <p-button size="sm" icon="PlusIcon" :to="routes.workPoolQueueCreate(workPoolName)" />
       </template>
 
-      <template v-if="workPool?.can.delete">
+      <template v-if="workPool?.can.update">
         <WorkPoolQueuesDeleteButton :work-pool-name="workPoolName" :work-pool-queues="selected" @delete="handleDelete" />
       </template>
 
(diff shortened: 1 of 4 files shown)


znicholasbrown (approved):

src/components/WorkPoolQueuesTable.vue
```diff
@@ -3,9 +3,14 @@
...
       <ResultsCount v-if="selected.length == 0" label="Work Queue" :count="filteredWorkPoolQueues.length" />
       <SelectedCount v-else :count="selected.length" />
-      <p-button v-if="can.create.work_queue && !selected.length" size="sm" icon="PlusIcon" :to="routes.workPoolQueueCreate(workPoolName)" />
 
-      <WorkPoolQueuesDeleteButton v-if="can.delete.work_queue" :work-pool-name="workPoolName" :work-pool-queues="selected" @delete="handleDelete" />
+      <template v-if="workPool?.can.update && !selected.length">
+        <p-button size="sm" icon="PlusIcon" :to="routes.workPoolQueueCreate(workPoolName)" />
+      </template>
+
+      <template v-if="workPool?.can.delete">
```
collincchoy, line 11: was thinking that these seemed potentially confusing at a glance that creating a work queue is tied to workPool.can.update and deleting a work queue is tied to workPool.can.delete. 


but then on second thought... i think it shouldn't make a difference right b/c both are ultimately `manage` level permissions?

collincchoy (approved):

src/components/WorkPoolQueuesTable.vue
pleek91: Hmm yeah now that I'm thinking about it the delete should probably be `update` because you're updating the work queues? But I think you're right they map to manage either way.

pleek91: Update in https://github.com/PrefectHQ/prefect-ui-library/pull/2689/commits/886424b286d970535e76231c03e8e875c322b3ed

PR #2692, opened 2024-08-27
Fix: Make deployment toggles visible but disabled for deployments when a user does not have update deployment permissions

### LINKED ISSUE #15071: Runner Role should be able to view individual schedule toggles and the pause/resume toggle if the deployment is paused
### Bug summary

First image is what I personally see with workspace role `runner`. Second image is what a colleague sees with workspace role `owner`.

For me as a runner it looks like this deployment should run every night, but in actuality it is disabled, which I have no way of seeing.

### Version info (`prefect version` output)

```Text
Latest using Prefect Cloud
```

### Additional context

image
image

zhen0:
Currently we hide action items from users who do not have permission to take the relevant actions.  This includes hiding the deployment schedule toggle from a user who does not have permission to update a deployment.  However, the toggle shows information as well as an action.  

This PR updates the different deployment toggles to allow a toggle to convey information about a deployment's/schedule's status and instead disables the toggle so a user without permissions cannot update the deployment. It also updates the related tooltip to advise the user they don't have permission to update the deployment. 

Current - A user with Runner permissions can see that their deployment has two schedules but can not tell that one has been disabled:
image

This PR - A user with Runner permissions can see that a schedule is disabled but can not update it:
image

This PR is separate from the deployment status updates in https://github.com/PrefectHQ/prefect-ui-library/pull/2645 and is aimed only at making status more visible for users who do not have permissions to take action. 

Resolves https://github.com/PrefectHQ/prefect/issues/15071

### DIFF
--- a/src/components/DeploymentDisableToggle.vue
+++ b/src/components/DeploymentDisableToggle.vue
@@ -1,6 +1,6 @@
 <template>
   <p-tooltip :text="tooltipText" side="left">
-    <p-toggle v-if="deployment.can.update" v-model="internalValue" :state :disabled="deployment.deprecated" />
+    <p-toggle v-model="internalValue" :state :disabled="deployment.deprecated || !deployment.can.update" />
   </p-tooltip>
 </template>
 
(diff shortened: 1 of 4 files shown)


znicholasbrown (approved):
lovely improvement

PR #2695, opened 2024-08-28
Fix: UI strips new-line characters for Secret block

### LINKED ISSUE #6949: UI strips new-line characters for Secret block
### First check

### Bug summary

Pasting a value with new line characters into the `Value` field of the `Secret` strips the new line characters.

### Reproduction

1. Create a new Secret block in the UI with `Block Name` as `from-ui` and `Value` as,

```
line one
line two
```

2. Create a new Secret block with Python as,

```py
from prefect.blocks.system import Secret

data = """
line one
line two
"""

s = Secret(value=data)
s.save("from-code")
```

### Error

```py
>>> from prefect.blocks.system import Secret
>>> Secret.load("from-code").get()
'\nline one\nline two\n'
>>> Secret.load("from-ui").get()
' line one line two'
```

### Browers

- [X] Chrome
- [ ] Firefox
- [X] Safari
- [ ] Edge

### Prefect version

```
Version:             2.4.2
API version:         0.8.0
Python version:      3.9.13
Git commit:          65807e84
Built:               Fri, Sep 23, 2022 10:43 AM
OS/Arch:             darwin/arm64
Profile:             emil-default
Server type:         cloud
```

### Additional context

_No response_

pleek91:
# Description
Closes https://github.com/PrefectHQ/prefect/issues/6949

Updates how we generate inputs for schema parameters that are type `text`. All text values will use a `textarea` rather than an `input` with type text/password. This specifically fixes entering multi line secrets but also makes working with text more flexible for all text values.

### DIFF
--- a/src/schemas/components/SchemaFormPropertyString.vue
+++ b/src/schemas/components/SchemaFormPropertyString.vue
@@ -3,7 +3,7 @@
 </template>
 
 <script lang="ts" setup>
-  import { PCombobox, PCodeInput, PTextInput, State } from '@prefecthq/prefect-design'
+  import { PCombobox, PCodeInput, PTextarea, State } from '@prefecthq/prefect-design'
   import { computed } from 'vue'
   import SchemaFormPropertyDate from '@/schemas/components/SchemaFormPropertyDate.vue'
   import SchemaFormPropertyDateTime from '@/schemas/components/SchemaFormPropertyDateTime.vue'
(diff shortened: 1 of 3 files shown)


src/schemas/components/SchemaFormPropertyString.vue
```diff
@@ -70,9 +61,10 @@
       })
     }
 
-    return withProps(PTextInput, {
+    return withProps(PTextarea, {
```
znicholasbrown, line 64: I forget how we treat this - do we need to restrict resizing at all?

znicholasbrown (approved):

src/schemas/components/SchemaFormPropertyString.vue
pleek91: Its 1 row by default but allowed to resize. And PTextarea only allows vertical resizing so it doesn't overflow the layout.

znicholasbrown: ah nice ok - thanks for confirming, that's exactly what i was hoping for.

PR #2702, opened 2024-09-03
Add "Reset" option for v2 concurrency limits

abrookins:
In OSS-5405, users report not understanding what "active slots" does on the v2 concurrency limit forms. We removed the field from the create form but left it on "Edit." The reason it's on the Edit form is to allow users to reset active slots to 0, which is effectively what the "Reset" menu option does for V1/Task Run concurrency limits.

But why not have consistent UX between them? We could have a reset menu option for v2 limits.

Screenshot 2024-09-03 at 12 15 40 PM

Screenshot 2024-09-03 at 12 15 50 PM


src/components/ConcurrencyLimitV2ResetModal.vue
```diff
@@ -0,0 +1,64 @@
...
+  }>()
+
+  const internalShowModal = computed({
+    get() {
+      return props.showModal
+    },
+    set(value: boolean) {
+      emit('update:showModal', value)
+    },
+  })
```
collincchoy, line 38: nit:

You can use [`defineModel`](https://vuejs.org/api/sfc-script-setup.html#definemodel) now for this! 

Simplifies this boilerplate down to something like:
```vue
const props = defineProps<{ concurrencyLimit: ConcurrencyV2Limit }>()
const showModal = defineModel<boolean>("showModal")

// then just use showModal instead of internalShowModal
```

Example - https://github.com/PrefectHQ/prefect-ui-library/blob/ca85777d377dde6c21103f2df4a9b3b6577bfcee/src/components/FlowRunResumeModal.vue#L35-L36

collincchoy (approved):
Wonder if the menu option being labeled a bare "Reset" is explicit enough but code changes lgtm.

left 1 comment that'd let you reduce some boilerplate-y stuff using a newer vue feature, but non-blocking

src/components/ConcurrencyLimitV2ResetModal.vue
abrookins: Thanks @collincchoy, I'll try that!

abrookins:
@cicdw @desertaxle What do you think about this? Summary: add a "Reset" menu item to global concurrency limits that works the same way as the "Reset" menu item on task run concurrency limits.

cicdw:
@abrookins this makes sense to me!

abrookins:
> Wonder if the menu option being labeled a bare "Reset" is explicit enough but code changes lgtm.
> 
> left 1 comment that'd let you reduce some boilerplate-y stuff using a newer vue feature, but non-blocking

I'm going to run with "Reset" to match the "Reset" in the task run concurrency limits UI, though in general I agree with you.

PR #2718, opened 2024-09-09
Add a Resume Flow Run action option

abrookins:
A follow-up to https://github.com/PrefectHQ/prefect/pull/15269 that adds the option to select the Resume Flow Run action when creating an automation.


src/automations/components/AutomationActionDescriptionResumeFlowRun.vue
```diff
@@ -0,0 +1,3 @@
+<template>
+  <p>Resume a flow run</p>
```
pleek91, line 2: We try to use some more specific language when we're inferring based on the trigger. For suspend and cancel we say "Suspend flow run inferred from the triggering event". Might be good to be consistent here. 

Alternatively, you could update the "AutomationActionDescriptionSuspendCancelFlowRun.vue" component to handle resume as well. But I kinda like the simplicity of this static component.

pleek91 (approved):
LGTM! One description suggestion but not blocking.

src/automations/components/AutomationActionDescriptionResumeFlowRun.vue
abrookins: Oh, thanks for the heads-up!

PR #2722, opened 2024-09-10
Enhancement: Excessive polling in flow run menu component

znicholasbrown:
This component was hitting the flow run endpoint every 3 seconds to watch for state changes (among other things). However, all of the places where this component is used should have access to the full flow run object so it's safer to rely on upstream fetch strategies.


pleek91 (approved):
Yep nice 👍

PR #2740, opened 2024-09-18
Chore: Update empirical policy

znicholasbrown:
Deprecates `max_retries` and `retry_delay_seconds` fields, adds `retry_jitter_factor` field.

### DIFF
--- a/src/components/TaskRunDetails.vue
+++ b/src/components/TaskRunDetails.vue
@@ -60,16 +60,9 @@
 
     <p-key-value label="Version" :value="taskRun.taskVersion" :alternate="alternate" />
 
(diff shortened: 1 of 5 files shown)


src/components/TaskRunDetails.vue
```diff
@@ -60,16 +60,9 @@
 
     <p-key-value label="Version" :value="taskRun.taskVersion" :alternate="alternate" />
 
-    <div v-if="taskRun.empiricalPolicy?.maxRetries">
```
znicholasbrown, line 63: This field has been deprecated in Core for ~2 years so I think we're safe to stop displaying it.

dylanbhughes (approved):

PR #2745, opened 2024-09-20
Deployment concurrency limits/configurable collision strategy

collincchoy:
This PR adds the ability to edit/update a deployment's `deployment.concurrency_options.collision_strategy` from the UI.

Screenshot 2024-09-20 at 2 39 43 PM


pleek91 (approved):
LGTM

abrookins:
I wonder if we can make the options more self-explanatory. 🤔 What if the user-facing text for the options was something like "Enqueue new runs" or "Cancel new runs"?

collincchoy:
> I wonder if we can make the options more self-explanatory. 🤔 What if the user-facing text for the options was something like "Enqueue new runs" or "Cancel new runs"?

It's a good question. I considered it but decided to lean on parity with the client-side python implementation to reduce confusion.

PR #2754, opened 2024-09-24
Add a page heading component for an automation with "Explore Events" button

pleek91:
# Description
Adds a page heading component for an automation. The heading is currently duplicated in both oss and cloud. Adding a single page heading component here and adding the "Explore Events" button as part of closing out https://github.com/PrefectHQ/prefect/issues/15430


src/components/PageHeadingAutomation.vue
```diff
@@ -0,0 +1,48 @@
...
+  const routes = useWorkspaceRoutes()
+
+  const crumbs = computed<Crumb[]>(() => [
+    { text: 'Automations', to: routes.automations() },
+    { text: props.automation.name },
+  ])
+
+  const filter = computed<PartialWorkspaceEventsFilter>(() => ({
+    anyResource: {
+      id: [`prefect-cloud.automation.${props.automation.id}`],
```
znicholasbrown, line 41: is this still correct?

znicholasbrown (approved):
1 question, otherwise lgtm

src/components/PageHeadingAutomation.vue
pleek91: As far as I can tell both cloud and oss use that id.

PR #2783, opened 2024-10-18
Deployment schedule active out of sync edit modal and cards

collincchoy:
This PR fixes some data inconsistency issues between the deployment schedule cards and the edit schedule modal. In particular, the active/inactive toggle state is not in-sync between the following UI elements following updates from either one. This is mostly due to overuse of component-local state that should instead be derived from props. 

| Minimal card view | Edit modal |
| ---- | ---- |
| image | image |

Closes CLOUD-358


pleek91 (approved):
LGTM

PR #2789, opened 2024-10-24
Enhancement: Show "No access" when missing read access

znicholasbrown:
This is only for a few components but feels like a material improvement.


pleek91 (approved):
Nice, I like it

PR #2835, opened 2024-12-02
Fix flow run activity chart

pleek91:
# Description
Fixes an issue where when viewing future times could result in bars overflowing the activity.

### DIFF
--- a/src/components/FlowRunsBarChart.vue
+++ b/src/components/FlowRunsBarChart.vue
@@ -57,6 +57,10 @@
       return null
     }
 
+    if (barsDebounced.value === 0) {
+      return null
+    }
+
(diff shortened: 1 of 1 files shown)


src/components/FlowRunsBarChart.vue
```diff
@@ -57,6 +57,10 @@
       return null
     }
 
+    if (barsDebounced.value === 0) {
+      return null
+    }
```
pleek91, line 62: Unrelated fix. Noticed that before the chart was sized it would sometimes make a request with a limit of `0`. Which of course would return 0 flow runs. So we don't need to make a request for that.

znicholasbrown (approved):

PR #2844, opened 2024-12-03
Update prefect-design and fix type errors

pleek91:


src/components/WorkspaceEventsTimeline.vue
```diff
@@ -19,15 +19,17 @@
...
   import WorkspaceEventIcon from '@/components/WorkspaceEventIcon.vue'
   import WorkspaceEventListItemDate from '@/components/WorkspaceEventListItemDate.vue'
   import WorkspaceEventsTimelineContent from '@/components/WorkspaceEventsTimelineContent.vue'
   import { WorkspaceEvent, isWorkspaceEvent } from '@/models/workspaceEvent'
 
+  type WorkspaceEventTimelineItem = TimelineItem & WorkspaceEvent
+
   const props = defineProps<{
-    events: WorkspaceEvent[],
+    events: WorkspaceEventTimelineItem[],
```
pleek91, line 32: Since `WorkspaceEvent` has no properties in common with `TimelineItem` it was triggering typescripts [week type detection](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-4.html#weak-type-detection). But since `TimelineItem` has no require properties we can just use an intersection type.

znicholasbrown (approved):

PR #2851, opened 2024-12-09
Enhancement: Work pool name validation rule

znicholasbrown:
Short cuts an existing API-side validation rule by surfacing the internal name restriction directly in the work pool creation form.

Screenshot 2024-12-08 at 7 33 29 PM


pleek91 (approved):
❤️

PR #2857, opened 2024-12-16
Fix FlowRunMenu components subscribing to the flow run multiple times

pleek91:
# Description
The `FlowRunMenu` component accepts a `FlowRun` but some of the modal components accepts just the id and then created their own subscriptions. We already have events set up for things like refreshing so there isn't any reason these component should do their own subscribing. 

This was causing api request spam on the new runs page in cloud because for some reason the batch processor wasn't grouping the requests together (likely due to slightly different intervals due to a rendering waterfall).


znicholasbrown (approved):
nice find

PR #2870, opened 2024-12-19
Add context for notification actions

devangrose:
Added a description component for notification actions. 

Will conditionally parse email recipients and display as well

image

image

### DIFF
--- a/src/automations/components/AutomationActionDescriptionSendNotification.vue
+++ b/src/automations/components/AutomationActionDescriptionSendNotification.vue
@@ -1,0 +1,44 @@
+<template>
+  <div class="automation-action-description-send-notification">
+    Send {{ indefiniteArticle() }} {{ blockDocument?.blockType.name.toLocaleLowerCase() ?? '' }} notification using
+    <BlockIconText :block-document-id="action.blockDocumentId" />
+    {{ recipients }}
+  </div>
+</template>
+
+<script lang="ts" setup>
+  import { computed } from 'vue'
+  import { AutomationActionSendNotification } from '@/automations/types/actions'
+  import { BlockIconText } from '@/components'
+  import { useBlockDocument } from '@/compositions'
+
+  const props = defineProps<{
+    action: AutomationActionSendNotification,
+  }>()
+
+  const { blockDocument } = useBlockDocument(() => props.action.blockDocumentId)
+
+  const indefiniteArticle = (): string => {
+    return ['a', 'e', 'i', 'o', 'u'].includes(blockDocument.value?.blockType.name[0].toLowerCase() ?? '') ? 'an' : 'a'
+  }
+
+  const recipients = computed(() => {
+    if (blockDocument.value?.blockType.name === 'Email') {
+      try {
+        return `to ${JSON.parse(blockDocument.value.data.emails as string).join(', ')}`
+      } catch (error) {
+        return ''
+      }
+    }
+    return ''
+  })
+</script>
+
+<style>
+.automation-action-description-send-notification { @apply
+  flex
+  flex-wrap
+  gap-2
+  items-center
+}
+</style>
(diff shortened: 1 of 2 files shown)


src/automations/components/AutomationActionDescriptionSendNotification.vue
```diff
@@ -0,0 +1,52 @@
...
+
+  const props = defineProps<{
+    action: AutomationActionSendNotification,
+  }>()
+
+  const { blockDocument } = useBlockDocument(props.action.blockDocumentId)
+
+  const {
+    blockDocumentId,
+  } = props.action
```
pleek91, line 27: Since props is reactive if we destructure like this the reactivity of blockDocumentId will be broken. Meaning if the action being passed to this component changes, the blockDocumentId won't update to reflect that the action changed. 

Either 
a. Use a computed 
b. Use action.blockDocumentId directly (preferred when just using this in the template)

src/automations/components/AutomationActionDescriptionSendNotification.vue
```diff
@@ -0,0 +1,52 @@
...
+
+  defineOptions({
+    inheritAttrs: false,
+  })
+
+  const props = defineProps<{
+    action: AutomationActionSendNotification,
+  }>()
+
+  const { blockDocument } = useBlockDocument(props.action.blockDocumentId)
```
pleek91, line 23: Accessing `props.action.blockDocumentId` directly like this will break the reactivity because it's passing the value of the id but it cannot track the effects. Here you can use a getter

```suggestion
  const { blockDocument } = useBlockDocument(() => props.action.blockDocumentId)
```

src/automations/components/AutomationActionDescriptionSendNotification.vue
```diff
@@ -0,0 +1,52 @@
...
+
+<script lang="ts" setup>
+  import { computed } from 'vue'
+  import { AutomationActionSendNotification } from '@/automations/types/actions'
+  import { BlockIconText } from '@/components'
+  import { useBlockDocument } from '@/compositions'
+
+  defineOptions({
+    inheritAttrs: false,
+  })
```
pleek91, line 17: Is there a specific reason you don't want to inherit attrs?

pleek91 (changes requested):

src/automations/components/AutomationActionDescriptionSendNotification.vue
devangrose: Ah that makes sense! Definitely something I'm less familiar with in Vue but the concept rings true.

src/automations/components/AutomationActionDescriptionSendNotification.vue
pleek91: Ah I see its on `AutomationActionDescriptionRunDeployment` which is a fragment because it also renders a modal. When there are multiple root elements vues [fallthrough attributes](https://vuejs.org/guide/components/attrs.html#fallthrough-attributes) won't work automatically so you have to specify where they should fallthrough to. Which is done by disabling `inheritAttrs` and then binding `$attrs` to the element you want them to fall through to. 

In your case you have a single root element so you don't need to do anything special. So that can just be removed. Otherwise the `automation-action-description` class from `AutomationActionDescription.vue` won't end up being bound to your component.

src/automations/components/AutomationActionDescriptionSendNotification.vue
```diff
@@ -0,0 +1,44 @@
+<template>
+  <div class="automation-action-description-send-notification">
+    Send {{ indefiniteArticle() }} {{ blockDocument?.blockType.name.toLocaleLowerCase() ?? '' }} notification using
+    <BlockIconText :block-document-id="props.action.blockDocumentId" />
```
pleek91, line 4: This works as is but is kinda an anti pattern. Props are automatically available in the template by their names. So you don't need to access the `props` reactive object. 

```suggestion
    <BlockIconText :block-document-id="action.blockDocumentId" />
```

pleek91 (approved):
One last suggestion but otherwise this LGTM!

PR #2879, opened 2025-01-07
Enhancment: show slas on deployment details component

dylanbhughes:


src/components/DeploymentDetails.vue
```diff
@@ -137,6 +146,7 @@
 
   const can = useCan()
   const api = useWorkspaceApi()
+  const serviceLevelAgreements = computed(() => props.slas)
```
pleek91, line 149: Don't use a computed for this. You can just reference `slas` directly in the template.

pleek91 (approved):
One nit, not blocking

PR #2919, opened 2025-02-04
[Feature] Allow the specification of parameter overrides per schedule

desertaxle:
This PR updates the `ScheduleFormModal` to provide an input for parameter overrides for an individual schedule.

By default, schedules start with no overrides, but users can choose which parameters should be overridden. After choosing a parameter to override, the form will update to include fields for that parameter with the default value defined on the base deployment.

Here's what it looks like in action:

https://github.com/user-attachments/assets/7e971ccc-7a93-4748-bf0d-d82697681e26

Closes OSS-6191

### DIFF
--- a/src/components/DeploymentDetails.vue
+++ b/src/components/DeploymentDetails.vue
@@ -120,19 +120,19 @@
 </template>
 
 <script lang="ts" setup>
-  import { showToast } from '@prefecthq/prefect-design'
-  import { computed } from 'vue'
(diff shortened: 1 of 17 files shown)


cicdw (approved):
Approving from the UX perspective - LGTM!

znicholasbrown (approved):

znicholasbrown:
cooool

PR #2920, opened 2025-02-06
[Feature] Add ability to set a slug for a deployment schedule

desertaxle:
This PR adds the ability to create a unique slug for a deployment schedule. This slug allows the users to create meaningful identifiers for schedules. This is useful with the recent addition of parameter overrides since two schedules could run at the same frequency with different parameters. The slug will also be used as the created by display value for auto-scheduled runs.

Here's what it looks like in action:

https://github.com/user-attachments/assets/a5143819-d523-40bd-9c7c-42059dde83a8


desertaxle:
I'm going to add some validation to the slug field to ensure it doesn't have spaces and whatnot.

pleek91 (approved):
LGTM

cicdw (approved):
Approving from UX perspective!

desertaxle:
@pleek91 I added validation in https://github.com/PrefectHQ/prefect-ui-library/pull/2920/commits/feb048d7bb4da87c4fea42d6466e1b278d99aef9. Probably worth another look to make sure I didn't do anything foolish.

pleek91:
@desertaxle LGTM

PR #2995, opened 2025-04-18
enhancement: Updates DeploymentVersionInfo typings

devinvillarosa:
Decided to re-add this PR, because it looks like these types _are_ expected, and should provide a better devX in knowing what fields to expect

### DIFF
--- a/src/models/DeploymentVersionInfo.ts
+++ b/src/models/DeploymentVersionInfo.ts
@@ -1,7 +1,35 @@
-export type DeploymentVersionInfo = {
-  type: string,
-  base: string | null,
-  branch: string | null,
+type SimpleVersionInfo = {
+  type: 'prefect:simple',
+  /** "" as default */
   version: string,
-  url: string | null,
-} & Record<string, unknown>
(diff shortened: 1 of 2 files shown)


src/models/DeploymentVersionInfo.ts
```diff
@@ -1,7 +1,35 @@
...
+  type: 'container:docker',
+  version: string,
+  branch: string,
+  url: string,
+  imageName: string,
+  registry: string,
+  image: string,
+}
+
+export type DeploymentVersionInfo = SimpleVersionInfo | GithubVersionInfo | GitVersionInfo | DockerVersionInfo
```
znicholasbrown, line 35:
```suggestion
export type DeploymentVersionInfo = (SimpleVersionInfo | GithubVersionInfo | GitVersionInfo | DockerVersionInfo) & Record<string, unknown>
```

znicholasbrown: This'll give us some breathing room as we expect these types

znicholasbrown (approved):

PR #3006, opened 2025-04-24
Chore: Fix some warnings on the flow run filter

znicholasbrown:
These were throwing a bunch of console warnings cause the deployment id is required (but the type doesn't require it) and the mapper wasn't removing the empty object in time. Doesn't impact functionality.


src/maps/filters.ts
```diff
@@ -173,6 +173,7 @@ export const mapDeploymentVersionIdFilter: MapFunction<DeploymentVersionIdFilter
 }
 
 export const mapDeploymentVersionInfoFilter: MapFunction<DeploymentVersionInfoFilter, DeploymentVersionInfoFilterRequest | undefined> = function(source) {
+  console.log(source)
```
devinvillarosa, line 176: Should probably remove?

devinvillarosa (approved):

PR #3042, opened 2025-05-30
Enhancement: Add link to deployment from dashboard run list item

zhen0:
Adds a link to the flow run list item on the dashboard so that a user can go straight to a deployment from the dashboard.  

Addresses https://github.com/PrefectHQ/prefect/issues/15405


devinvillarosa (approved):

znicholasbrown (approved):

src/components/DeploymentIconText.vue
```diff
@@ -33,4 +33,12 @@
...
\ No newline at end of file
+</script>
+
+<style>
+.deployment-icon-name { @apply
+  truncate
+  max-w-48
+  block
+}
+</style>
```
znicholasbrown, line 44: Ah wait actually - i don't know this is correct - we'd want to put this sort of styling on the parent of this component and not the component itself, setting only the styles this component actually needs. So in this case it's probably `max-w-full`

znicholasbrown (changes requested):

src/components/DeploymentIconText.vue
zhen0: I'm not sure I follow you exactly.  What I have added is working though.  

Updated:

image


Previous:

image

PR #3056, opened 2025-06-24
feat: Adds a flow state type override to be in sync with websocket data

devinvillarosa:
Adds support from web socket events to be at a terminal state from web socket data.
This is an opt-in prop

https://github.com/user-attachments/assets/9779d7d0-0cdf-4671-9433-8d989314dca2

### DIFF
--- a/src/components/FlowRunGraph.vue
+++ b/src/components/FlowRunGraph.vue
@@ -29,13 +29,14 @@
   import { useTaskRunsCount } from '@/compositions/useTaskRunsCount'
   import { useWorkspaceApi } from '@/compositions/useWorkspaceApi'
   import { FlowRun } from '@/models/FlowRun'
-  import { ServerStateType, isTerminalStateType } from '@/models/StateType'
+  import { ServerStateType, StateType, isTerminalStateType } from '@/models/StateType'
   import { getApiErrorMessage } from '@/utilities/errors'
 
   const NODE_COUNT_TO_REQUIRED_OPT_IN = 2000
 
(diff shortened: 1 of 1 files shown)


src/components/FlowRunGraph.vue
```diff
@@ -29,13 +29,14 @@
...
   import { FlowRun } from '@/models/FlowRun'
-  import { ServerStateType, isTerminalStateType } from '@/models/StateType'
+  import { ServerStateType, StateType, isTerminalStateType } from '@/models/StateType'
   import { getApiErrorMessage } from '@/utilities/errors'
 
   const NODE_COUNT_TO_REQUIRED_OPT_IN = 2000
 
   const props = defineProps<{
     flowRun: FlowRun,
+    flowRunStateTypeOverride?: StateType,
```
znicholasbrown, line 39: Just so I understand - this is necessary because we're expecting a lag between the flow run state change event and the state on the `flowRun` object, right?

znicholasbrown (approved):

src/components/FlowRunGraph.vue
devinvillarosa: yes.
There's `flowRun` which is provided by the REST API.
And then there's the `flowRunStateType` provided by the web socket.

So if there's no web socket event found, use the REST API data. else, override it with the web socket event data

devinvillarosa: ☝️ Let me try that out before re-opening this PR

PR #3106, opened 2025-11-20
fix(FlowRunFilteredList): reset pagination to page 1 when state filter changes

### LINKED ISSUE #3105: Pagination not reset when flow run filters change in FlowRunFilteredList
**Describe the bug**

When using the `FlowRunFilteredList` component (e.g., on Work Pool, Deployment, or Flow pages), changing the flow run state filter (or other filters) does **not reset the pagination to page 1**. 

If the user is on page 2+ and applies a new filter that returns fewer results, they may land on an empty page or see stale/incorrect data because the page number remains unchanged.

**Expected behavior**

Whenever any filter affecting the result set changes (e.g., state, search term), the pagination should automatically reset to **page 1** to ensure users see relevant results.

**To Reproduce**

1. Go to a Work Pool with many flow runs.
2. Navigate to page 2 or higher in the Runs list.
3. Apply a state filter (e.g., "Crashed").
4. Observe that the URL still contains `runs.page=3`, and the list may be empty even if crashed runs exist on page 1.

**Additional context**

This may affects all views using `FlowRunFilteredList`?  including:
- Work Pool → Runs tab
- Deployment → Runs tab
- Flow → Runs tab

A fix can be implemented by watching filter changes in the component and resetting `filter.page = 1`.

TomMonkeyMan:
Closes #3105

Prefect main repo [issue-19480](https://github.com/PrefectHQ/prefect/issues/19480).

Added a watcher in `FlowRunFilteredList.vue` to monitor changes to `filter.flowRuns.state.name`. When the filter changes, `filter.page` is reset to `1`.

This ensures users always start from the first page of filtered results, avoiding empty or misleading pages.

- Go to a Work Pool with many runs
- Navigate to page 3+
- Change the state filter
- Verify URL updates to `runs.page=1` and correct results are shown

### DIFF
--- a/src/components/FlowRunFilteredList.vue
+++ b/src/components/FlowRunFilteredList.vue
@@ -52,7 +52,7 @@
 <script lang="ts" setup>
   import { useDebouncedRef, useLocalStorage } from '@prefecthq/vue-compositions'
   import merge from 'lodash.merge'
-  import { computed, ref } from 'vue'
+  import { computed, ref, watch } from 'vue'
   import {
     ResultsCount,
     StateNameSelect,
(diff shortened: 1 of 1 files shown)


znicholasbrown (approved):
Thanks @TomMonkeyMan !

